PR #5, opened 2025-01-15
Henryh/setup dataloader

Hgherzog:
Initial draft and setup of dataloader
prototype of integration with olmo-core soon to come

### DIFF
--- a/helios/dataset/index.py
+++ b/helios/dataset/index.py
@@ -9,7 +9,7 @@ import numpy as np
 from upath import UPath
 
 from helios.data.constants import ALL_DATA_SOURCES
-from helios.helios.dataset.utils import (
+from helios.dataset.utils import (
     DataSourceMetadataRegistry,
     FrequencyType,
     load_data_index,
@@ -38,7 +38,6 @@ class DatasetIndexParser:
         self.data_sources = ALL_DATA_SOURCES
         self.data_index_path = UPath(data_index_path)
         self.root_dir = self.data_index_path.parent
-        # Using a df as initial ingest due to ease of inspection and manipulation,
         self.data_index_df = load_data_index(data_index_path)
         self.example_id_to_sample_metadata_dict = self.data_index_df.set_index(
             "example_id"
@@ -47,11 +46,28 @@ class DatasetIndexParser:
         self.freq_metadata_df_dict = {}
         self.monthly_metadata_df_dict = {}
         for data_source in self.data_sources:
+            metadata_path = self.get_path_to_data_source_metadata(data_source, "freq")
             self.freq_metadata_df_dict[data_source] = (
-                DataSourceMetadataRegistry.load_and_validate(data_source, "freq")
+                DataSourceMetadataRegistry.load_and_validate(
+                    data_source,
+                    "freq",
+                    sentinel2_frequency_metadata_path=metadata_path
+                    if data_source == "sentinel2"
+                    else None,
+                )
+            )
+
+            metadata_path = self.get_path_to_data_source_metadata(
+                data_source, "monthly"
             )
             self.monthly_metadata_df_dict[data_source] = (
-                DataSourceMetadataRegistry.load_and_validate(data_source, "monthly")
+                DataSourceMetadataRegistry.load_and_validate(
+                    data_source,
+                    "monthly",
+                    sentinel2_monthly_metadata_path=metadata_path
+                    if data_source == "sentinel2"
+                    else None,
+                )
             )
 
         # Intersect available data sources with index column names
--- a/helios/dataset/utils.py
+++ b/helios/dataset/utils.py
@@ -8,7 +8,7 @@ import pandas as pd
 import pandera as pa
 from pandera.typing import DataFrame
 
-from helios.helios.dataset.schemas import (
+from helios.dataset.schemas import (
     Sentinel2FrequencyMetadataDataModel,
     Sentinel2MonthlyMetadataDataModel,
     TrainingDataIndexDataModel,
@@ -76,7 +76,10 @@ class DataSourceMetadataRegistry:
             raise ValueError(
                 f"Unknown frequency type {frequency_type} for data source {data_source}"
             )
-        return cls._registry[data_source][frequency_type](**kwargs)
+
+        # Get the registered function and call it with the provided kwargs
+        func = cls._registry[data_source][frequency_type]
+        return func(**kwargs)
 
 
 @DataSourceMetadataRegistry.register("sentinel2", "monthly")
--- a/tests/unit/dataset/test_index.py
+++ b/tests/unit/dataset/test_index.py
@@ -2,7 +2,7 @@
 
 import pytest
 
-from helios.helios.dataset.index import DatasetIndexParser
+from helios.dataset.index import DatasetIndexParser
 
 
 @pytest.fixture
@@ -15,16 +15,27 @@ def test_dataset_index(sample_index_path: str) -> None:
     """Test the dataset index."""
     index_parser = DatasetIndexParser(sample_index_path)
     assert len(index_parser.samples) == 4
-    assert set(index_parser.samples[0].keys()) == {
-        "data_source_paths",
-        "data_source_metadata",
-        "sample_metadata",
-    }
+
+    # Access the first sample's information
+    first_sample_info = index_parser.samples[0]
+
+    # Check the keys in the first sample's data_source_paths
+    assert set(first_sample_info.data_source_paths.keys()) == {"sentinel2"}
+
+    # Define expected paths
     expected_paths = {
         "sentinel2": "tests/fixtures/sample-dataset/sentinel2_monthly/example_001.tif"
     }
-    actual_paths = index_parser.samples[0]["data_source_paths"]
+
+    # Compare actual paths with expected paths
+    actual_paths = first_sample_info.data_source_paths
     assert len(expected_paths) == len(actual_paths)
     for data_source, expected_path in expected_paths.items():
         assert data_source in actual_paths
         assert str(actual_paths[data_source]) == expected_path
+
+    # Check the keys in the first sample's data_source_metadata
+    assert "sentinel2" in first_sample_info.data_source_metadata
+
+    # Check the keys in the first sample's sample_metadata
+    assert "example_id" in first_sample_info.sample_metadata


.github/workflows/lint.yaml
```diff
@@ -18,10 +18,10 @@ jobs:
         with:
           lfs: false
 
-      - name: Set up Python 3.12
+      - name: Set up Python 3.11
```
pbeukema, line 21: whats the reason or this?

helios/data/collator.py
```diff
@@ -0,0 +1,166 @@
...
+    static_mask: torch.Tensor  # Mask for static features
+    time_info: torch.Tensor  # Time information associated with each index
+    patch_size: float  # Size of patches
+    collate_info: dict | None  # Additional collation metadata
+
+
+def variable_time_collate_fn(
+    items: list[DatasetOutput],
+    pad_token_value: float = 0.0,
+    patch_size: int = 4,
```
pbeukema, line 38: prefer not to set default here.

helios/data/dataset.py
```diff
@@ -0,0 +1,325 @@
...
+    "B11",
+    "B12",
+]
+
+# Quick and dirty interface for data source variation types
+DATA_SOURCE_VARIATION_TYPES = Literal[
+    "space_time_varying", "time_varying_only", "space_varying_only", "static_only"
+]
+
+DATA_SOURCE_TO_VARIATION_TYPE = {
```
pbeukema: somewhat unclear what this means until you read the dict.

helios/data/collator.py
```diff
@@ -0,0 +1,166 @@
...
+    encode_ratio: float = 0.5,
+    decode_ratio: float = 0.5,
+) -> CollateFnOutput:
+    """Collate function for inputs with variable time data."""
+    assert items
+    max_len = max(item.sample_metadata["num_timesteps"] for item in items)
+    print(f"Max len: {max_len}")
```
pbeukema: log instead of print. also this will probably get flagged by linter?

helios/data/dataset.py
```diff
@@ -0,0 +1,325 @@
...
+from upath import UPath
+
+from helios.data.utils import (
+    load_data_index,
+    load_sentinel2_frequency_metadata,
+    load_sentinel2_monthly_metadata,
+)
+
+# TODO: Move these to a .sources folder specific to each data source
+# WARNING: TEMPORARY BANDS: We forgot to pull B9, B10 from the export
```
pbeukema: where is this warning from?

helios/data/collator.py
Hgherzog: totally, plan on changing this to be passed in by a config

helios/data/dataset.py
Hgherzog: agreed

helios/data/collator.py
Hgherzog: I haven't set up logging yet in this repo because we could ideally use olmo-core logging that already natively handles logging only on certain ranks and things like this.

helios/data/dataset.py
Hgherzog: This is from favyen as of this week

.github/workflows/lint.yaml
Hgherzog: gsutil doesn't work with 3.12 so It is simpler for me to just develop in 3.11 always, open to other strategies

convert_List_to_list.py
```diff
@@ -0,0 +1,27 @@
+"""Autogenerated code to convert list[] to list[]."""
```
favyen2: Would be more clear if it said List[] to list[] I think.

Also, I created a test file where I import `from typing import List` and use it `List[int]` and ruff automatically changes it to `list[int]`. So it doesn't seem like this script is needed.

helios/data/collator.py
```diff
@@ -0,0 +1,121 @@
...
+from torch.nn import functional as F
+
+from helios.data.constants import DATA_SOURCE_TO_VARIATION_TYPE
+from helios.data.masking import apply_random_masking
+
+logger = logging.getLogger(__name__)
+
+
+class CollateFnOutput(NamedTuple):
+    """Output of the collate function.
```
favyen2, line 17: Suggestion: name and document this based on what it is (e.g. "CollatedFeatures") rather than based on which function produces it.

helios/data/dataloader.py
```diff
@@ -0,0 +1,249 @@
...
+from olmo_core.utils import roundrobin, threaded_generator
+from upath import UPath
+
+from helios.data.dataset import HeliosDataset
+
+logger = logging.getLogger(__name__)
+
+
+def iter_batched_helios(
+    iterable: Iterable[dict[str, Any]], local_batch_size: int
```
favyen2, line 23: Is there a reason not to use HeliosDataset here instead of the more generic type? Especially if it is intended to use things like input_ids in the future (which would only work for HeliosDataset).

Also the docstring is missing the arguments documentation.

helios/data/dataloader.py
```diff
@@ -0,0 +1,249 @@
...
+
+class HeliosIterableDatasetWrapper(_IterableDatasetWrapper):
+    """Helios iterable dataset wrapper."""
+
+    def __init__(self, data_loader: NumpyDataLoaderBase):
+        """Initialize the iterable dataset wrapper."""
+        super().__init__(data_loader)
+
+    def __iter__(self) -> Iterator[dict]:
+        """Iterate over the dataset."""
```
favyen2: It looks like parts of this are copied from the super-class and some parts may be customized. It would be helpful to comment which parts are copied and which are new (e.g. this will be useful when things are updated upstream).

helios/data/dataloader.py
```diff
@@ -0,0 +1,249 @@
...
+        #         f"got {tuple(x['input_ids'].shape)}"
+        #     )
+        # shape = tuple(x["input_ids"].shape)
+
+    if batch:
+        yield tuple(batch)
+
+
+class HeliosIterableDatasetWrapper(_IterableDatasetWrapper):
+    """Helios iterable dataset wrapper."""
```
favyen2: It would be helpful for these docstrings to indicate how HeliosIterableDatasetWrapper should be used / what it is for / what it might be wrapping / why the wrapping is needed instead of just using HeliosDataset directly. The current docstring does not provide context about the class's function.

helios/data/dataloader.py
```diff
@@ -0,0 +1,249 @@
...
+
```
favyen2, line 64: Is it the case that Helios dataset does not fit the paradigm of NumpyFSLDataset (for fixed sequence length) or NumpyVSLDataset (for variable sequence length) but it still fits with NumpyDatasetBase?

helios/data/dataset.py
```diff
@@ -0,0 +1,135 @@
...
+        """
+        if data_source == "sentinel2":
+            space_bands = S2_BANDS
+        else:
+            raise ValueError(f"Unknown data source: {data_source}")
+        # We will need different ingestion logic for different data sources at this point
+
+        variation_type = DATA_SOURCE_TO_VARIATION_TYPE[data_source]
+        if variation_type == "space_time_varying":
+            with cast(xr.Dataset, rioxarray.open_rasterio(tif_path)) as data:
```
favyen2: Are you using xarray here? It seems like it reads the GeoTIFF file and then immediately casts it back to numpy array. If that is the case, then it is best to drop the xarray / rioxarray and just use rasterio directly. Reading a GeoTIFF with rasterio will yield a numpy array.

helios/data/dataset.py
```diff
@@ -0,0 +1,135 @@
...
+            ), f"{tif_path} has incorrect number of channels {space_bands} \
+                {values.shape[0]=} {len(space_bands)=}"
+            space_time_x = rearrange(
+                values, "(t c) h w -> h w t c", c=len(space_bands), t=int(num_timesteps)
+            )
+            return space_time_x, num_timesteps
+        else:
+            raise NotImplementedError(f"Unknown variation type: {variation_type}")
+
+    def _tif_to_array_with_checks(
```
favyen2, line 95: This function doesn't seem to do anything different from tif_to_array. It catches exception but then raises it again anyway. So I would remove unless there's a different reason to keep it.

helios/data/index.py
```diff
@@ -0,0 +1,156 @@
...
+        frequency_type: Literal["monthly", "freq"],
+    ) -> UPath:
+        """Get the path to the tif file."""
+        return self.root_dir / f"{data_source}_{frequency_type}" / f"{example_id}.tif"
+
+    def get_metadata_for_data_source_in_sample(
+        self, data_source: str, example_id: str, frequency_type: str
+    ) -> dict:
+        """Get the metadata for a sample."""
+        metadata_df = self.freq_metadata_df_dict[data_source]
```
favyen2: does this lookup need to depend on the frequency_type (which should also use the Literal instead of str)? i.e. I think self.monthly_metadata_df_dict should be used if frequency_type is monthly.

helios/data/index.py
```diff
@@ -0,0 +1,156 @@
...
+                data_source, example_id, freq_type
+            )
+        return {
+            "data_source_metadata": data_source_metadata,
+            "data_source_paths": data_source_paths,
+            "sample_metadata": sample_metadata,
+        }
+
+    def get_sample_information_from_example_id_list(
+        self, example_ids: list[str], freq_type: Literal["monthly", "freq"]
```
favyen2: Suggestion: define this Literal as a type variable and use it consistently.

helios/data/index.py
```diff
@@ -0,0 +1,156 @@
...
+        # Using a df as initial ingest due to ease of inspection and manipulation,
+        self.data_index_df = load_data_index(data_index_path)
+        self.example_id_to_sample_metadata_dict = self.data_index_df.set_index(
+            "example_id"
+        ).to_dict("index")
+
+        self.freq_metadata_df_dict = {}
+        self.monthly_metadata_df_dict = {}
+        for data_source in self.data_sources:
+            self.freq_metadata_df_dict[data_source] = (
```
favyen2: I wonder if we can use one function to load all data source metadata instead of having one for each data source and frequency type.

helios/data/index.py
```diff
@@ -0,0 +1,156 @@
...
+            self.freq_example_ids, "freq"
+        )
+
+        self._samples = samples
+
+        self.root_dir = self.data_index_path.parent
+
+    def get_sample_information_from_example_id(
+        self, example_id: str, freq_type: Literal["monthly", "freq"]
+    ) -> dict:
```
favyen2: I think it would be helpful for SampleInformation to be a class instead of dict so it is clear what the attributes (and their types) are.

helios/data/masking.py
```diff
@@ -0,0 +1,175 @@
...
+            ),
+            np.ones(tokens_the_decoder_will_unmask, dtype=np.int_) * 2,
+            np.zeros(
+                tokens_the_encoder_will_encode,
+                dtype=np.int_,
+            ),
+        )
+    )
+    batch_flat_tokens = repeat(flat_tokens, "t -> b t", b=batch_size)
+    # hopefully this will allow for reproducibility, since random is seeded
```
favyen2, line 63: It seems like it should be possible to create a numpy generator once and reuse that globally, instead of seeding a new generator with random.randint each time this function is called.

helios/data/index.py
```diff
@@ -0,0 +1,156 @@
+"""Module for navigating Dataset Index.
```
favyen2: Currently it seems `data/index.py`, `data/schemas.py`, and `data/utils.py` are for parsing the dataset index while the others are for torch dataset/dataloader. Consider re-organizing the modules so that these three files are in their own module that's separate from the torch stuff. Because `index.py`, `schemas.py`, and `utils.py` are deeply coupled, e.g. it is hard to understand `index.py` without referring to `schemas.py` but `dataloader.py` is not relevant.

helios/data/dataloader.py
Hgherzog: Ideally this function should work for any iterable and even a completely different dataset class so I think it makes sense to leave it agnostic. I can change the name.

helios/data/dataloader.py
Hgherzog: Yes

helios/data/dataset.py
Hgherzog: Perhaps it was needed in the original galileo code but i agree it doesn't make sense here

helios/data/dataset.py
Hgherzog: this function eventually should make the decision about what to do if an example or a path is bad and handle it so training can continue smoothly. For now I am just raising an error so it fails obviously. Added in a comment as a todo

helios/data/index.py
Hgherzog: at least for the current two sample metadatas the meaning of the columns changes so I wanted a different pandera model for it. I made a registry to hopefully simplify this logic

helios/data/index.py
Hgherzog: I agree. i think these files would make more sense in the dataset folder as in your PR.

PR #7, opened 2025-01-17
Add code for creating Helios dataset

favyen2:
Here is some starter code for creating the sample Helios dataset. We will need to rewrite most of this as we scale up and do this in many Beaker jobs rather than one machine.

### DIFF
--- a/helios/dataset_creation/make_meta_summary.py
+++ b/helios/dataset_creation/make_meta_summary.py
@@ -5,6 +5,44 @@ import csv
 
 from upath import UPath
 
+from .util import get_modality_temp_meta_dir
+
+
+def make_meta_summary(helios_path: UPath, modality: str) -> None:
+    """Create the concatenated metadata file for the specified modality.
+
+    The data files and per-example temporary metadata files must be populated for the
+    modality already. This function just concatenates those temporary metadata files
+    together into one big CSV.
+
+    Args:
+        helios_path: Helios dataset path.
+        modality: modality to write summary for.
+    """
+    # Concatenate the CSVs while keeping the header only from the first file that we
+    # read.
+    column_names: list[str] | None = None
+    csv_rows = []
+    meta_dir = get_modality_temp_meta_dir(helios_path, modality)
+    for fname in meta_dir.iterdir():
+        with fname.open() as f:
+            reader = csv.DictReader(f)
+            if reader.fieldnames is None:
+                raise ValueError(f"CSV at {fname} does not contain header")
+            if column_names is None:
+                column_names = list(reader.fieldnames)
+            for csv_row in reader:
+                csv_rows.append(csv_row)
+
+    if column_names is None:
+        raise ValueError(f"did not find any files in {meta_dir}")
+
+    with (helios_path / f"{modality}.csv").open("w") as f:
+        writer = csv.DictWriter(f, fieldnames=column_names)
+        writer.writeheader()
+        writer.writerows(csv_rows)
+
+
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(
         description="Create concatenated metadata file",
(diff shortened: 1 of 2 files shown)


data/rslearn_dataset_configs/config_landsat.json
```diff
@@ -0,0 +1,431 @@
+{
```
Hgherzog, line 1: How do we want track and organize these configs for different versions of the dataset as we grow? A separate folder of configs per dataset? What if we want to pull more landsat data when we already have landsat data

helios/dataset/create_windows/naip.py
```diff
@@ -0,0 +1,120 @@
...
+    query_config = QueryConfig()
+
+    for lon_base, lat_base in LOCATIONS:
+        for lon_offset in [-0.03, 0, 0.03]:
+            for lat_offset in [-0.03, 0, 0.03]:
+                lon = lon_base + lon_offset
+                lat = lat_base + lat_offset
+                window = create_window_naip_time(ds_path, lon, lat)
+                if window is None:
+                    print(f"warning: did not find any NAIP image at {lon}, {lat}")
```
Hgherzog: log instead of print, if we intend to keep this long term, otherwise don't care. We should be checking for print in pre-commit though

helios/dataset/make_index.py
```diff
@@ -0,0 +1,66 @@
...
+    for modality in MODALITIES:
+        for fname in (helios_path / modality).iterdir():
+            example_id = fname.name.split(".")[0]
+            if example_id not in modalities_by_example:
+                modalities_by_example[example_id] = []
+            modalities_by_example[example_id].append(modality)
+
+    # Now we can write the CSV.
+    with (helios_path / "index.csv").open("w") as f:
+        column_names = [
```
Hgherzog: Suggestion: column names of the index should be a constant

helios/dataset/make_index.py
```diff
@@ -0,0 +1,66 @@
...
+        type=str,
+        help="Helios dataset path",
+        required=True,
+    )
+    args = parser.parse_args()
+
+    helios_path = UPath(args.helios_path)
+
+    # List the available modalities for each example.
+    modalities_by_example: dict[str, list[str]] = {}
```
Hgherzog: Suggestion: wrap this into a function for testability and add a test for this

helios/dataset/make_meta_summary.py
```diff
@@ -0,0 +1,49 @@
...
+    )
+    parser.add_argument(
+        "--modality",
+        type=str,
+        help="Modality to summarize",
+        required=True,
+    )
+    args = parser.parse_args()
+
+    helios_path = UPath(args.helios_path)
```
Hgherzog: Suggestion: wrap the logic into a function and add a test.

helios/dataset/rslearn_to_helios/openstreetmap.py
```diff
@@ -0,0 +1,99 @@
...
+    window = Window.load(window_path)
+    vector_format = GeojsonVectorFormat(coordinate_mode=GeojsonCoordinateMode.CRS)
+
+    if not window.is_layer_completed(LAYER_NAME):
+        return
+
+    layer_dir = window.get_layer_dir(LAYER_NAME)
+    features = vector_format.decode_vector(layer_dir, window.bounds)
+    vector_format.encode_to_file(
+        fname=helios_path / "openstreetmap" / f"{window.name}.geojson",
```
Hgherzog: Suggestion: add functions that encode the new directory structure so that it can easily stay consistent across data sources

data/rslearn_dataset_configs/config_landsat.json
favyen2: For now I think we can stick with one version for each modality in any one branch of the git repository.

If we want to pull more data we should be able to stick with the same rslearn dataset configuration. If we change the configuration then that may mean that we need to re-pull all of the data e.g. if we add a new band that wasn't there before or something like that, so we may not need to keep around the old version anyway.

PR #9, opened 2025-01-23
Henryh/integrate naip osm wc

Hgherzog:
- enable loading naip, worldcover, and osm
- add per modality collate function
- deduplicate and reorganize some constants

### DIFF
--- a/helios/constants.py
+++ b/helios/constants.py
@@ -1,5 +1,7 @@
 """Constants shared across the helios package."""
 
+from typing import Literal
+
 NAIP_BANDS = ["R", "G", "B", "IR"]
 
 S2_BANDS = [
@@ -20,7 +22,13 @@ S2_BANDS = [
 
 WORLDCOVER_BANDS = ["B1"]
 
+VARIATION_TYPES = Literal[
+    "space_time_varying", "time_varying", "space_varying", "static"
+]
 
 DATA_SOURCE_TO_VARIATION_TYPE = {
     "sentinel2": "space_time_varying",
+    "worldcover": "space_varying",
+    "naip": "space_varying",
+    "openstreetmap": "space_varying",  # How do we want to use open Streetmap
 }
--- a/helios/data/masking.py
+++ b/helios/data/masking.py
@@ -137,6 +137,10 @@ def apply_random_masking(
     )
 
     # get the space mask tokens
+    print(token_mask.shape)
+    print(num_space_time_tokens)
+    print(num_time_tokens)
+    print(num_static_tokens)
     space_tokens = token_mask[
         :, num_space_time_tokens : -(num_time_tokens + num_static_tokens)
     ]
--- a/helios/dataset/index.py
+++ b/helios/dataset/index.py
@@ -57,7 +57,10 @@ class DatasetIndexParser:
         "worldcover": "tif",
     }
 
-    def __init__(self, data_index_path: UPath | str):
+    def __init__(
+        self,
+        data_index_path: UPath | str,
+    ):
         """Initialize the dataset index parser."""
         self.data_source_and_freq_types = self._get_data_sources_and_freq_types()
         self.data_index_path = UPath(data_index_path)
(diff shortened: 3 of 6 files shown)


helios/constants.py
```diff
@@ -0,0 +1,34 @@
...
+VARIATION_TYPES = Literal[
+    "space_time_varying", "time_varying", "space_varying", "static"
+]
+
+DATA_SOURCE_TO_VARIATION_TYPE = {
+    "sentinel2": "space_time_varying",
+    "worldcover": "space_varying",
+    "naip": "space_varying",
+    "openstreetmap": "space_varying",  # How do we want to use open Streetmap
+}
```
yawenzzzz, line 34: We don't need this anymore since they're treated as separate modes, and no need to group by their characteristics, as we can apply different patch size (shape) to them.

helios/data/masking.py
```diff
@@ -137,6 +137,10 @@ def apply_random_masking(
     )
 
     # get the space mask tokens
+    print(token_mask.shape)
+    print(num_space_time_tokens)
+    print(num_time_tokens)
+    print(num_static_tokens)
```
yawenzzzz, line 143: I'm not sure if we will stick to this group definition as we expand to more modalities, we can mask each mode separately instead of grouping them. The grouping makes more sense for small tile modeling, and we would like the input tile size to be flexible.

helios/dataset/index.py
```diff
@@ -90,14 +94,76 @@ def __init__(self, data_index_path: UPath | str):
...
+            # No Frequency Type means static data source
+            freq_type = metadata.get("frequency_type", None)
+            data_source = (
+                column_name.replace(f"_{freq_type}", "") if freq_type else column_name
+            )
+            logger.debug(f"Adding {data_source} {freq_type}")
+            data_source_columns_and_freq_types.append((data_source, freq_type))
+        return data_source_columns_and_freq_types
+
+    def _load_all_data_source_metadata(
```
yawenzzzz, line 125: This is OK for now, but later, we would like to load more metadata from the data source, like spatial resolution, temporal steps, etc.

helios/dataset/index.py
```diff
@@ -30,51 +34,51 @@ class SampleInformation(NamedTuple):
...
+        "worldcover": "tif",
+    }
+
+    def __init__(
+        self,
+        data_index_path: UPath | str,
+    ):
         """Initialize the dataset index parser."""
-        self.data_sources = ALL_DATA_SOURCES
+        self.data_source_and_freq_types = self._get_data_sources_and_freq_types()
```
yawenzzzz, line 65: Frequency type will only impact the number of bands in the geotiff, as long as the number of bands is fixed, we don't need to differentiate the frequency type, by reading the sampler, it will tell the temporal length, and the padding will be based on the maximum length in the sample.

helios/dataset/index.py
```diff
@@ -30,51 +34,51 @@ class SampleInformation(NamedTuple):
...
-        # Intersect available data sources with index column names
-
         assert (
-            len(self.data_sources) > 0
+            len(self.data_source_and_freq_types) > 0
         ), "No data sources found in index, check naming of columns"
+        all_metadata = self._load_all_data_source_metadata()
+        self.static_metadata_df_dict = all_metadata.static
+        self.freq_metadata_df_dict = all_metadata.freq
+        self.monthly_metadata_df_dict = all_metadata.monthly
```
yawenzzzz, line 80: Similar comment as above, we can have two separate pretraining dataset, one is at 2-week granularity, the other at monthly, and mix them for pretraining.

helios/data/masking.py
Hgherzog: yea this masking is based on the presto-v3 codebase likely not needed here anymore

helios/dataset/index.py
Hgherzog: I am loading all metadata that is included in the metadata csv and all the other columns for each sample. If there is something missing we should also put it in the metadata

helios/dataset/index.py
Hgherzog: The point of this is that we didn't want sentinel-2_freq and sentinel-2_month to be in the same sample so we are parsing that here so each sample contains either monthly + single timestep or freq + single timestep. Do you think a sample should be defined differently

helios/dataset/index.py
Hgherzog: What is happening here is I am essentially having. a set of indices at a 2 week and a set of indices at a monthly. So, when we sample rnadomly we get a mixed dataset

PR #10, opened 2025-01-24
Add Landsat data

favyen2:
Adds Landsat data and restructures the code since Sentinel-2 and Landsat (and I suspect other multi-temporal ones like Sentinel-1) look quite similar.


helios/dataset_creation/rslearn_to_helios/landsat.py
```diff
@@ -0,0 +1,93 @@
...
+import argparse
+import multiprocessing
+
+import tqdm
+from rslearn.utils.mp import star_imap_unordered
+from upath import UPath
+
+from .multitemporal_raster import convert_freq, convert_monthly
+
+BANDS = [
```
Hgherzog, line 12: Landsat bands will probably be needed as a constant across the entire repo so i think we should move those to the main constants.py

helios/dataset_creation/const.py
Hgherzog, line 1: nit: we use constants.py and const.py in different places, I don't care which we use but we should pick 1

helios/dataset_creation/rslearn_to_helios/sentinel2.py
```diff
@@ -1,22 +1,35 @@
...
 
-import numpy as np
 import tqdm
-from rslearn.data_sources import Item
-from rslearn.dataset import Window
 from rslearn.utils.mp import star_imap_unordered
-from rslearn.utils.raster_format import GeotiffRasterFormat
 from upath import UPath
 
-from helios.constants import S2_BANDS
```
Hgherzog, line 16: Is there a reason you think these bands constants shouldn't be shared? If not prefer putting these in library wide constants

favyen2: This was from merge conflict and I resolved by keeping new change since it will change anyway in #37.

I think it is better to not share so that we can update dataset creation and then later update data loader instead of needing to update together.

helios/dataset_creation/const.py
favyen2: I will change to constants.py in #37

helios/dataset_creation/rslearn_to_helios/sentinel2.py
favyen2: Actually I think it's fine to change it back to use helios.constants once the rest of the helios code is updated but for #37 it made sense to change it back to a local version of the constants since it is now list of BandSet to include the resolution.

PR #31, opened 2025-01-30
v2 setup for olmo core

Hgherzog:


helios/data/dataloader.py
```diff
@@ -283,6 +282,25 @@ def _iter_batches(self) -> Iterable[dict[str, Any]]:
...
+            ),
+            "worldcover": torch.rand(
+                (self.global_batch_size, 1, 256, 1), dtype=torch.float32
+            ),
+            "sample_metadata": dict(
+                sample_id=torch.rand((self.global_batch_size,)),
+                num_timesteps=torch.rand((self.global_batch_size,)),
+            ),
+        }
+
```
gabrieltseng, line 303: What are the dims here? In general, I think it would be good for us to agree on some format that the dataloaders will return, and that the models will ingest (this would also make it easier to write evals that don't break every time we change the model architecture).

Here's an example, obviously imperfect for our use case but still informative I think: https://github.com/WorldCereal/prometheo/blob/main/prometheo/predictors.py

helios/data/dataloader.py
```diff
@@ -283,6 +282,25 @@ def _iter_batches(self) -> Iterable[dict[str, Any]]:
...
 
+    def get_mock_batch(self) -> dict[str, Any]:
+        """Get a mock batch."""
+        # TODO: warning assign dtypes like this is bad and w probaby should have it be a shared state
+        return {
+            "sentinel2": torch.rand(
+                (self.global_batch_size, 256, 256, 12, 13), dtype=torch.float32
+            ),
+            "naip": torch.rand(
+                (self.global_batch_size, 24, 256, 4), dtype=torch.float32
```
gabrieltseng: Should this be `(self.global_batch_size, 256, 256, 4)`? 

(or `(self.global_batch_size, 2560, 2560, 4)` if its a consistent tile size with S2?

helios/data/dataloader.py
Hgherzog: 100% agreed, favyen reformatted the dataset a bit and yawen is planning on re doing the data loader so we should likely agree soon. I like using Named tuple with an as_dict method

helios/data/dataloader.py
Hgherzog: Naip should be consistent, I think how we keep the tile size consistent is changing with the new dataset format as we are not resampling to the same resolution. for now it should be 256, 256, 4

yawenzzzz: Yes, it's 256 * 256 * 4 for now, Favyen has changed the dataset design: https://docs.google.com/document/d/1qzT9sSwhaCIMlZkKDnaw4EhddLLhxu91ixY7IWoUGSA/edit?tab=t.el4oqc57bsv4, which updates the resolution of each modality, and it will keep NAIP as 256 * 256.

helios/data/dataloader.py
yawenzzzz: +1, since we just updated the dataset format this week, I'll adjust the dataloader accordingly and ensure we align on its outputs.

helios/data/collator.py
```diff
@@ -74,9 +74,9 @@ def per_modality_collate_fn(items: Sequence[dict]) -> PerModalityCollatedOutput:
...
-        naip=naip_batch,
-        worldcover=worldcover_batch,
-        sample_metadata=sample_metadata,
-    )
+    return {
+        "sentinel2": sentinel2_batch,
+        "naip": naip_batch,
+        "worldcover": worldcover_batch,
+        "sample_metadata": sample_metadata,
+    }
```
yawenzzzz, line 82: The returned dict looks good, we may also want to add other keys, like sentinel2_dates, which include timestamps information, can be useful in positional embedding.

helios/train/train_module.py
yawenzzzz, line 1: This TrainModule in v2 looks great! Once the batch is ready, train_batch / eval_batch / model_forward can be easily updated.

PR #34, opened 2025-01-31
Evals - add EuroSat

gabrieltseng:


helios/evals/datasets/geobench/dataset.py
```diff
@@ -0,0 +1,205 @@
...
+                        # we have a match!
+                        band_idx = names_list.index(src)
+                        new_image_list.append(image_list[band_idx])
+                        break
+        return new_image_list
+
+    @staticmethod
+    def normalize_bands(
+        image: np.ndarray, means: np.array, stds: np.array, method: str = "norm_no_clip"
+    ):
```
gabrieltseng: I think its obvious this shouldn't live here. Two thoughts:

1. Do we want to couple eval and pre-training normalization? (pros - probably good to share code, cons - eval normalization code will be a bit more complicated (probably) since it needs to handle more cases)
2. Any thoughts on which normalization we should use? Galileo and CROMA both used `norm_no_clip` so its probably a safe bet

tests/fixtures/sample_geobench/classification_v1.0/m-eurosat/band_stats.json
yawenzzzz, line 1: Is this directly computed from the EuroSAT dataset?

helios/evals/embeddings.py
```diff
@@ -0,0 +1,32 @@
...
+    model = model.eval()
+    with torch.no_grad():
+        for batch in data_loader:
+            batch_labels = batch.pop("target")
+            if "s1" in batch:
+                batch["s1"] = batch["s1"].to(device).to(torch.bfloat16)
+            if "s2" in batch:
+                batch["s2"] = batch["s2"].to(device).to(torch.bfloat16)
+            if "months" in batch:
+                batch["months"] = batch["months"].to(device).long()
```
yawenzzzz: Do you think we need to include dates for each modality in the batch: https://github.com/allenai/helios/pull/31/files#diff-969cc9076a122938877a8fcf4b5e5206884bd336e2b02f80572020181f00e3b5R77-R82.

helios/evals/datasets/geobench/dataset.py
Hgherzog: I agree that for any preprocessing, and normalization that should be the same should share code.

helios/evals/datasets/geobench_dataset.py
```diff
@@ -0,0 +1,268 @@
...
+
+        # check if label is an object or a number
+        if not (isinstance(label, int) or isinstance(label, list)):
+            label = label.data
+            # label is a memoryview object, convert it to a list, and then to a numpy array
+            label = np.array(list(label))
+
+        target = torch.tensor(label, dtype=torch.long)
+        s2 = repeat(x, "h w c -> c t h w", t=1)[GEOBENCH_TO_HELIOS_S2_BANDS, :, :, :]
+        return HeliosSample(s2=s2), target
```
yawenzzzz, line 252: Shall we also add latlon and timestamps? https://github.com/allenai/helios/blob/ec93f9f553df3f6b0b0abbe818b334b011d2bb1d/helios/data/dataset.py#L20

helios/olmo_core_proto.py
```diff
@@ -185,4 +185,37 @@ def patch_disc_loss(
...
+
+    # eval. Currently this will fail because by default our model ingests 4 timesteps.
+    # we should update the model architecture to ingest variable numbers of timesteps
+    from torch.utils.data import DataLoader
+
+    from helios.evals.datasets import GeobenchDataset
+    from helios.evals.embeddings import get_embeddings
+    from helios.evals.knn import run_knn
+
+    geobench_dir = UPath("/weka/skylight-default/presto-geobench/dataset/geobench")
```
yawenzzzz, line 197: We have a new WEKA folder: `/weka/dfive-default/`, datasets can be uploaded: https://weka.allen.ai/b/dfive-default/.

tests/unit/eval/test_geobench_dataset.py
```diff
@@ -0,0 +1,46 @@
...
+    """Test the dataset works."""
+    d = GeobenchDataset(
+        dataset="m-eurosat",
+        geobench_dir=geobench_dir,
+        split="train",
+        partition="0.01x_train",
+    )
+    sample, _ = d[0]
+    assert isinstance(sample.s2, torch.Tensor)
+    assert sample.s2.shape == (13, 1, 64, 64)
```
yawenzzzz, line 28: The pretraining S2 images are 256×256. We may need to crop them into diverse tile sizes to better adapt to downstream tasks.

helios/evals/datasets/geobench_dataset.py
```diff
@@ -0,0 +1,268 @@
...
+                )
+        return image
+
+    def __getitem__(self, idx: int) -> tuple[HeliosSample, torch.Tensor]:
+        """Return a single GeoBench data instance."""
+        label = self.dataset[idx].label
+
+        x_list = []
+        for band_idx in self.band_indices:
+            x_list.append(self.dataset[idx].bands[band_idx].data)
```
yawenzzzz: use `sample = self.dataset[idx]` to avoid repeated lookup of the same item, also, it's clearer

helios/evals/datasets/geobench_dataset.py
```diff
@@ -0,0 +1,268 @@
...
+                image = (
+                    image.astype(original_dtype) / 255
+                )  # back to original_dtype between 0 and 1
+            elif method == "norm_no_clip":
+                pass
+            else:
+                raise ValueError(
+                    f"norm type must be norm_yes_clip, norm_yes_clip_int, norm_no_clip, or standardize, not {method}"
+                )
+        return image
```
yawenzzzz, line 220: It's good for now, but a suggestion will be consolidating norm methods into a dict that maps each method to a function. This would make it easier to add new methods and avoid the if-elif checks.

tests/fixtures/sample_geobench/classification_v1.0/m-eurosat/band_stats.json
gabrieltseng: yes

helios/evals/embeddings.py
gabrieltseng: Not with the current approach: https://github.com/allenai/helios/blob/aaf3868c26f5e038a19c1b5cd5171a6f4306c59d/helios/data/dataset.py#L20

helios/evals/datasets/geobench_dataset.py
gabrieltseng: We don't have them for these samples. EuroSat doesn't share their locations (at least not in this geobench version).

helios/olmo_core_proto.py
gabrieltseng: awesome. For testing, I might keep this as is (since the data is already there) but lets add future eval data to dfive. I'll add a todo to move the data over there too.

tests/unit/eval/test_geobench_dataset.py
gabrieltseng: Yes, this is the plan! Vary the patch sizes and the number of input spatial and temporal patches.

helios/evals/datasets/geobench_dataset.py
gabrieltseng: Updated

helios/evals/datasets/geobench_dataset.py
gabrieltseng: Maybe once we have this PR and the dataloader PR merged, we can have someone do consistent normalization for all of them in a single place?

https://github.com/allenai/helios/pull/34#discussion_r1937615898

Issue #25, opened 2025-01-29
Revisit Helios data loader


yawenzzzz:
Here's the current implementation using OLMo-core: https://github.com/allenai/helios/blob/main/helios/data/dataloader.py. 

Another option is dataloader specifically for RS data, in particular multimodal RS data, like: 

Presto: https://github.com/nasaharvest/presto-v3/blob/main/src/data/dataset.py
AnySat: https://github.com/gastruc/AnySat/blob/main/src/data/S2Naip.py

This is related to Issue #16. Need a revisit.

Issue #16, opened 2025-01-27
Revise Helios dataset format


yawenzzzz:
Here's the most recent Helios dataset: https://console.cloud.google.com/storage/browser/ai2-helios/data/20250124-sample-dataset-helios?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&hl=en&invt=AbmiOA&project=skylight-proto-1

PR #36, opened 2025-01-31
Update dataset and dataloader

yawenzzzz:
From Favyen:

This should contain NAIP, Sentinel-2, Sentinel-1, and WorldCover data in [new format](https://docs.google.com/document/d/1qzT9sSwhaCIMlZkKDnaw4EhddLLhxu91ixY7IWoUGSA/edit?tab=t.el4oqc57bsv4).
gs://ai2-helios/data/20250130-sample-dataset-helios/

Issue #16 #25

### DIFF
--- a/helios/data/dataloader.py
+++ b/helios/data/dataloader.py
@@ -158,7 +158,7 @@ class HeliosDataLoader(DataLoaderBase):
             )
         return np.memmap(self._global_indices_file, mode="r", dtype=np.uint32)
 
-    def _iter_batches(self) -> Iterable[dict[str, Any]]:
+    def _iter_batches(self) -> Iterable[HeliosSample]:
         """Iterate over the dataset in batches."""
         return torch.utils.data.DataLoader(
             _IterableDatasetWrapper(self),


README.md
```diff
@@ -1,32 +1,46 @@
...
 
-1. Create a virtual environment in prefered directory with python 3.12 `python3 -m venv .venv-helios` \
-2. Navigate to root directory of this repo and run `pip install -e .`
-3. Run `pre-commit install`
+1. Create a virtual environment in prefered directory with python 3.10 `python3.10 -m venv .venv-helios`
+2. Activate the virtual environment `source .venv-helios/bin/activate`
+3. Navigate to root directory of this repo and run `pip install -e .`
+4. Run `pre-commit install`
+
+Note: Python 3.12 may be incompatible with `ai2-olmo-core==1.8.0`.
```
Hgherzog: what about with the v2 branch what is the python compatibility? we are no longer using pip installable version

helios/constants.py
```diff
@@ -2,6 +2,7 @@
 
 from typing import Literal
 
+# TODO: data related constants should be moved to the data package.
```
Hgherzog: Bands constants likely will also be needed in train to form channel group constants so not sure if these need to move.

helios/data/constants.py
```diff
@@ -0,0 +1,182 @@
...
+    this in the future to avoid the extra code here.
+    """
+    resolution = BASE_RESOLUTION * resolution_factor
+    if float(int(resolution)) == resolution:
+        return int(resolution)
+    return resolution
+
+
+@dataclass(frozen=True)
+class BandSet:
```
Hgherzog, line 33: I seem to remember this also being in one of favyens PRs, just double chekcing this class isn't duplicated

helios/data/constants.py
```diff
@@ -0,0 +1,182 @@
...
+# Everything else is a factor (which is a power of 2) coarser than this resolution.
+BASE_RESOLUTION = 0.625
+
+# The default image tile size.
+# Some images may be smaller if they are stored at a coarser resolution compared to the
+# resolution that the grid is based on.
+IMAGE_TILE_SIZE = 256
+
+
+def get_resolution(resolution_factor: int) -> float | int:
```
Hgherzog, line 19: Suggestion prefer placing this in data/utils.py rather than constant

helios/data/constants.py
```diff
@@ -0,0 +1,182 @@
...
+        """Hash this ModalitySpec."""
+        return hash(self.name)
+
+    def get_tile_resolution(self) -> float:
+        """Compute the tile resolution."""
+        return get_resolution(self.tile_resolution_factor)
+
+
+# Modalities supported by helios
+class Modality:
```
Hgherzog, line 82: This is a good abstraction.

helios/data/constants.py
```diff
@@ -0,0 +1,182 @@
...
+                    "works",
+                ],
+                4,
+            )
+        ],
+        is_multitemporal=False,
+    )
+
+
+ALL_MODALITIES = [
```
Hgherzog: SUggestion: Make this as as a classmethod  rather than a seperate constant

helios/data/dataset.py
```diff
@@ -113,100 +120,124 @@ def w(self) -> int:
         return self.s2.shape[-1]
 
 
-class HeliosDataset(NumpyDatasetBase, Dataset):
+class HeliosDataset(Dataset):
```
Hgherzog, line 142: It seems like perhaps there should be a not language model specific base class to numpyDataset base, we should probably take note of what is needed for that and share it with pete

helios/data/dataset.py
```diff
@@ -113,100 +120,124 @@ def w(self) -> int:
...
-                data_input = data_input.astype(self.dtype)
-            data_inputs[data_source] = data_input
-
-        return {
-            "data_inputs": data_inputs,
-            "sample_metadata": sample.sample_metadata,
-            "num_timesteps": max_num_timesteps,
-            "data_source_metadata": sample.data_source_metadata,
-        }
+        sample = self.samples[index]
```
Hgherzog, line 240: We can keep it as S2 only for now but we should add an issue or todo that this has to be updated to support arbitrary data

helios/data/dataset.py
```diff
@@ -113,100 +120,124 @@ def w(self) -> int:
...
+        sample_s2 = sample.modalities[Modality.S2]
+        timestamps = [i.start_time for i in sample_s2.images]
+        image = load_image_for_sample(sample_s2, sample)
+        s2_data = image.transpose(1, 0, 2, 3)  # [T, C, H, W] -> [C, T, H, W]
```
Hgherzog: suggestion: we have been using einops rearrange for ops like this and I think it communicates intent via code more readably than transpose + comment

helios/data/dataloader.py
```diff
@@ -4,21 +4,258 @@
...
+            indices = indices[worker_info.id :: worker_info.num_workers]
+
+        # Finally slice batches into micro batches for the local DP rank.
+        indices = indices[:, self.dp_rank :: self.dp_world_size].reshape((-1,))
+        return indices
+
+    def _get_dataset_item(self, idx: int) -> dict[str, Any]:
+        """Get a dataset item."""
+        item = self.dataset[idx]
+        if isinstance(item, dict):
```
Hgherzog: HeliosDataset get item returns a Named tuple rather than a dict. Is this still compatible

helios/data/dataloader.py
```diff
@@ -4,21 +4,258 @@
...
+
+    def _format_fname_from_fields(self, prefix: str, **fields: Any) -> str:
+        parts = [prefix]
+        for key in sorted(fields):
+            value = fields[key]
+            if value is not None:
+                parts.append(f"{key}{value}")
+        return "_".join(parts)
+
+    def get_mock_batch(self) -> dict[str, Any]:
```
Hgherzog: should mock batch be a dictionary or still a named tuple?

README.md
yawenzzzz: Just reproduced the error, seems more related to the torch version, with Python 3.12, I hit this error, with Python 3.10, the installation is smooth. I just cloned the v2 branch, and run `pip install -e .`. 
Screenshot 2025-02-07 at 9 30 36 AM

helios/constants.py
yawenzzzz: make sense, removed it. Given this, it may be better to have train related constants together.

yawenzzzz: Also, since the `SampleInformation` already carry all the bands info, we can later modify `__getitem__` to include more information.

helios/data/constants.py
yawenzzzz: Yes, this was added by Favyen, all in one PR.

helios/data/constants.py
yawenzzzz: I'm not sure if we want to move this, `BandSet` and `ModalitySpec` all directly use this function, and it provides more context here. We can move it later if there's more data related utils.

helios/data/constants.py
yawenzzzz: good point, I modified this into a classmethod.

README.md
Hgherzog: I tried with 3.11.5 and it definitely works with that. I am going to try to repro 3.12

helios/data/dataset.py
yawenzzzz: Good idea!

helios/data/dataset.py
yawenzzzz: The arbitrary data will be supported in a very similar way, as long as our model accept varying C and T and H and W. Will add issue for this, now is mainly for end-to-end testing purpose.

helios/data/dataset.py
yawenzzzz: Fixed.

README.md
Hgherzog: I 
Screenshot 2025-02-07 at 9 51 12 AM
was able to build it with 3.12

Hgherzog: maybe it is due to something being set in your base conda env

helios/data/dataloader.py
yawenzzzz: I updated it to return HeliosSample.

helios/data/dataloader.py
yawenzzzz: The dict is due to using default_collate from pytorch, I updated the collate fn to process batches into a HeliosSample, so the return is still a named tuple.

Issue #28, opened 2025-01-29
Convert vector data into tile


yawenzzzz:
Vector data like OpenStreetMap (points, lines, polygons) could be further converted into raster tile for SSL training.

PR #37, opened 2025-01-31
New dataset format

### LINKED ISSUE #16: Revise Helios dataset format

favyen2:
* Changes dataset format to the one described at https://docs.google.com/document/d/1qzT9sSwhaCIMlZkKDnaw4EhddLLhxu91ixY7IWoUGSA/edit?tab=t.el4oqc57bsv4
* I changed the code back to keep constants separate between the dataset creation code and the rest of the codebase. IT will be easier to merge in updated dataset creation code before merging in updates to the rest, so some degree of separation here is useful, so that things can continue working with the old dataset outside the dataset_creation module.
* Add Sentinel-1 and rasterized OpenStreetMap.

Resolves #16 and #28.

Depends on #10 being merged first.

### DIFF
--- a/helios/dataset_creation/constants.py
+++ b/helios/dataset_creation/constants.py
@@ -2,19 +2,6 @@
 
 from rslearn.utils.raster_format import GeotiffRasterFormat
 
-MODALITIES = [
-    "landsat_freq",
-    "landsat_monthly",
-    "naip",
-    "openstreetmap",
-    "openstreetmap_raster",
-    "sentinel1_freq",
-    "sentinel1_monthly",
-    "sentinel2_freq",
-    "sentinel2_monthly",
-    "worldcover",
-]
-
 # Columns in the per-modality metadata CSVs.
 METADATA_COLUMNS = [
     "crs",


data/rslearn_dataset_configs/config_sentinel1.json
yawenzzzz, line 1: Hmm, I’m wondering if there’s a better way to structure the data config for this 1-year dataset, like using a base layer and modifying only `layer_name` and `time_offset` instead of repeating most of the content. But I think this is also good since it’s reusable for any timestamp requiring S1 data.

helios/dataset_creation/rslearn_to_helios/openstreetmap.py
```diff
@@ -22,7 +23,11 @@
 LAYER_NAME = "openstreetmap"
 
 # Modality in the output Helios dataset.
-MODALITY = "openstreetmap"
+MODALITY = "10_openstreetmap"
```
yawenzzzz, line 35: Since OSM vectors are saved with a CRS resolution of 0.625m, why is the prefix for this modality 10_? a little confused. Is it because it still follows the 10m resolution GridTile?

helios/dataset_creation/rslearn_to_helios/rasterize_openstreetmap.py
yawenzzzz, line 1: This is really cool! I only conceptually know rasterizing vectors, it’s great to see it how it works. Just a minor thought, should some lines, like highways, be thicker? Also, I’m not sure what is the best way to represent points as different points may have varying radius. It’s fine to keep them for now.

helios/dataset_creation/constants.py
```diff
@@ -0,0 +1,32 @@
...
+    "landsat_monthly",
+    "naip",
+    "openstreetmap",
+    "openstreetmap_raster",
+    "sentinel1_freq",
+    "sentinel1_monthly",
+    "sentinel2_freq",
+    "sentinel2_monthly",
+    "worldcover",
+]
```
yawenzzzz: We can either drop this as there's no more index file or update with resolution prefix, both are fine.

data/rslearn_dataset_configs/config_sentinel1.json
favyen2: I think we could use something similar to https://github.com/allenai/rslearn_projects/blob/ec95814eabcd0e730694d69d668f5731075bbd33/rslp/satlas/data_sources.py#L62 to select the monthly mosaics rather than creating a separate layer for each one.

Ideally there would just be one layer for one-year data instead of one per month.

I will add issue to change this in the future since it is difficult to do it now (e.g. the PR is not merged and helios currently does not have rslp as dependency).

helios/dataset_creation/rslearn_to_helios/openstreetmap.py
favyen2: Added comment to this file about it -- the reason to use the 10 m/pixel grid is because the data is actually small in size so we can afford to store it at all of the 10 m/pixel tiles rather than only the 0.625 m/pixel tiles (which likely would only cover a portion). That way we can also use this data during training at 10 m/pixel resolution instead of only 0.625 m/pixel resolution.

It is vector data so the resolution only impacts the precision of the coordinates, it doesn't really change the data size much.

So it is stored at 0.625 m/pixel to retain the precision of the coordinates. I guess it may not be necessary if it uses floating point to store the coordinates, I thought it might cast to integer coordinate.

For now I added comment to make it more clear.

helios/dataset_creation/rslearn_to_helios/rasterize_openstreetmap.py
favyen2: Yeah for now I just did the lines and points as a single pixel. I think how we rasterize it depends on how we're training -- if we are training with segmentation then a bigger radius may be better (although the model likely would predict high probabilities near the line/point anyway), but if we use I-JEPA or something like that it may not be necessary.

helios/dataset_creation/constants.py
favyen2: Removed. I will also switch back to using helios.data.constants in a future PR.

PR #38, opened 2025-01-31
Henryh/setup ddp

Hgherzog:
set up for ddp and also additional clean ups and documentations, as well as starting to use typical olmo core style configs

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,10 +10,17 @@ earth system foundation model data, training, and eval
 3. Run `pre-commit install`
 
 ## Setup Instructions for running olmo_core_proto.py
-1. Clone the [Olmo-core](https://github.com/allenai/OLMo-core/tree/v2) repo and switch to the v2 branch `git checkout v2`
+1. Clone the [Olmo-core](https://github.com/allenai/OLMo-core/tree/v2) repo and switch to the v2 branch
+    ```bash
+    git clone --branch v2 https://github.com/allenai/OLMo-core.git
+    cd OLMo-core
+    ```
 2. Navigate to the root directory of olmo-core repository and run `pip install -e .`
 3. (Skip if dataset is on weka) Make sure you have access to the relevant bucket 'gcloud auth default login' or using beaker secrets
 4. Set `WANDB_API_KEY' api key environment variable (or povide it via --secret-env flag when you start your beaker session)
+    ```bash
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -3,8 +3,23 @@ Highly Efficient Learning for Integrated Observation Systems (HELIOS)
...
 
-Setup
+1. Create a virtual environment in prefered directory with python 3.12 `python3 -m venv .venv-helios` \
+2. Navigate to root directory of this repo and run `pip install -e .`
+3. Run `pre-commit install`
 
-1. pip install -e .
-2. Log into gcloud
+## Setup Instructions for running olmo_core_proto.py
+1. Clone the [Olmo-core](https://github.com/allenai/OLMo-core/tree/v2) repo and switch to the v2 branch `git checkout v2`
```
pbeukema: provide the commands so they can be directly copy pasted
```
git clone --branch v2 https://github.com/allenai/OLMo-core.git
cd OLMo-core
```

helios/olmo_core_proto.py
```diff
@@ -1,57 +1,69 @@
...
-
-## Config does not yet support our new dataset type so we will construct manually for now
-
+# THings that need a config
+# Data Loader
+# Model
 
 if __name__ == "__main__":
-    setup_logging()
+    # Variables to be changed per user
```
pbeukema, line 36: does it make sense to switch this to argparse?

helios/olmo_core_proto.py
```diff
@@ -69,52 +81,60 @@
...
+    train_module = train_module_config.build(model=model)
+    dp_process_group = train_module.dp_process_group
+    dataloader = HeliosDataLoader.wrap_numpy_dataset(
+        dataset=HeliosDataset(
+            *samples,
+            ignore_data_sources=["openstreetmap"],
+            filter_samples_with_missing_inputs=True,
+            dtype=np.dtype("float32"),
+        ),
+        global_batch_size=8,
```
pbeukema: use variable: (`GLOBAL_BATCH_SIZE`)

helios/olmo_core_proto.py
Hgherzog: There is a olmo-core main launch script that once we have all the components we should be able to use directly if not we will make this into a cli

PR #41, opened 2025-02-03
Henryh/component mocks

Hgherzog:


helios/data/dataset.py
```diff
@@ -1,19 +1,111 @@
...
+        for field in self._fields:
+            val = getattr(self, field)
+            if ignore_nones and (val is None):
+                continue
+            else:
+                return_dict[field] = val
+        return return_dict
+
+    @staticmethod
+    def attribute_to_bands() -> dict[str, list[str]]:
```
gabrieltseng, line 60: I'd added a test to make sure the attributes were all represented in this method:

```python
from helios.dataseg import HeliosSample


def test_all_attrs_have_bands():
    attribute_to_bands = HeliosSample.attribute_to_bands()
    for attribute_name in HeliosSample._fields:
        assert attribute_name in attribute_to_bands
```

It's probably a good idea to add it in this branch too

helios/data/dataset.py
```diff
@@ -1,19 +1,111 @@
...
+        return {"s2": S2_BANDS, "latlon": LATLON_BANDS, "timestamps": TIMESTAMPS}
+
+    @property
+    def b(self) -> int:
+        """Get the batch size.
+
+        Returns:
+            The batch size of the sample.
+        """
+        raise NotImplementedError
```
gabrieltseng: Why is this not implemented? Is it because we aren't sure if this is a batch or a single instance?

If so, can we just check if there are 4 or 5 dims in the `s2` array?

helios/train/loss.py
```diff
@@ -48,3 +59,113 @@ def patch_disc_loss(
...
+        Returns:
+            The computed loss value.
+
+        Raises:
+            NotImplementedError: This method needs to be implemented.
+        """
+        raise NotImplementedError
+
+
+class CompositeLoss:
```
gabrieltseng: What use case do you see for the composite loss? 

In the use cases I have in my head, we will want to have different losses for different inputs / outputs, so I don't know if this class is necessary.

helios/train/masking.py
```diff
@@ -0,0 +1,165 @@
...
+        raise NotImplementedError
+
+
+class CompositeMaskingStrategy(MaskingStrategy):
+    """Combines multiple masking strategies with configurable application logic."""
+
+    def __init__(
+        self,
+        strategies: list[MaskingStrategy],
+        probabilities: list[float] | None = None,
```
gabrieltseng: Is there a specific use case you have in mind for randomly picking a masking strategy? 

In Galileo we did this deterministically since the masking strategy determined the loss and the target encoding.

helios/train/masking.py
```diff
@@ -0,0 +1,165 @@
...
+import random
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from typing import Any, NamedTuple
+
+from class_registry import ClassRegistry
+from olmo_core.config import Config
+
+from helios.data.dataset import HeliosSample
+from helios.helios.types import ArrayTensor
```
gabrieltseng: typo - this should be `helios.types`

helios/data/dataset.py
gabrieltseng: Added in https://github.com/allenai/helios/pull/42/commits/b63fa881c62fdfd6af1c0b465f3cf96955bcb46a

Hgherzog: great thanks for catching this

helios/train/loss.py
Hgherzog: I imagined that we might want to try different combinations of losses at some point, it might be premature right now though.

helios/train/masking.py
Hgherzog: I was thinking that randomly picking a mask could be equivalent to masking the same batch in 4 different ways. I don't think this is needed yet though (or maybe ever) so I am going to simplify it

PR #42, opened 2025-02-04
Add patch disc loss

gabrieltseng:
Adds a patch disc loss with tests

### DIFF
--- a/helios/train/loss.py
+++ b/helios/train/loss.py
@@ -140,71 +140,20 @@ class PatchDiscriminationLoss(Loss):
         return loss
 
 
-class CompositeLoss:
-    """Combines multiple loss functions with optional weights."""
-
-    def __init__(self, losses: list[Loss], weights: list[float] | None = None):
-        """Initialize a composite loss.
-
-        Args:
-            losses: List of loss functions to combine
-            weights: Optional weights for each loss function.
-                    If None, losses are weighted equally.
-        """
-        self.losses = losses
-        self.weights = weights or [1.0] * len(losses)
-
-        if len(self.weights) != len(losses):
-            raise ValueError("Number of weights must match number of losses")
-
-    def compute(
-        self, predictions: TokensAndMasks, targets: TokensAndMasks, **kwargs: Any
-    ) -> torch.Tensor:
-        """Compute weighted combination of multiple losses.
-
-        Args:
-            predictions: Model predictions.
-            targets: Ground truth targets.
-            **kwargs: Additional keyword arguments.
-
-        Returns:
-            Combined loss value as a tensor.
-        """
-        total_loss = torch.zeros(1, device=predictions.device)
-        for loss_fn, weight in zip(self.losses, self.weights):
-            total_loss += weight * loss_fn.compute(predictions, targets, **kwargs)
-        return total_loss
-
-
 @dataclass
 class LossConfig(Config):
     """Configuration for loss functions.
 
     Args:
-        losses: List of loss configs in the format of
+        loss_config: Loss config in the format of
         e.g.
         {
             "type": "patch_discrimination",
             # rest of init kwargs
-        }
-        weights: Optional weights for each loss function.
     """
 
-    losses: list[dict[str, Any]]  # List of loss configs
-    weights: list[float] | None = None
+    loss_config: dict[str, Any]  # List of loss configs
 
-    def validate(self) -> None:
-        """Validate the loss configuration."""
-        if self.weights and len(self.weights) != len(self.losses):
-            raise ValueError("Number of weights must match number of losses")
-
-    def build(self) -> CompositeLoss:
-        """Build a CompositeLoss from the config."""
-        built_losses = []
-        for loss_config in self.losses:
-            loss_type = loss_config.pop("type")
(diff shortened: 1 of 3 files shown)


helios/train/loss.py
```diff
@@ -77,10 +30,33 @@ def compute(self, predictions: Any, targets: Any, **kwargs: Any) -> float:
...
+            return rearrange(x, "b c h w d -> b (h w c) d")
+        elif x.dim() == 4:
+            # (B, C_G, T, D)
+            return rearrange(x, "b c t d -> b (t c) d")
+        elif x.dim() == 3:
+            # (B, C_G, D)
+            return x
+
+    @staticmethod
+    def _expand_and_reciprocate(t):
```
Hgherzog, line 69: Suggestion: add a doc string showing an example of this manipulation (this is from an AI so verify it)
```
# Example transformation:
# If t = tensor([2, 3])
# reciprocals = tensor([0.5000, 0.3333])  # 1/2, 1/3
# final output = tensor([0.5000, 0.5000, 0.3333, 0.3333, 0.3333])
#               repeats each value t[i] times: 0.5 repeated 2x, 0.3333 repeated 3x
```

helios/train/loss.py
```diff
@@ -94,11 +70,59 @@ def compute(
...
+                self._flatten(getattr(predictions, f"{d}_mask").unsqueeze(dim=-1))
+                for d in predictions.data_fields
+            ],
+            dim=1,
+        )[:, :, 0]
+        all_targets = torch.cat(
+            [self._flatten(getattr(targets, d)) for d in predictions.data_fields], dim=1
+        )
+
+        pred = all_preds[all_masks == 2].unsqueeze(dim=0)
```
Hgherzog: We should have an Enum for the different mask values and there meanings for readability and consistency

helios/train/loss.py
```diff
@@ -77,10 +30,33 @@ def compute(self, predictions: Any, targets: Any, **kwargs: Any) -> float:
 class PatchDiscriminationLoss(Loss):
     """Loss function for patch discrimination task."""
 
-    def __init__(self, tau: float = 0.07, pred2unit: bool = True):
+    def __init__(
+        self, tau: float = 0.07, pred2unit: bool = True, mask_other_samples: bool = True
+    ):
         """Initialize patch discrimination loss."""
```
Hgherzog: add args to doc string

helios/train/loss.py
```diff
@@ -94,11 +70,59 @@ def compute(
...
+            pred = (pred - pred_mu) / (pred_std + 1e-4)
+
+        pred = F.normalize(pred, p=2, dim=-1)
+        target = F.normalize(target, p=2, dim=-1)
+
+        scores = torch.einsum("npd,nqd->npq", pred, target) / self.tau
+        count = (all_masks == 2).sum(dim=-1)
+
+        if self.mask_other_samples:
+            logit_mask = torch.full_like(scores, -torch.finfo(scores.dtype).max)
```
Hgherzog, line 121: Suggestion: add a helper function for the mask_other_samples

helios/train/loss.py
```diff
@@ -77,10 +30,33 @@ def compute(self, predictions: Any, targets: Any, **kwargs: Any) -> float:
 class PatchDiscriminationLoss(Loss):
```
Hgherzog, line 31: What we are calling patch discrimination loss here seems to be the AllDisc loss from the Galileo paper where we also average across the batch dimension? Maybe add a reference in the function doc string explaining this?

tests/unit/train/test_loss.py
```diff
@@ -0,0 +1,26 @@
...
+        latlon_mask=torch.ones((b, 1)) * 2,
+    )
+    targets = TokensAndMasks(
+        s2=torch.ones((b, t, t_h, t_w, d)),
+        s2_mask=torch.zeros((b, t, t_h, t_w)),
+        latlon=torch.ones((b, 1, d)),
+        latlon_mask=torch.zeros((b, 1)),
+    )
+    loss = PatchDiscriminationLoss()
+    _ = loss.compute(preds, targets)
```
Hgherzog: Suggestion: add an assert for the loss value, at least that it is approximately what we expect (checking full precision will probably create more trouble than it is worth

helios/train/loss.py
gabrieltseng: Added.

helios/train/loss.py
gabrieltseng: Added.

helios/train/loss.py
gabrieltseng: Added

helios/train/loss.py
gabrieltseng: Added

tests/unit/train/test_loss.py
gabrieltseng: Added

helios/train/loss.py
gabrieltseng: I am reluctant to excessively nest this function, so would prefer to keep it as is. If you feel strongly, I'll change it

Issue #11, opened 2025-01-27
Add model architecture


yawenzzzz:
The approach begins with Early Fusion, where each modality is patchified separately and processed using its own linear projector, then concatenate token embeddings as inputs of multimodal encoder.

PR #45, opened 2025-02-06
Henryh/add flexi gallileo arch

Hgherzog:
This PR contains the flexi-helios architecture which is basically galilleo with per modality parch embeddings rather than per data-variation type.  adresses #11 

Still to do/ I think we should leave for a separate PR
-  Add s1 and worldcover to tokensandmasks
-  Set up configs
- Hook up the rest of the pipeline for this

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,0 +1,1 @@
+"""Test suite for Helios."""


helios/nn/model.py
```diff
@@ -0,0 +1,1176 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class TokensOnly(NamedTuple):
+    s2: torch.Tensor
+
+
+# THis  should be in a utility file
+class ModuleListWithInit(nn.ModuleList):
```
gabrieltseng: This is not necessary - sorry, its code from older experiments. We can replace this with a normal `ModuleList`

helios/nn/flexihelios.py
```diff
@@ -0,0 +1,1199 @@
...
+        # WE want to be able to remove certain bands and moda
+        # dict will be modality -> channel_group -> bands
+        self.per_modality_embeddings = nn.ModuleDict({})
+        for (
+            modality,
+            channel_groups_dict,
+        ) in self.modalities_to_channel_groups_dict.items():
+            self.per_modality_embeddings[modality] = nn.ModuleDict(
+                {
+                    channel_group: FlexiPatchEmbed(
```
gabrieltseng, line 86: Future fix - not all modalities will want to be a convolutional projection. For example, latlons can just be projected with an `nn.Linear` into the token space.

helios/nn/flexihelios.py
```diff
@@ -0,0 +1,1199 @@
...
+            torch.nn.init.xavier_uniform_(m.weight)
+            if isinstance(m, nn.Linear) and m.bias is not None:
+                nn.init.constant_(m.bias, 0)
+
+    @staticmethod
+    def calculate_gsd_ratio(input_res: float, patch_size: int) -> float:
+        """Calculate the Ground Sample Distance ratio."""
+        return input_res * patch_size / BASE_GSD
+
+    # TODO: Maybe this should just return a dict so we don't do so much unpacking and repacking
```
gabrieltseng: nit - I think this comment can be removed?

helios/nn/flexihelios.py
```diff
@@ -0,0 +1,1199 @@
...
+
+            # find the resolution that each token represents, which will be
+            # the number of pixels in a patch * the resolution of each pixel
+
+            gsd_ratio = self.calculate_gsd_ratio(input_res, patch_size)
+
+            # We also want a 2D space
+            assert (
+                h == w
+            ), "get_2d_sincos_pos_encoding_with_resolution currently requires that h==w"
```
gabrieltseng: nit - could this be higher (e.g. line 262) so it triggers as quickly as possible? Also I think this should be a ValueError not an assert.

helios/nn/flexihelios.py
```diff
@@ -0,0 +1,1199 @@
...
+            output_dict[modality] = torch.stack(modality_tokens, dim=-2)
+            output_dict[masked_modality_name] = torch.stack(modality_masks, dim=-1)
+        # Sort of Hacky way to satisfy the output being a named tuple we already have
+        output_dict["latlon"] = input_data.latlon
+        output_dict["latlon_mask"] = input_data.latlon_mask
+        return TokensAndMasks(**output_dict)
+
+
+class FlexiHeliosCompositeEncodings(nn.Module):
+    """Composite encodings for the FlexiHelios model."""
```
gabrieltseng, line 167: This is very nice !

helios/nn/flexihelios.py
```diff
@@ -0,0 +1,1199 @@
...
+        s2_mask: sentinel 2 mask indicating which tokens are masked/unmasked
+        latlon: lat lon data containing geographical coordinates
+        latlon_mask: lat lon mask indicating which coordinates are masked/unmasked
+        timestamps: timestamps of the data
+    """
+
+    s2: Tensor  # (B, C_G, T, P_H, P_W)
+    s2_mask: Tensor
+    # Not sure how these fit in yet will be needed when there is a missing timestamp missing timestamp and latlon are the same thing
+    latlon: Tensor | None = None
```
gabrieltseng, line 39: The loss function currently will fail with Nones - it expects all tensors to be present (even if they are fully masked out)

helios/nn/flexihelios.py
Hgherzog: Agreed, i will add an issue for this

helios/nn/flexihelios.py
Hgherzog: I am not returning Nones from predicitons or masks but I haven't dealt with latlons well in this PR, so this None allows me to ignore it throughout and just pass back the initial values, I will add this to the issue

gabrieltseng: Can you add a small comment noting this "none" is temporary and should eventually be undone?

PR #48, opened 2025-02-07
Random masking

gabrieltseng:
There are a few missing features:

1. Handle fully missing modalities (this is what the extra value in the masking enum is intended for)
2. Handle other shapes in the random masking
3. Add the space & time masking

but this is a good minimum working example to get us started I think.


helios/train/masking.py
```diff
@@ -95,7 +186,61 @@ def apply_mask(self, batch: HeliosSample, **kwargs: Any) -> MaskedHeliosSample:
...
         """
-        raise NotImplementedError
+        # should these not be kwargs but instead be explicitly
+        # in the function signature?
+        patch_size: int = kwargs["patch_size"]
+        modalities_to_channel_groups_dict: dict[str, dict[str, list[int]]] = kwargs[
+            "modalities_to_channel_groups_dict"
+        ]
+        encode_ratio: float = kwargs["encode_ratio"]
+        decode_ratio: float = kwargs["decode_ratio"]
```
gabrieltseng, line 222: Copying the comment for visibility:

Should these not be kwargs but instead be explicitly in the function signature?

helios/train/masking.py
```diff
@@ -95,7 +186,61 @@ def apply_mask(self, batch: HeliosSample, **kwargs: Any) -> MaskedHeliosSample:
...
+                b = modality.shape[0]
+                mask = self._create_mask_per_static_modality(
+                    b,
+                    encode_ratio,
+                    decode_ratio,
+                    modalities_to_channel_groups_dict[modality_name],
+                    return_device,
+                )
+            else:
+                raise ValueError(f"Unsupported modality shape {modality.shape}")
```
gabrieltseng, line 264: Just to call out that `_create_mask_per_space_time_modality` and `_create_mask_per_static_modality` behave a little differently.

`_create_mask_per_space_time_modality` will mask out the same ratios of values per instance in the batch.

However, since a static modality might have very few tokens in a batch (e.g. 1 for latlons) instead we mask out a certain ratios of values across the entire batch.

helios/train/masking.py
Hgherzog: It depends if they will be vars in every masking strategy? I think it is preferable to leave make the apply masks call in the train module agnostic to the masking strategy. It should likely be explicit at the config build level though. Let's wait on this

helios/train/masking.py
Hgherzog: can you add this comment to the doc strings of the static methods

gabrieltseng: Added

PR #51, opened 2025-02-08
Integrate components

Hgherzog:
https://docs.google.com/document/d/1JxedTmIlFPEZcVtYIhjxhPUCfWzGIj58nbLxx_xQhjE/edit?tab=t.0

### DIFF
--- a/tests/unit/train/test_masking.py
+++ b/tests/unit/train/test_masking.py
@@ -33,6 +33,9 @@ def test_random_masking() -> None:
     # TODO: Add assert that checks input output shapes
     # check that each modality has the right masking ratio
     for modality_name in masked_sample._fields:
+        # TODO: SKipping lat lon for now
+        if modality_name.startswith("latlon"):
+            continue
         if modality_name.endswith("mask"):
             mask = getattr(masked_sample, modality_name)
             total_elements = mask.numel()


helios/data/dataloader.py
```diff
@@ -243,13 +243,17 @@ def _format_fname_from_fields(self, prefix: str, **fields: Any) -> str:
 
     def get_mock_batch(self) -> HeliosSample:
```
gabrieltseng, line 244: nit: can this be a staticmethod?

helios/data/dataset.py
```diff
@@ -65,6 +64,23 @@ def as_dict(self, ignore_nones: bool = True) -> dict[str, Any]:
                 return_dict[field] = val
         return return_dict
 
+    def to_device(self, device: torch.device) -> "HeliosSample":
```
gabrieltseng, line 67: nit: should this just be `to` to be consistent with pytorch syntax?

helios/data/dataloader.py
Hgherzog: It is a subclass of an abstractmethod that has a self argument in olmo-core , also when using a subset of different modalities we might want to be able to dynamically change this  method

helios/data/dataset.py
Hgherzog: @favyen2 pointed out that pytorch to also converts datatype which we are not supporting here, so this could be confusing.

gabrieltseng: sounds good !

PR #54, opened 2025-02-11
Add other modalities to HeliosSample and dataset

yawenzzzz:
Added S1 and WorldCover data to HeliosSample and HeliosDataset, also updated the bands to use BandSets definitions. The next step is updating MaskedHeliosSample, which is better in another PR to avoid breaking other tests.

### DIFF
--- a/helios/data/constants.py
+++ b/helios/data/constants.py
@@ -78,7 +78,7 @@ class ModalitySpec:
         return get_resolution(self.tile_resolution_factor)
 
     def get_band_names(self) -> list[str]:
-        """Get the band names."""
+        """Get the combinedband names."""
         return [band for band_set in self.band_sets for band in band_set.bands]
 
 
@@ -121,8 +121,7 @@ class Modality:
             # 15 m/pixel bands that we store at 10 m/pixel.
             BandSet(["B8"], 16),
             # 30 m/pixel bands that we store at 20 m/pixel.
-            # TODO: Check if the resolution factor is 16 or 32?
-            BandSet(["B1", "B2", "B3", "B4", "B5", "B6", "B7", "B9", "B10", "B11"], 16),
+            BandSet(["B1", "B2", "B3", "B4", "B5", "B6", "B7", "B9", "B10", "B11"], 32),
         ],
         is_multitemporal=True,
     )
--- a/helios/data/dataset.py
+++ b/helios/data/dataset.py
@@ -31,9 +31,6 @@ from helios.types import ArrayTensor
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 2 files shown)


helios/data/constants.py
```diff
@@ -180,7 +184,7 @@ def get_all_modalities(cls) -> list[ModalitySpec]:
         return [
             cls.S1,
             cls.S2,
-            cls.NAIP,
+            cls.WORLDCOVER,
```
Hgherzog, line 187: Suggestion: I think this should actually return all modalities in a non hardcoded way. We should be able to configurable pass in which modalities we want to use or not

helios/data/dataloader.py
```diff
@@ -244,14 +244,18 @@ def _format_fname_from_fields(self, prefix: str, **fields: Any) -> str:
     def get_mock_batch(self) -> HeliosSample:
```
Hgherzog, line 244: Suggestion: Add an integration test for mock batch, as this changes it could cause some bugs if we forget to update it.

tests/integration/data/test_helios_dataset.py
Hgherzog, line 1: nit: naming convention is usually test_{file name}

helios/data/dataset.py
```diff
@@ -31,22 +37,71 @@ class HeliosSample(NamedTuple):
...
-    s2: ArrayTensor | None = None  # [B, len(S2_bands), T H, W]
+    # Input shape is (B, H, W, T, C)
+    s2: ArrayTensor | None = None  # [B, H, W, T, len(S2_bands)]
+    s1: ArrayTensor | None = None  # [B, H, W, T, len(S1_bands)]
+    worldcover: ArrayTensor | None = None  # [B, H, W, T, len(WC_bands)]
     latlon: ArrayTensor | None = None  # [B, 2]
-    timestamps: ArrayTensor | None = None  # [B, D=3, T], where D=[day, month, year]
+    timestamps: ArrayTensor | None = None  # [B, T, D=3], where D=[day, month, year]
+
+    def shape(self, attribute: str, num_channels: int | None = None) -> Sequence[int]:
```
Hgherzog, line 49: Suggestion: A unit tests verifying all the modalities are included in the shape and as_dict methods

helios/data/dataset.py
```diff
@@ -185,21 +252,6 @@ def __init__(
         self._work_dir: Path | None = None  # type: ignore
```
Hgherzog, line 252: Suggestion: Set the supported/used modalities in the init so it can be changed configurably and be shared across components in the training pipeline

PR #56, opened 2025-02-12
Replace channel groups with band sets

gabrieltseng:
Small updates:

- [x] Rename all namedtuple attributes to be consistent with the `Modality` names

### DIFF
--- a/helios/nn/flexihelios.py
+++ b/helios/nn/flexihelios.py
@@ -27,8 +27,8 @@ class TokensAndMasks(NamedTuple):
     """Output to compute the loss on.
 
     Args:
-        sentinel2: sentinel 2 data of shape (B, P_H, P_W, T, Band_Sets)
-        sentinel2_mask: sentinel 2 mask indicating which tokens are masked/unmasked
+        sentinel2: sentinel 2 data of shape (B, P_H, P_W, T, Band_Sets, D)
+        sentinel2_mask: sentinel 2 mask indicating which tokens are masked/unmasked (B, P_H, P_W, T, Band_Sets)
         latlon: lat lon data containing geographical coordinates
         latlon_mask: lat lon mask indicating which coordinates are masked/unmasked
         timestamps: timestamps of the data


helios/nn/flexihelios.py
```diff
@@ -64,56 +65,60 @@ class FlexiHeliosPatchEmbeddings(nn.Module):
...
+                instantiation supports
             max_patch_size: Maximum size of patches
             embedding_size: Size of embeddings
         """
         super().__init__()
-        self.modalities_to_channel_groups_dict = modalities_to_channel_groups_dict
         self.max_patch_size = max_patch_size
         self.embedding_size = embedding_size
+        self.supported_modalities = supported_modalities
         # TODO: want to be able to remove certain bands and modalities
```
Hgherzog, line 84: Remove comments about channel groups

helios/nn/flexihelios.py
```diff
@@ -64,56 +65,60 @@ class FlexiHeliosPatchEmbeddings(nn.Module):
...
-        for modality in self.modalities_to_channel_groups_dict.keys():
+        for modality in self.supported_modalities:
             self.per_modality_embeddings[modality] = (
                 self._get_patch_embedding_module_for_modality(modality)
             )
 
     def _get_patch_embedding_module_for_modality(self, modality: str) -> nn.Module:
         """Get the patch embedding module for a modality."""
-        channel_groups_dict = self.modalities_to_channel_groups_dict[modality]
+        modality_spec = Modality.get_modality_from_name(modality)
```
Hgherzog: This is great, I totally agree we needed to move away from the channel group dict

helios/data/constants.py
```diff
@@ -177,11 +201,18 @@ class Modality:
...
         """Get all modalities."""
-        return [
-            cls.S1,
-            cls.S2,
-            cls.NAIP,
-        ]
+        return [cls.S1, cls.S2, cls.NAIP, cls.LATLONS]
+
+    @classmethod
+    def get_modality_from_name(cls, name: str) -> ModalitySpec:
```
Hgherzog: nit: get_spec_from_name or get_modality_spec might be clearer name here

helios/nn/flexihelios.py
```diff
@@ -284,7 +291,8 @@ def _apply_encodings_per_modality(
             Tensor with encodings applied based on modality type
         """
         # TODO: Improve this implementation
-        if modality == "latlon":
+        if modality_tokens.shape == 3:
```
Hgherzog: Suggestion: Is there a way we could prescribe what encodings are needed in the Modality Spec?

helios/nn/flexihelios.py
```diff
@@ -284,7 +291,8 @@ def _apply_encodings_per_modality(
...
+            # modality_tokens = [B, C_G, D]; static in space, static in time
```
Hgherzog: nit: we shouldn't use c_g if we aren't doing channel groups

helios/constants.py
```diff
@@ -23,5 +23,11 @@
 
 WORLDCOVER_BANDS = ["B1"]
 
+MODALITY_NAME_TO_BANDS = {
```
favyen2: I think maybe the bands should be removed from `helios/constants.py` and add helper function to BandSet to get flat list of bands. It can be in alphabetical order or just in the order of the bands / BandSets.

helios/data/constants.py
favyen2: I made Modality a normal class since it wasn't working as Enum (since the values are objects) but maybe it would be better to use a ClassRegistry where each ModalitySpec is a separate instance (and then could rename ModalitySpec -> Modality, and Modality -> MODALITIES which is a ClassRegistry). This would probably be better although would cause a lot of refactoring needed.

helios/constants.py
gabrieltseng: +1 its confusing to have band information split into two files at the moment. It would be helpful to have a function which can return the correct band ordering from the `ModalitySpec`

Hgherzog: I removed these cosntants and added a helper function to be in the order of the bandsets. The data is stacked by the order of the band sets anyway so previoulsy I think the band sets were being grouped incorrectly

helios/train/masking.py
```diff
@@ -50,8 +51,8 @@ class MaskedHeliosSample(NamedTuple):
         timestamps: ArrayTensor  # [B, T, D=3], where D=[day, month, year]
     """
 
-    s2: ArrayTensor
-    s2_mask: ArrayTensor
+    sentinel2: ArrayTensor
+    sentinel2_mask: ArrayTensor
```
gabrieltseng, line 55: Not a change you made but taking the opportunity to bring this up:

Should these just be `Tensor`s? Is there a time where we'd want them to be `array`s? I don't think so - their intent is just to be passed to a model.

helios/data/constants.py
```diff
@@ -53,50 +55,64 @@ def get_resolution(self) -> float:
...
         return get_resolution(self.tile_resolution_factor)
 
+    def bandsets_as_indices(self) -> list[list[int]]:
+        """Return the band sets as indices."""
+        # TODO: Add Integration test that we actually load the data in the correct order from the band sets
+        band_specs_as_indices = []
+        for band_set in self.band_sets:
+            # TODO: I think the bands are not actually in the order of the old constant but stacked succesively from band sets
+            band_specs_as_indices.append(list(range(len(band_set.bands))))
+        return band_specs_as_indices
```
gabrieltseng: I'm not sure this is right. The original S2 bands were

```
S2_BANDS = [
    "B01",
    "B02",
    "B03",
    "B04",
    "B05",
    "B06",
    "B07",
    "B08",
    "B8A",
    "B09",
    "B10",
    "B11",
    "B12",
]
```

According to these bands, the indices of the first band set (`["B02", "B03", "B04", "B08"]`) would be `[1, 2, 3, 7]`.

More generally this doesn't keep track of indices against band sets. Originally, this function would have returned (for s2):

```
[[1, 2, 3, 9], [4, 5, 6, 8, 11, 12], [0, 9, 10]
```
but I think now it would return:
```
[[0, 1, 2, 3], [0, 1, 2, 3, 4, 5], [0, 1, 2]]
```

helios/data/constants.py
```diff
@@ -172,16 +183,18 @@ class Modality:
...
-        """Get all modalities."""
-        return [
-            cls.S1,
-            cls.S2,
-            cls.NAIP,
-        ]
-
+    ),
+    # TODO: decide if we want to include latlon as a modality
+    # The issue is that parse_modality_csv will search for the csv file and relevant ModalityTile
```
gabrieltseng, line 197: I am in favor of including latlon as a modality. Could we add a field to the modalityspec which would tell parse_modality_csv to ignore this modality?

`ignore_when_parsing_csv = True`

helios/data/constants.py
Hgherzog: Okay so this is definitely wrong, let me add a unit test, The expected behavior is to just return `[0,1,2,3], [4,5,6,..], [10, 11, 12]]` Previously, we were loading the bands by bandset not by the order of this constant so it was wrong

helios/data/dataset.py
```diff
@@ -111,69 +124,17 @@ def to_device(self, device: torch.device) -> "HeliosSample":
             A new HeliosSample with all tensors moved to the specified device.
         """
         return HeliosSample(
-            s2=self.s2.to(device) if self.s2 is not None else None,
+            sentinel2=self.sentinel2.to(device) if self.sentinel2 is not None else None,
+            sentinel1=self.sentinel1.to(device) if self.sentinel1 is not None else None,
+            worldcover=(
+                self.worldcover.to(device) if self.worldcover is not None else None
+            ),
```
gabrieltseng: nit: here's an implementation which doesn't require us to explicitly update this function anytime we add a modality

```python
def to_device(self, device: torch.device) -> "HeliosSample":
    return HeliosSample(**{key: val.to(device) for key, val in self.as_dict().items()})
```

I think this (as much as possible not explicitly defining the modalities in our functions) is desirable since it otherwise introduces more friction / room for bugs when we do add more modalities.

helios/data/constants.py
```diff
@@ -98,6 +98,11 @@ def get_band_names(self) -> list[str]:
         """Get the combinedband names."""
         return [band for band_set in self.band_sets for band in band_set.bands]
 
+    @property
+    def num_channels(self) -> int:
```
gabrieltseng: nit: can we stick with `bands` to be consistent here?

helios/nn/flexihelios.py
```diff
@@ -26,22 +27,22 @@ class TokensAndMasks(NamedTuple):
     """Output to compute the loss on.
 
     Args:
-        s2: sentinel 2 data of shape (B, C_G, T, P_H, P_W)
-        s2_mask: sentinel 2 mask indicating which tokens are masked/unmasked
+        sentinel2: sentinel 2 data of shape (B, P_H, P_W, T, Band_Sets)
```
gabrieltseng: nit - this comment should be `(B, P_H, P_W, T, Band_Sets, D)`

helios/train/masking.py
Hgherzog: Perhaps it might be useful for an eval or downstream task. I am going to leave it for now as it isn't clear to me right now.

helios/data/constants.py
Hgherzog: will do this in the support missing modalities PR

PR #57, opened 2025-02-12
Bigger dataset with Sentinel-1 and Sentinel-2

favyen2:
This is bigger dataset with Sentinel-1 and Sentinel-2 so far.

It refactors the dataset creation code to use same BandSet/ModalitySpec from data/constants.py.

I had to move TimeSpan from dataset/parse.py to data/constants.py to avoid circular import (since dataset/parse.py depends on dataset_creation code).


helios/data/constants.py
yawenzzzz, line 1: Things to update due to this change: 1. dataset.py – update the import path for TimeSpan. 2. data folder – update norm_configs to reflect the vv and vh band names for Sentinel-1 in both computed and predefined JSON files.

Issue #12, opened 2025-01-27
Add multiple loss implementations


yawenzzzz:
Include multiple loss functions for pretraining and evaluation, such as CrossEntropy, MAE/MSE, and PatchDiscrimination.

References: 
Presto - https://github.com/nasaharvest/presto-v3/blob/main/src/loss.py
Anysat - https://github.com/gastruc/AnySat/blob/main/src/models/losses.py

Ideally, we would like to have a losses meta-object that can take a mix of losses, which improve flexibility in terms of pretraining objective.

PR #58, opened 2025-02-12
adding MAE (L1) loss and test

pjreddie:
Poked around in the loss functions for https://github.com/allenai/helios/issues/12 , would be great to chat with someone before I add L2, cross entropy, etc.

### DIFF
--- a/helios/train/loss.py
+++ b/helios/train/loss.py
@@ -18,6 +18,10 @@ from helios.train.masking import MaskValue
 class Loss(ABC):
     """Abstract base class for loss functions."""
 
+    @staticmethod
+    def _flatten(x: Tensor) -> Tensor:
+        return rearrange(x, "b ... d -> b (...) d")
+
     @abstractmethod
     def compute(self, predictions: Any, targets: Any, **kwargs: Any) -> float:
         """Compute the loss between predictions and targets."""
@@ -49,10 +53,6 @@ class PatchDiscriminationLoss(Loss):
         self.tau = tau
         self.pred2unit = pred2unit
         self.mask_other_samples = mask_other_samples
-
-    @staticmethod
-    def _flatten(x: Tensor) -> Tensor:
-        return rearrange(x, "b ... d -> b (...) d")
 
     @staticmethod
     def _expand_and_reciprocate(t: Tensor) -> Tensor:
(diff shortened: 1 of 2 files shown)


helios/train/loss.py
```diff
@@ -131,6 +131,59 @@ def compute(
...
+
+        Args:
+            predictions: Model predictions.
+            targets: Ground truth targets.
+            **kwargs: Additional keyword arguments.
+
+        Returns:
+            The computed loss value.
+        """
+        # TODO: How will we deal with only training with some subset of modalities? If we use passed in modalities channels dict to define which modalities is one way but using class directly implies all used
```
gabrieltseng: I think the plan is for unused modalities to be reflected in the mask value (`mask_value == MASK_VALUE.MISSING`)

helios/train/loss.py
```diff
@@ -131,6 +131,59 @@ def compute(
...
+@LOSS_REGISTRY.register("mae")
+class MAELoss(Loss):
+    """Loss function for mean average error."""
+
+    @staticmethod
+    def _flatten(x: Tensor) -> Tensor:
+        return rearrange(x, "b ... d -> b (...) d")
+
+    @staticmethod
+    def _expand_and_reciprocate(t: Tensor) -> Tensor:
```
gabrieltseng: Is this function necessary for MAELoss? If not I think we can remove it.

helios/train/loss.py
```diff
@@ -131,6 +131,59 @@ def compute(
         return loss
 
 
+@LOSS_REGISTRY.register("mae")
+class MAELoss(Loss):
+    """Loss function for mean average error."""
+
+    @staticmethod
+    def _flatten(x: Tensor) -> Tensor:
```
gabrieltseng: nit - if we will be using this function many times (which seems likely) maybe we can move it into the Loss ABC?

helios/train/loss.py
```diff
@@ -131,6 +131,59 @@ def compute(
         return loss
 
 
+@LOSS_REGISTRY.register("mae")
+class MAELoss(Loss):
```
gabrieltseng: nit - this could be confusing since we will probably want to test MAE (masked autoencoding) models. Could we call this L1 loss instead?

pjreddie: as follow up I'm wondering for registery name and class name something like:
l1 - L1Loss
l2 - L2Loss
OR
l1 - L1Loss
mse - MSELoss
or do we want to write out full names for these?

pjreddie: it's very strange to me that Pytorch uses L1Loss and MSELoss like why not L2 loss? but maybe we want to stick with their convention idk

gabrieltseng: I am happy with l1 and l2 loss

Issue #19, opened 2025-01-27
Add normalizer to the dataset


yawenzzzz:
Either preload from a multimodal normalizer, or compute directly from the dataset.

PR #62, opened 2025-02-13
Add normalization to modality data

yawenzzzz:
Issue #19. 

Based on rslearn and Presto, I added two normalization strategies:

1. Predefined min/max values
2. Computed values (calculates mean & std offline and applies a std multiplier)

Normalization can be applied per modality, allowing more flexibility, e.g., using predefined values for WorldCover, latlon, while applying computed values for Sentinel-1 and Sentinel-2.

I think later we can have a lookup table to map each modality to its normalization strategy.

### DIFF
--- a/helios/data/normalize.py
+++ b/helios/data/normalize.py
@@ -36,7 +36,8 @@ class Normalizer:
 
         Args:
             strategy: The strategy to use for normalization (predefined or computed).
-            std_multiplier: The multiplier for the standard deviation when using computed values.
+            std_multiplier: Optional, only for strategy COMPUTED.
+                            The multiplier for the standard deviation when using computed values.
 
         Returns:
             None


helios/data/dataset.py
```diff
@@ -307,19 +310,22 @@ def __getitem__(self, index: int) -> HeliosSample:
...
+            if modality == Modality.SENTINEL1:
+                image = convert_to_db(image)
+            # Normalize data using predefined or computed values
+            if NORMALIZE_STRATEGY[modality] == Strategy.PREDEFINED:
+                sample_dict[modality.name] = self.normalizer_predefined.normalize(
+                    modality, image
+                ).astype(self.dtype)
+            elif NORMALIZE_STRATEGY[modality] == Strategy.COMPUTED:
+                sample_dict[modality.name] = self.normalizer_computed.normalize(
+                    modality, image
```
Hgherzog: nit: apply astype to the data once after the normalizing step and then assign to the dictionary once

helios/data/normalize.py
```diff
@@ -0,0 +1,114 @@
...
+            data: The data to normalize.
+
+        Returns:
+            The normalized data.
+        """
+        modality_bands = modality.band_order
+        modality_norm_values = self.norm_config[modality.name]
+        # When using predefined values, we have the min and max values for each band
+        if self.strategy == Strategy.PREDEFINED:
+            min_vals = []
```
Hgherzog: suggestion: Have separate methods for the different strategies so they can be tested independently

yawenzzzz: Good call, I'm actually also thinking separating them this morning. Fixed.

helios/data/dataset.py
yawenzzzz: Sure, fixed.

PR #63, opened 2025-02-14
Henryh/ignore missing modalities

Hgherzog:
Major Changes to Note
- MaskedHeliosSample and TokensAndMasks will be None if the modality isn't present
- to use only certain modalities we pass a supported modalities list to the parser, dataset and model (this is redundant for now) 
- Inside flexihelios, we are only going to use named tuples as an api between seperate components not with the nn.modules themselves as dicts offer more simplicity and flexibility 

Next steps are to actually try to get the run going with at least S1 and S2

### DIFF
--- a/helios/olmo_core_proto.py
+++ b/helios/olmo_core_proto.py
@@ -113,7 +113,6 @@ if __name__ == "__main__":
     loss_config = LossConfig(
         loss_config={
             "type": "patch_discrimination",
-            "supported_modalities": supported_modalities,
         }
     )
     train_module_config = HeliosTrainModuleConfig(
--- a/helios/train/masking.py
+++ b/helios/train/masking.py
@@ -39,6 +39,7 @@ class MaskValue(Enum):
 class MaskedHeliosSample(NamedTuple):
     """A masked sample of the data from the Helios dataset.
 
+    We always require sentinel2 data.
     This is a namedtuple that contains the data for a single sample from the Helios dataset.
     latlon and timestamps are the same for all modalities.
     For each modality. we have an ArrayTensor named by modality, and a mask for each modality named by modality_mask.
@@ -48,8 +49,8 @@ class MaskedHeliosSample(NamedTuple):
     timestamps: (
         ArrayTensor  # [B, T, D=3], where D=[day, month, year] (months are zero indexed)
     )
-    sentinel2: ArrayTensor | None = None
-    sentinel2_mask: ArrayTensor | None = None
+    sentinel2: ArrayTensor
+    sentinel2_mask: ArrayTensor
     sentinel1: ArrayTensor | None = None
     sentinel1_mask: ArrayTensor | None = None
     worldcover: ArrayTensor | None = None
--- a/helios/data/dataset.py
+++ b/helios/data/dataset.py
@@ -37,11 +37,12 @@ logger = logging.getLogger(__name__)
 class HeliosSample(NamedTuple):
     """A sample of the data from the Helios dataset.
 
+    We always require sentinel2 data.
     This is a namedtuple that contains the data of a single sample or a batch of samples from the Helios dataset.
     For each modality, we have an ArrayTensor named by the modality, along with the latlon and timestamps.
     """
 
-    sentinel2: ArrayTensor | None = None  # [B, H, W, T, len(S2_bands)]
+    sentinel2: ArrayTensor  # [B, H, W, T, len(S2_bands)]
     sentinel1: ArrayTensor | None = None  # [B, H, W, T, len(S1_bands)]
     worldcover: ArrayTensor | None = None  # [B, H, W, len(WC_bands)]
     latlon: ArrayTensor | None = None  # [B, 2]
--- a/helios/train/loss.py
+++ b/helios/train/loss.py
@@ -30,10 +30,6 @@ class Loss(ABC):
         """Compute the loss between predictions and targets."""
         pass
 
-    @staticmethod
-    def _flatten(x: Tensor) -> Tensor:
-        return rearrange(x, "b ... d -> b (...) d")
-
     @classmethod
     def _flatten_tokens_or_masks(
         cls, x: TokensAndMasks, is_masks: bool = False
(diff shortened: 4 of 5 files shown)


helios/train/train_module.py
```diff
@@ -369,7 +371,9 @@ def train_batch(self, batch: HeliosSample, dry_run: bool = False) -> None:
         # Move tensors to the right device.
         # we may want to modify this
         batch = batch.to_device(self.device)
-        kwargs = {"patch_size": 8, "encode_ratio": 0.5, "decode_ratio": 0.5}
+        # TODO: ENsure patch size stuff is the same between encoder and target encoder
+        # TODO: Need to make this dynamic and configurable
+        kwargs = {"patch_size": 16, "encode_ratio": 0.5, "decode_ratio": 0.5}
```
yawenzzzz, line 376: FYI, in the configuration PR, I have fixed this issue: https://github.com/allenai/helios/pull/65.

helios/train/masking.py
```diff
@@ -42,24 +43,21 @@ class MaskedHeliosSample(NamedTuple):
...
-    sentinel2_mask: ArrayTensor
-    latlon: ArrayTensor  # [B, 2]
-    latlon_mask: ArrayTensor
     timestamps: (
         ArrayTensor  # [B, T, D=3], where D=[day, month, year] (months are zero indexed)
     )
-
-    def as_dict(self) -> dict[str, Any]:
+    sentinel2: ArrayTensor | None = None
+    sentinel2_mask: ArrayTensor | None = None
```
yawenzzzz: Our setup strongly assumes `sentinel2` always exist, so `sentinel2` and `sentinel2_mask` shouldn’t be optional, similarly in the `HeliosSample`.

helios/train/masking.py
```diff
@@ -229,6 +258,51 @@ def _create_mask_per_space_time_modality(
         else:
             return space_time_mask
 
+    # TODO: We should be able to do this agnostic of dimensionality
+    @staticmethod
+    def _create_mask_per_space_modality(
```
yawenzzzz, line 264: +1, better to consolidate these functions -> include the common logic into a `_create_mask` function, and then have separate functions for static, space, space_time create mask, only changing the reshape pattern. With this approach, we can pass `shape` directly instead of specific `b`, `h`, `w`, 't'.

tests/conftest.py
```diff
@@ -127,19 +133,17 @@ def prepare_samples_func(data_path: Path) -> list[SampleInformation]:
...
         return samples
 
-    return prepare_samples_func
+    return (
+        prepare_samples_func,
+        [
+            Modality.SENTINEL2,
+            Modality.SENTINEL1,
+            Modality.WORLDCOVER,
+            Modality.LATLON,  # We want to include latlon even though it is not a read in modality
```
yawenzzzz, line 147: adding `ignore_when_parsing` is a good fix.

helios/olmo_core_proto.py
```diff
@@ -64,7 +69,12 @@
     logger.setLevel(logging.DEBUG)
     logger.info("Starting Helios training")
 
-    supported_modalities = ["sentinel2", "latlon"]
+    supported_modalities = [
+        Modality.SENTINEL2,
+        Modality.LATLON,
+        Modality.SENTINEL1,
+        # Modality.WORLDCOVER,
```
yawenzzzz, line 76: Why dropping WorldCover?

helios/olmo_core_proto.py
```diff
@@ -95,8 +105,17 @@
...
-    loss_config = LossConfig(loss_config={"type": "patch_discrimination"})
+    masking_config = MaskingConfig(
+        strategy_config={
+            "type": "random",
+        }
+    )
+    loss_config = LossConfig(
+        loss_config={
+            "type": "patch_discrimination",
+            "supported_modalities": supported_modalities,
```
yawenzzzz: I don’t see `supported_modalities` used in the loss. Any reason for adding it?

helios/nn/flexihelios.py
```diff
@@ -174,11 +184,19 @@ def is_any_data_seen_by_encoder(modality_mask: Tensor) -> bool:
         """Check if any data is seen by the encoder."""
         return modality_mask.min() == MaskValue.ONLINE_ENCODER.value
 
+    def _get_modalities_to_process(self, input_data: MaskedHeliosSample) -> list[str]:
+        """Get the modalities to process."""
+        available_modalities = input_data.modalities
+        modalities_to_process = set(self.supported_modality_names).intersection(
+            set(available_modalities)
+        )
+        return list(modalities_to_process)
```
yawenzzzz, line 193: Do we still need this check after `supported_modalities` been added into `HeliosDataset`?

helios/nn/flexihelios.py
```diff
@@ -309,64 +331,96 @@ def _apply_encodings_per_modality(
...
+            #     w,
+            #     b_s,
+            #     self.embedding_dim_per_embedding_type * 2,
+            #     device=current_device,
+            # )
+            # spatial_embed = rearrange(spatial_embed, "b (h w) d -> b h w d", h=h, w=w)
+            # spatial_embed = repeat(spatial_embed, "b h w d -> b h w b_s d", b_s=b_s)
+            # modality_embed = torch.cat(
+            #     [modality_channel_embed, sp_zeros, spatial_embed], dim=-1
+            # )
```
yawenzzzz, line 374: It looks like this shape checking happens in multiple places, like `shape` function in `HeliosSample`, `create_mask` in `masking`, and also here. They check if one modality is static, space only, or space temporal. It might be better to have a common util function that returns the modality type for a given tensor. Another option, as discussed, is to add `is_spatial` into `ModalitySpec`, then given a modality name, we can decide its shape type using `is_spatial` and `is_multitemporal`. No need to address this in this PR.

helios/nn/flexihelios.py
```diff
@@ -192,14 +210,16 @@ def forward(
...
+        for modality in modalities_to_process:
+            logger.debug(f"Processing modality {modality}")
             modality_tokens, modality_masks = self.apply_embedding_to_modality(
                 modality, input_data, patch_size
             )
             output_dict[modality] = modality_tokens
             modality_mask_name = input_data.get_masked_modality_name(modality)
             output_dict[modality_mask_name] = modality_masks
-        return TokensAndMasks(**output_dict)
+        return output_dict
```
yawenzzzz, line 222: I’m not sure if switching from TokensAndMasks to output_dict is a good choice, seems NameTuple do have the advantage of faster attribute lookup, and less memory. See: https://stackoverflow.com/questions/56193305/why-do-namedtuples-use-less-memory-than-dictionaries

helios/olmo_core_proto.py
Hgherzog: WorldCover is not completely supported yet, I will add that in a follow up PR.

helios/nn/flexihelios.py
Hgherzog: agreed.

helios/nn/flexihelios.py
Hgherzog: I agree this is a little redundant. However, I think it is more explicit though because we have to pass in supported modalities to initialize the model (for per modality patch embeddings) but for a given sample we may be missing a modality. This function returns what we should loop over for a given pass of the model.

helios/nn/flexihelios.py
Hgherzog: I think it is simpler and more flexible to have dicts internally and named tuples externally. We can re-evaluate if this becomes a bottleneck.

Issue #64, opened 2025-02-14
Make dataloader and model configurable


yawenzzzz:
Make dataloader & modal configurable in `olmo_core_proto.py`

PR #65, opened 2025-02-14
Make dataloader & model configurable

yawenzzzz:
Issue #64, add configuration & validate

- [x] make dataset & dataloader configurable
- [x] make model configurable
- [x] fix hardcoded patch_size in multiple places
- [x] make train_module model agnostic

### DIFF
--- a/helios/train/train_module/latent_mim.py
+++ b/helios/train/train_module/latent_mim.py
@@ -1,6 +1,6 @@
 """Training and optimizer abstraction for Helios."""
 
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from logging import getLogger
 from typing import Any
 
@@ -40,8 +40,8 @@ class LatentMIMTrainModuleConfig(HeliosTrainModuleConfig):
         ema_decay: EMA decay rate for target encoder (default: 0.99).
     """
 
-    loss_config: LossConfig = field(init=False)
-    masking_config: MaskingConfig = field(init=False)
+    loss_config: LossConfig = LossConfig(loss_config={"type": "patch_discrimination"})
+    masking_config: MaskingConfig = MaskingConfig(strategy_config={"type": "random"})
     ema_decay: float = 0.99
 
     def build(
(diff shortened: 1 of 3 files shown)


helios/data/dataloader.py
```diff
@@ -356,4 +358,49 @@ def __iter__(self) -> Iterator[HeliosSample]:
...
+    def validate(self) -> None:
+        """Validate the configuration."""
+        if self.work_dir is None:
+            raise ValueError("Work directory is not set")
+
+    def build(
+        self, dataset: HeliosDataset, collator: Callable = default_collate
+    ) -> "HeliosDataLoader":
+        """Build the HeliosDataLoader."""
+        if not isinstance(dataset, HeliosDataset):
```
favyen2, line 394: Should this go in validate? Or if you put it in validate is the type checker still not happy even though validate is called from build?

Just something to consider, feel free to ignore if it doesn't make sense.

helios/data/dataset.py
```diff
@@ -334,3 +354,33 @@ def __getitem__(self, index: int) -> HeliosSample:
...
 
         return HeliosSample(**sample_dict)
+
+
+@dataclass
+class HeliosDatasetConfig(Config):
+    """Configuration for the HeliosDataset."""
+
+    tile_path: UPath
+    samples: list[SampleInformation] | None = None
```
favyen2, line 508: Is there a use for this? Fine to keep but if you can't think of any use where the samples would be configured directly then consider getting rid of it.

Actually one might be if you want to debug with just a few SampleInformation. But then maybe it's better to pass the example IDs or a limit on the number of examples to load so you don't need to put the whole SampleInformation in the config. I'm not sure.

helios/data/dataloader.py
yawenzzzz: Currently, `HeliosDataLoaderConfig` can be initialized without a `dataset`, which is only provided when calling `build`. Therefore, we validate `dataset` in the `build` function instead of `validate`. I will keep it as it is for now.

helios/data/dataset.py
yawenzzzz: Yes, it's mainly for debugging and testing purpose, we have this test `tests/integration/data/test_dataset.py` that use `samples` as input. The `samples` are generated by this function `prepare_samples_and_supported_modalities` in `conftest.py`.

PR #68, opened 2025-02-17
Add functionality to subset a sample

gabrieltseng:
This PR aims to fix two issues we were running into.

1. **Problem**: OOM on runs. 
    **Solution**: Subset each batch so that the total number of tokens in an instance respects a token budget (1500 is based on Galileo, but it's probably something we could play around with).

2. **Problem**: Model not learning
    (partial) **Solution**: I don't know if this will fix it, but one thing I noticed is that the online encoder and target encoder receive the same masks. This means the target encoder has no additional information compared to the online encoder. I updated it so that the target encoder receives the unmasked input (which was the best configuration for Galileo, see Table 9 "target context" of the [paper](https://arxiv.org/abs/2502.09356)).

Some smaller cleanups / improvements as a consequence of this:
- Subsetting supports varying patch sizes, so we now train on varying patch sizes (and number of timesteps)
- Remove `base_patch_size`, which didn't seem like it was being used anywhere
- Rename the mask value from `DECODER_ONLY` to `DECODER` for clarity, since the target encoder will see those values too.

### DIFF
--- a/helios/train/masking.py
+++ b/helios/train/masking.py
@@ -82,6 +82,8 @@ class MaskedHeliosSample(NamedTuple):
         """
         return_dict: dict[str, ArrayTensor] = {}
         for key, val in self.as_dict().items():
+            if val is None:
+                continue
             if key.endswith("mask"):
                 # 1s where it is missing, 0 elsewhere
                 all_but_missing = val == MaskValue.MISSING
--- a/tests/unit/train/test_masking.py
+++ b/tests/unit/train/test_masking.py
@@ -60,4 +60,5 @@ def test_random_masking_and_unmask() -> None:
     for modality_name in unmasked_sample._fields:
         if modality_name.endswith("mask"):
             mask = getattr(unmasked_sample, modality_name)
-            assert (mask == 0).all()
+            if mask is not None:
+                assert (mask == 0).all()


helios/data/dataset.py
```diff
@@ -136,6 +138,142 @@ def maybe_move_to_device(tensor: torch.Tensor | None) -> torch.Tensor | None:
...
+
+        This function assumes we apply (H, W, T=1 patchifying)
+        """
+        used_tokens = 0
+        time_multiply_tokens = 0
+        for attribute in self.as_dict(ignore_nones=True).keys():
+            if attribute == "timestamps":
+                continue
+            modality_spec = Modality.get(attribute)
+            if (modality_spec.get_tile_resolution() > 0) and (
```
Hgherzog: Suggestion: have a property for modality spec that is this check

helios/data/dataset.py
```diff
@@ -136,6 +138,142 @@ def maybe_move_to_device(tensor: torch.Tensor | None) -> torch.Tensor | None:
...
+    def time(self) -> int:
+        """Get the number of time steps in the data."""
+        if self.timestamps is None:
+            raise ValueError("Timestamps are not present in the sample")
+        return self.timestamps.shape[1]
+
+    def _t_from_hw(self, h_w_p: int, max_tokens_per_instance: int) -> int:
+        """Find max t possible when subsetting.
+
+        Given a sampled h_w (in tokens),
```
Hgherzog: Is this h_w or h_w_p does this need to be patchified

helios/data/dataset.py
```diff
@@ -136,6 +138,142 @@ def maybe_move_to_device(tensor: torch.Tensor | None) -> torch.Tensor | None:
...
+        return self.sentinel2.shape[2]
+
+    @property
+    def time(self) -> int:
+        """Get the number of time steps in the data."""
+        if self.timestamps is None:
+            raise ValueError("Timestamps are not present in the sample")
+        return self.timestamps.shape[1]
+
+    def _t_from_hw(self, h_w_p: int, max_tokens_per_instance: int) -> int:
```
Hgherzog: Suggestion: Use a clearer function name here e.g _get_max_t_within_token_budget

helios/data/dataset.py
```diff
@@ -136,6 +138,142 @@ def maybe_move_to_device(tensor: torch.Tensor | None) -> torch.Tensor | None:
...
+
+        The returned sample will have shape:
+            height = hw_t * patch_size
+            width = hw_t * patch_size
+            time = max_t
+        where hw_t is sampled from hw_to_sample and max_t is the maximum number
+        of timesteps allowable so that the total tokens (per instance) is >=
+        max_tokens_per_instance
+        """
+        max_height_width = max(self.height, self.width)
```
Hgherzog, line 217: nit: We are assuming in many places that h == w, if we don't need to support this we can simplify here

helios/train/masking.py
```diff
@@ -71,6 +71,22 @@ def as_dict(self) -> dict[str, Any]:
             return_dict[field] = val
         return return_dict
 
+    def unmask(self) -> "MaskedHeliosSample":
```
Hgherzog, line 77: Nice catch. This may be why loss isn't going down.

helios/data/dataset.py
```diff
@@ -136,6 +138,142 @@ def maybe_move_to_device(tensor: torch.Tensor | None) -> torch.Tensor | None:
...
+    @property
+    def time(self) -> int:
+        """Get the number of time steps in the data."""
+        if self.timestamps is None:
+            raise ValueError("Timestamps are not present in the sample")
+        return self.timestamps.shape[1]
+
+    def _t_from_hw(self, h_w_p: int, max_tokens_per_instance: int) -> int:
+        """Find max t possible when subsetting.
+
```
Hgherzog, line 163: Can you add an args docstring here to make it clear that h_w_p is the height/width value after patchification

helios/train/train_module.py
```diff
@@ -368,12 +369,18 @@ def train_batch(self, batch: HeliosSample, dry_run: bool = False) -> None:
 
         # Move tensors to the right device.
         # we may want to modify this
-        batch = batch.to_device(self.device)
-        kwargs = {"patch_size": 8, "encode_ratio": 0.5, "decode_ratio": 0.5}
-        masked_batch = self.masking_strategy.apply_mask(batch, **kwargs)
+        token_budget = 1500
+        h_w_to_sample = list(range(1, 13))
+
+        patch_size = np.random.choice(self.model.encoder.max_patch_size)
```
Hgherzog, line 377: I updated the patch size and some of this logic in my PR, but we should go with what's here, feel free to do that when you merge

helios/data/dataset.py
gabrieltseng: ah - I tried to make this clear with the `in_tokens` - would it be clearer to say `given a sampled h_w_p (the number of tokens along the h and w dimensions ...`?

helios/data/dataset.py
gabrieltseng: updated

helios/data/dataset.py
gabrieltseng: for now this is cheap so I am happy to leave it here.

helios/data/dataset.py
gabrieltseng: updated

helios/data/dataset.py
gabrieltseng: updated as per the comment above

helios/train/train_module.py
```diff
@@ -370,14 +371,18 @@ def train_batch(self, batch: HeliosSample, dry_run: bool = False) -> None:
...
         # we may want to modify this
-        batch = batch.to_device(self.device)
-        # TODO: ENsure patch size stuff is the same between encoder and target encoder
-        # TODO: Need to make this dynamic and configurable
-        kwargs = {"patch_size": 16, "encode_ratio": 0.5, "decode_ratio": 0.5}
-        masked_batch = self.masking_strategy.apply_mask(batch, **kwargs)
+        token_budget = 1500
+        h_w_to_sample = list(range(1, 13))
+
+        patch_size = np.random.choice(self.model.encoder.max_patch_size)
```
Hgherzog, line 377: Patch size can be zero here due to random choice

Hgherzog:
```suggestion
        patch_size = np.random.choice(np.arange(1, self.model.encoder.max_patch_size))
```

PR #75, opened 2025-02-18
Henryh/fix loss bug

Hgherzog:
- small fixes to gabi's PR

### DIFF
--- a/helios/data/dataset.py
+++ b/helios/data/dataset.py
@@ -13,19 +13,25 @@ import numpy as np
 import pandas as pd
 import torch
 from einops import rearrange
-from helios.data.constants import (BASE_RESOLUTION, IMAGE_TILE_SIZE,
-                                   TIMESTAMPS, Modality, ModalitySpec,
-                                   TimeSpan)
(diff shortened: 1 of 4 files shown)


helios/data/dataset.py
```diff
@@ -227,11 +227,11 @@ def subset(
             sampled_hw_p, max_tokens_per_instance
         )
         sampled_hw = sampled_hw_p * patch_size
-
-        start_h = np.random.choice(self.height - sampled_hw_p + 1)
-        start_w = np.random.choice(self.width - sampled_hw_p + 1)
+        max_start_h = self.height - sampled_hw + 1
+        start_h = np.random.choice(max_start_h)
+        # TODO: FORCE h == w for now other option is to update 2d pos encoding
+        start_w = start_h  # np.random.choice(self.width - sampled_hw_p + 1)
```
gabrieltseng, line 233: @Hgherzog I think its fine - the 2d pos encoding just comunicates the relative position of tokens within an instance. The subsetting will change what an instance is.

I'd prefer to have the original setup since it allows for many more augmentations per instance.

PR #76, opened 2025-02-19
Henryh/fix eval

Hgherzog:
- updates Eval so we can get EuroSat results


helios/evals/embeddings.py
```diff
@@ -21,14 +23,26 @@ def get_embeddings(
...
+                    )
+            masked_helios_sample = MaskedHeliosSample.from_dict(
+                masked_helios_sample_dict
+            )
             with torch.amp.autocast(device_type=device.type, dtype=torch.bfloat16):
                 # TODO: Model expects masked helios sample we need to pass empty masks
                 # Likely we want to have a flag that checks for eval mode and passes empty masks
                 batch_embeddings = model(
-                    masked_helios_sample, patch_size=model.base_patch_size
+                    masked_helios_sample, patch_size=patch_size
```
gabrieltseng, line 41: ah I looked for all calls to `base_patch_size` when I removed that attribute! thanks

PR #77, opened 2025-02-19
patch for default factory

Hgherzog:


scripts/latent_mim.py
```diff
@@ -52,8 +52,8 @@
     SUPPORTED_MODALITIES = [
         Modality.SENTINEL2,
         Modality.LATLON,
-        # Modality.SENTINEL1,
-        Modality.WORLDCOVER,
```
gabrieltseng, line 56: do we support worldcover in the encodings?

Hgherzog: yes we do based on another PR but I am now running into a small bug with grabbing the samples with worldcover that I need to fix real quick. Something changed from when Yawen merged in her big PR that I missed

PR #80, opened 2025-02-20
Add downstream eval in training loop

yawenzzzz:
- Added a callback to run downstream evaluation tasks (can be a list of tasks) during training. It currently supports classification but can be extended for other tasks
- Fixed the train/loss name issue
- Some clean up

### DIFF
--- a/helios/data/dataset.py
+++ b/helios/data/dataset.py
@@ -521,7 +521,6 @@ class HeliosDataset(Dataset):
         try:
             return self.normalizer_computed.normalize(modality, image)
         except Exception:
-            logger.warning(f"Failed to normalize {modality} using computed strategy")
             return self.normalizer_predefined.normalize(modality, image)
 
     def __getitem__(self, index: int) -> HeliosSample:


helios/data/normalize.py
```diff
@@ -18,8 +18,8 @@ class Strategy(Enum):
 
 # Normalization strategy for each modality
 NORMALIZE_STRATEGY = {
-    Modality.SENTINEL1: Strategy.PREDEFINED,
-    Modality.SENTINEL2: Strategy.PREDEFINED,
+    Modality.SENTINEL1: Strategy.COMPUTED,
```
Hgherzog: Why are we changing this? Should this be a constant or configruable for a specific run?

helios/nn/flexihelios.py
```diff
@@ -351,7 +351,8 @@ def _apply_encodings_per_modality(
         Returns:
             Tensor with encodings applied based on modality type
         """
-        logger.info(f"Applying encodings to modality {modality}")
```
Hgherzog, line 354: nit: make logger debug instead

helios/train/callbacks/evaluator_callback.py
```diff
@@ -0,0 +1,141 @@
...
+                f"Downstream evaluator {self.name} {METRIC_NAME} score: {val_result}"
+            )
+            return val_result
+        except Exception as e:
+            logger.error(f"Error during evaluation: {e}")
+            return 0
+
+
+@dataclass
+class DownstreamEvaluatorCallback(Callback):
```
Hgherzog, line 88: Suggestion: add a test for this

helios/train/loss.py
```diff
@@ -21,6 +21,10 @@
 class Loss(ABC):
     """Abstract base class for loss functions."""
 
+    def __init__(self, name: str) -> None:
```
Hgherzog: Suggestion: make this a class variable as all instances of a specific type of loss function should have the same name.

helios/train/train_module/latent_mim.py
```diff
@@ -184,7 +178,7 @@ def train_batch(self, batch: HeliosSample, dry_run: bool = False) -> None:
         decoded, loss = self.model_forward(masked_batch, patch_size)
 
         self.trainer.record_metric(
-            TRAIN_PATCH_DISC_LOSS_METRIC,
+            f"train/{self.base_loss.name}",  # get loss name
```
Hgherzog: suggestion: create a static method for train related metrics

helios/data/normalize.py
gabrieltseng: I think maybe configurable from a config? But in general I would suggest preferring the computed strategy and falling back to the predefined stats only if the computed stats don't exist.

helios/train/callbacks/evaluator_callback.py
```diff
@@ -0,0 +1,141 @@
...
+
+from helios.evals.datasets import GeobenchDataset
+from helios.evals.embeddings import get_embeddings
+from helios.evals.knn import run_knn
+
+logger = logging.getLogger(__name__)
+
+
+# Geobench classification
+METRIC_NAME = "f1"
```
gabrieltseng: The metric will actually be top-1 accuracy for eurosat. `run_knn` only returns the f1 score in the multilabel classification case (e.g. BigEarthNet).

See logic here: https://github.com/allenai/helios/blob/8806818a5aae5f9df7559a23e89dc02c94f1f6fe/helios/evals/knn.py#L37

helios/data/normalize.py
yawenzzzz: agree, I'm borrowing the pre-computed mean & std from Presto right now, we will need to compute those stats offline later, and update the numbers. OK, I ended up follow Gabi's suggestion, use computed strategy first and then falling back to predefined stats.

helios/nn/flexihelios.py
yawenzzzz: fixed.

helios/train/callbacks/evaluator_callback.py
yawenzzzz: These callbacks are tightly integrated with the trainer, so testing is not very straightforward. They also require real evaluators. I tested it with real runs, and it worked.

helios/train/loss.py
yawenzzzz: Fixed, thanks!

helios/train/train_module/latent_mim.py
yawenzzzz: I think it’s good to use `record_metric` for logging train loss within `train_batch`, as it’s the only metric we record. We can add a static method later if more metrics are needed.

helios/train/callbacks/evaluator_callback.py
yawenzzzz: Thanks for the clarification, updated the metric name!

PR #81, opened 2025-02-20
Henryh/add-grad-clipping

Hgherzog:


helios/train/train_module/train_module.py
```diff
@@ -433,44 +430,9 @@ def _clip_grad_norm(
...
-                dist.all_reduce(
-                    total_norm, op=dist.ReduceOp.SUM, group=pp_mesh.get_group()
-                )
-                total_norm **= 1.0 / norm_type
-
-        torch.nn.utils.clip_grads_with_norm_(
-            parameters, max_grad_norm, total_norm, foreach=foreach
+        # Pipeline parallel grad clipping required nightly torch
+        return torch.nn.utils.clip_grad_norm_(
+            self.model.parameters(), max_grad_norm, norm_type=norm_type, foreach=foreach
```
gabrieltseng, line 437: Can we make this a seperate PR so that grad clipping is generally available?

PR #82, opened 2025-02-20
Fix the patch disc loss

gabrieltseng:
I am very very very very sorry 

https://wandb.ai/eai-ai2/helios-debug/runs/9jvh82hy

### DIFF
--- a/tests/unit/train/test_loss.py
+++ b/tests/unit/train/test_loss.py
@@ -26,7 +26,47 @@ def test_patch_disc_loss() -> None:
     loss_value = loss.compute(preds, targets)
     # not very good! since they are all the same
     # predictions and values
-    assert loss_value > 1
+    assert loss_value > 0.5
+
+
+def test_patch_disc_loss_averaged_over_batch_size() -> None:
+    """Test it doesn't scale with batch size."""
+    b, t_h, t_w, t, d = 3, 4, 4, 2, 2
+
+    preds = TokensAndMasks(
+        sentinel2=torch.ones((b, t_h, t_w, t, d)),
+        sentinel2_mask=torch.ones((b, t_h, t_w, t)) * 2,
+        latlon=torch.ones((b, 1, d)),
+        latlon_mask=torch.ones((b, 1)) * 2,
+    )
+    targets = TokensAndMasks(
+        sentinel2=torch.ones((b, t_h, t_w, t, d)),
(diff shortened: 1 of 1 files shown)


helios/nn/flexihelios.py
```diff
@@ -1303,7 +1303,7 @@ class PredictorConfig(Config):
     num_heads: int = 2
     max_sequence_length: int = 12
     drop_path: float = 0.0
-    learnable_channel_embeddings: bool = False
+    learnable_channel_embeddings: bool = True
```
gabrieltseng, line 1306: I think this is important  - otherwise the predictor has no way of knowing which mask tokens are relevant for a particular channel group

helios/train/loss.py
```diff
@@ -138,7 +138,8 @@ def compute(
 
         # emulate averaging across the batch dimension
         loss_multiplier = self._expand_and_reciprocate(count)
-        loss = (loss * loss_multiplier).sum() / bs
+        # can't use bs here since this is after the unsqueezing, so bs == 1
+        loss = (loss * loss_multiplier).sum() / all_preds.shape[0]
```
gabrieltseng, line 142: :facepalm: :facepalm: :facepalm:

helios/train/train_module/latent_mim.py
```diff
@@ -46,9 +46,8 @@ class LatentMIMTrainModuleConfig(HeliosTrainModuleConfig):
     masking_config: MaskingConfig = field(
         default_factory=lambda: MaskingConfig(strategy_config={"type": "random"})
     )
-    ema_decay: float = 0.99
-    # TODO: currently causes runs to fail
-    # max_grad_norm: float = 1.0
+    ema_decay: float = 0.999
```
gabrieltseng, line 49: In the original codebase (and in LatentMIM and IJEPA) the ema decay linearly increases from 0.996 to 1.0 by the end of training. This is probably what we want too.

helios/train/loss.py
Hgherzog: Can we add a test here that ensures this loss does not scale with batch size

helios/train/train_module/latent_mim.py
Hgherzog: I will create an issue for this

helios/train/loss.py
gabrieltseng: added

PR #96, opened 2025-02-21
patch: torchvision to requirements

gabrieltseng:
This is needed for the transforms I think

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,5 +11,6 @@ pytest
 rasterio
 rioxarray
 torch>=2.5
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -11,5 +11,6 @@ pytest
 rasterio
 rioxarray
 torch>=2.5
+torchvision>=0.20
```
Hgherzog, line 14: We should have something in CI that raises an error if there is some sort of dependency conflict. I can make an issue

PR #99, opened 2025-02-21
Henryh/dataset visualizations

Hgherzog:
- add a hook to upload dataset distribution to wandb before training
Screenshot 2025-02-21 at 1 22 02 PM

### DIFF
--- a/helios/data/dataset.py
+++ b/helios/data/dataset.py
@@ -24,6 +24,7 @@ from upath import UPath
 from helios.data.constants import (
     BASE_RESOLUTION,
     IMAGE_TILE_SIZE,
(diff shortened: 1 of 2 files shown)


helios/data/dataset.py
```diff
@@ -292,6 +292,8 @@ def stack_or_none(attr: str) -> torch.Tensor | None:
 class HeliosDataset(Dataset):
     """Helios dataset."""
 
+    PROJECTION_CRS = "EPSG:4326"
```
yawenzzzz: Shall we move this to `constants.py`? There could be other things require this projection.

PR #100, opened 2025-02-22
Henryh/data sample visualization

Hgherzog:


helios/data/constants.py
```diff
@@ -136,6 +136,11 @@ def num_bands(self) -> int:
         """
         return sum(len(band_set.bands) for band_set in self.band_sets)
 
+    @property
+    def is_spatial(self) -> bool:
+        """Does the modality have spatial data."""
+        return self.get_tile_resolution() > 0
```
yawenzzzz, line 142: Shall we add `is_spatial` into `ModalitySpec`? in this way, we can use `is_spatial` and `is_multitemporal` for the following functions, instead of checking `self.get_tile_resolution()`.

helios/internal/experiment.py
```diff
@@ -186,11 +227,14 @@ def main(
...
     ],
+    visualize_config_builder: (
+        Callable[[CommonComponents], HeliosVisualizeConfig] | None
+    ) = None,
 ) -> None:
     """Main entry point for Helios experiments.
 
     overrides:  A list of field attributes with dot notation, e.g. ``foo.bar=1``.
-    Current usage: torchrun script.py train
+    Current usage: torchrun script.py train, python script.py visualize
```
yawenzzzz: I think we can remove this line as the following examples give clear instruction on the running subcommands.

helios/data/visualize.py
```diff
@@ -0,0 +1,180 @@
...
+    # Hide unused columns
+    for empty_col in range(max_bands - 1, 0, -1):
+        axes[0, empty_col].axis("off")
+
+    for row_idx, (modality_spec, modality_tile) in enumerate(modalities, start=1):
+        logger.info(f"Plotting modality: {modality_spec.name}")
+
+        # 4B. Plot other modalities
+        modality_data = dataset.load_sample(modality_tile, sample)
+        logger.info(f"Modality data shape (loaded): {modality_data.shape}")
```
yawenzzzz: As mentioned, it will be nicer to specify if applying normalization or not, for visualization purpose, we could safely use the `PREDEFINED` norm as it includes all the norm values (vs the `COMPUTED` one may be unavailable for some bands or some modalities), but note that for WORLDCOVER we will not need normalization here. The `COMPUTED` stats are mainly for pretraining purpose. 

Some code I tested just now:
```
from helios.data.normalize import Normalizer, Strategy
normalizer = Normalizer(strategy=Strategy.PREDEFINED)
if modality_spec != Modality.WORLDCOVER:
   modality_data = normalizer.normalize(modality_spec, modality_data)
```

For Sentinel1 data, we will also need to convert the values from `load_sample` to db using the function `convert_to_db` in `data/utils.py`, similar as: https://github.com/allenai/helios/blob/main/helios/data/dataset.py#L552

helios/evals/datasets/geobench_dataset.py
```diff
@@ -280,3 +289,53 @@ def collate_fn(
         )
         collated_target = default_collate([t for t in targets])
         return MaskedHeliosSample(**collated_sample), collated_target
+
+    def visualize_sample_bands(self, x: np.ndarray, output_dir: str) -> None:
```
yawenzzzz, line 312: You may want to double check the visualization for the geobench dataset, e.g., the image title.

helios/data/constants.py
Hgherzog: I think it is better as a property of Modality Spec then an extra defining field

PR #101, opened 2025-02-24
Modality shapes

pjreddie:
For talking about with Gabi,

This makes a tensor of zeros then adds in the embeddings sequentially, at the end adding this into the original tokens.

We probably need more tests I don't have a good understanding of what kind of data can get passed in here, but it passes the current tests. It does still rely on examining the input token shape to figure out if there are spatial and temporal dimensions which may need some reconfiguring.

### DIFF
--- a/helios/nn/flexihelios.py
+++ b/helios/nn/flexihelios.py
@@ -438,7 +438,9 @@ class FlexiHeliosCompositeEncodings(nn.Module):
 
         # Channel embeddings
         channel_embed = self.per_modality_channel_embeddings[modality.name]
-        channel_embed = repeat(channel_embed, f"b_s d -> {ein_string}", **ein_dict).to(device)
+        channel_embed = repeat(channel_embed, f"b_s d -> {ein_string}", **ein_dict).to(
+            device
+        )
         modality_embed[..., :n] += channel_embed
 
         if modality.is_multitemporal:


helios/data/constants.py
```diff
@@ -136,6 +136,11 @@ def num_bands(self) -> int:
         """
         return sum(len(band_set.bands) for band_set in self.band_sets)
 
+    @property
+    def is_temporal(self) -> bool:
```
Hgherzog: nit: this property seems redundant.

tests/unit/nn/test_flexihelios.py
```diff
@@ -4,17 +4,81 @@
...
+                Modality.LATLON,
+                Modality.WORLDCOVER,
+            ],
+            max_sequence_length=12,
+            use_channel_embs=True,
+            random_channel_embs=True,
+        )
+        return flexi_helios_composite_encodings
+
+    def test_apply_encodings_per_modality(
```
Hgherzog: Suggestion: Make a separate test per modality for readability and ease of debugging, ie if multiple fail we can know all at once.

helios/nn/flexihelios.py
```diff
@@ -408,85 +408,53 @@ def _apply_encodings_per_modality(
...
             Tensor with encodings applied based on modality type
         """
-        logger.debug(f"Applying encodings to modality {modality}")
-        if modality == Modality.LATLON.name:
-            return modality_tokens
         # TODO: Improve this implementation it is quite bad
+
+        modality = Modality.get(modality_name)
+        logger.debug(f"Applying encodings to modality {modality}")
+        d = self.embedding_dim_per_embedding_type
```
Hgherzog: nit: d is a little confusing here because it is different from the d used in the einops below. One option is to hide the einops in a helper function

helios/nn/flexihelios.py
```diff
@@ -599,7 +549,7 @@ def __init__(
 
         self.composite_encodings = FlexiHeliosCompositeEncodings(
```
Hgherzog, line 553: Random channel embeds should be able to be configurable from the encoder config and should be passed in here as well

helios/nn/flexihelios.py
pjreddie: I used `n` instead and put it closer to it's actual usage

PR #102, opened 2025-02-24
Make pooling and norm for eval task configurable

yawenzzzz:
For each downstream evaluation tasks, we can specify which pooling type to use (mean or max), and also which norm stats to use (either from pretraining dataset, or from downstream dataset). I also added the batch size and num of workers for the downstream dataset to speed up their data loader, now more frequent evaluation check is supported.

### DIFF
--- a/helios/evals/datasets/geobench_dataset.py
+++ b/helios/evals/datasets/geobench_dataset.py
@@ -105,6 +105,7 @@ class GeobenchDataset(Dataset):
         self.split = split
         self.partition = partition
         self.norm_stats_from_pretrained = norm_stats_from_pretrained
+        # If normalize with pretrained stats, we initialize the normalizer here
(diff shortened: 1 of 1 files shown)


helios/evals/datasets/geobench_dataset.py
```diff
@@ -14,6 +14,7 @@
 
 from helios.data.constants import Modality
 from helios.data.dataset import HeliosSample
+from helios.data.normalize import Normalizer, Strategy
 from helios.train.masking import MaskedHeliosSample
 
 torch.multiprocessing.set_sharing_strategy("file_system")
```
Hgherzog: Suggestion: call this in the init or class of the method where it is needed. Also, add a comment of what this is used for.

yawenzzzz: Fixed.

PR #103, opened 2025-02-24
Small fixes

gabrieltseng:


helios/train/train_module/latent_mim.py
```diff
@@ -161,6 +161,7 @@ def eval_loss_fn(self, logits: torch.Tensor, labels: torch.Tensor) -> torch.Tens
 
     def train_batch(self, batch: HeliosSample, dry_run: bool = False) -> None:
         """Train a batch."""
+        self.model.train()
```
gabrieltseng, line 164: The model was being put in eval mode for the eurosat evals. I might have missed it but I can't see it being put back into train mode.

helios/train/train_module/latent_mim.py
```diff
@@ -201,6 +190,18 @@ def train_batch(self, batch: HeliosSample, dry_run: bool = False) -> None:
...
+        decoded, loss = self.model_forward(masked_batch, patch_size)
+
+        self.trainer.record_metric(
+            f"train/{self.base_loss.name}",
+            loss / get_world_size(self.dp_process_group),
+            ReduceType.mean,
+        )
+
+        # Backpropagate and optimize
+        loss.backward()
```
gabrieltseng, line 208: Is there a reason we were making loss possibly None? If not I think its safer to not have that if condition

helios/train/train_module/latent_mim.py
Hgherzog: I think it is good to have this here explicitly even though it doesn't change things if we evaluate on target encoder

helios/train/train_module/latent_mim.py
Hgherzog: This is a vestige of olmo-core port no need for it to be none at this point.

scripts/latent_mim.py
```diff
@@ -88,7 +88,7 @@ def build_train_module_config(
     """Build the train module config for an experiment."""
     LR = 0.0001
     RANK_BATCH_SIZE = (
-        16  # TODO: maybe this should be computed dynamically and not specified here
```
Hgherzog, line 91: Probably better to leave these config changes out of the PR, so it is easy to merge into other working branches

gabrieltseng: undone

PR #105, opened 2025-02-25
Henryh/add gradient accumulation

Hgherzog:
Important things to note that may make grad accumulation not behave as expected: 
- All Disc loss with same global batch size but different micro-batch sizes result in different gradients, though this matches the implementation in gallileo. (This is not really a bug, just something we should be clear on the semantics of). 
- If the min hw is too low when subsampling, we may get micro-batches with uneven numbers of tokens making the loss for token averaged losses like l1 and l2 weight microbatches with less tokens relatively more. 

Also, DDP with AllDisc only contrasts within per rank microbatches right now. It is not clear whether we would even want to have a larger effective batch size (ie contrast across global tokens) as though this can be done efficiently it adds quite a bit of complexity and communication overhead.

### DIFF
--- a/helios/train/train_module/latent_mim.py
+++ b/helios/train/train_module/latent_mim.py
@@ -231,7 +231,9 @@ class LatentMIMTrainModule(HeliosTrainModule):
                 masked_batch = self.masking_strategy.apply_mask(subsampled_batch)
 
                 # Run Encoder and decoder on the augmented input
-                decoded, target_output = self.model_forward(masked_batch, patch_size, self.token_exit_cfg)
+                decoded, target_output = self.model_forward(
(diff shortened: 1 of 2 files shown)


helios/train/utils.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    batch_size = batch.batch_size
+
+    # If the batch is already small enough, no need to split.
+    if batch_size <= microbatch_size:
+        return [batch]
+
+    # Calculate how many micro-batches we need.
+    num_microbatches = (batch_size + microbatch_size - 1) // microbatch_size
+    microbatches = []
```
yawenzzzz, line 26: Maybe double check this calculation? It should be fine for most common batch sizes though.

Issue #94, opened 2025-02-21
Make Exit configs configurable


Hgherzog:
currently always none

PR #106, opened 2025-02-25
Token exit debugging

gabrieltseng:
Mostly adding tests to make sure gradients are flowing as we expected (and making fixes where those tests failed). 

https://wandb.ai/eai-ai2/helios-debug/runs/ucqydct2

### DIFF
--- a/helios/train/train_module/latent_mim.py
+++ b/helios/train/train_module/latent_mim.py
@@ -46,7 +46,7 @@ class LatentMIMTrainModuleConfig(HeliosTrainModuleConfig):
         default_factory=lambda: MaskingConfig(strategy_config={"type": "random"})
     )
     token_exit_cfg: dict[str, int] = field(
-        default_factory=lambda: {modality.name: 0 for modality in Modality.values()}
+        default_factory=lambda: {modality: 0 for modality in Modality.names()}
     )
     ema_decay: tuple[float, float] = (0.996, 1.0)
     max_grad_norm: float = 1.0


helios/nn/flexihelios.py
```diff
@@ -917,45 +913,42 @@ def apply_attn(
...
-            # skip the 0th block since this is just the linear
-            # projection
-            if (exit_ids_seq is not None) and (i_blk > 0):
+            if exit_ids_seq is not None:
                 assert exited_tokens is not None
                 # If a token should exit, then we update the exit token with the current token at the same position
                 exited_tokens = torch.where(
                     condition=(exit_ids_seq == i_blk),
-                    input=tokens.detach(),
-                    other=exited_tokens.detach(),
```
gabrieltseng, line 958: We should only ever call this in a `torch.no_grad()` context so I think this is okay.

scripts/latent_mim.py
```diff
@@ -177,7 +177,7 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
             name="m-eurosat",
             batch_size=128,
             num_workers=8,
-            pooling_type=PoolingType.MAX,
+            pooling_type=PoolingType.MEAN,
```
gabrieltseng, line 182: I have a preference for mean not max; when running evals, all models (except Satlas, which was explicitly trained for it) performed better with mean.

helios/nn/flexihelios.py
```diff
@@ -1036,7 +1026,7 @@ def __init__(
         num_heads: int = 8,
         max_sequence_length: int = 24,
         drop_path: float = 0.0,
-        learnable_channel_embeddings: bool = False,
+        learnable_channel_embeddings: bool = True,
```
Hgherzog, line 1033: I think default values should be only in the configs that we use to build the classes rather than in both places. This is a case where we had different defaults in two different places so I think it is better to just have them in the intended place from which the class will be used.

scripts/latent_mim.py
Hgherzog: I suggest we change the default to mean in the config then so that we keep that insight. This specific script likely won't be super long lived and we will have a bunch of similar scripts for different experiments

tests/integration/nn/test_latent_mim.py
```diff
@@ -0,0 +1,214 @@
...
+    # Shared constants for encoder and predictor
+    MAX_PATCH_SIZE = 8
+    NUM_HEADS = 2
+    MLP_RATIO = 4.0
+    MAX_SEQ_LENGTH = 12
+    DEPTH = 2
+    DROP_PATH = 0.1
+    ENCODER_EMBEDDING_SIZE = 16
+    DECODER_EMBEDDING_SIZE = 16
+    encoder = Encoder(
```
Hgherzog, line 101: Suggestion: Use the configs to build these classes, as it will then use the defaults we actually use when we launch a job. I know I didn't do that in the other tests but I think it would be an improvement to add in everywhere.

tests/integration/nn/test_flexihelios.py
```diff
@@ -276,6 +276,21 @@ def test_forward_exit_config_none(
             )
         ), f"Expected output latlon_mask shape {latlon_mask.shape}, got {output.latlon_mask.shape}"
 
+        # test the gradients are correct too
+        output.sentinel2.sum().backward()
+
+        for name, param in encoder.named_parameters():
+            # the composite_encodings is a bug which will be fixed now
```
Hgherzog, line 283: Is this bug fixed in this PR? Or in a separate PR?

tests/integration/nn/test_flexihelios.py
```diff
@@ -610,6 +669,109 @@ def test_predictor_forward(
...
+            "worldcover": worldcover,
+            "worldcover_mask": worldcover_mask,
+            "timestamps": timestamps,
+        }
+        x = MaskedHeliosSample(**masked_sample_dict)
+
+        patch_size = 4
+        input_res = 1
+        # Shared constants for encoder and predictor
+        MAX_PATCH_SIZE = 8
```
Hgherzog, line 746: Suggestion: Use configs to build classes for integration tests

helios/train/train_module/latent_mim.py
```diff
@@ -230,10 +230,10 @@ def model_forward(
                 target_output = self.model.target_encoder.forward(
                     batch.unmask(),
                     patch_size=patch_size,
-                    # token_exit_cfg={
-                    #     modality: 0
-                    #     for modality in self.model.encoder.supported_modality_names
-                    # },
+                    token_exit_cfg={
```
Hgherzog: I assume we don't actually want to merge this right now, as yawen is soon to make this configurable?

helios/nn/flexihelios.py
Hgherzog: Suggestion: adding a comment indicating this or  an assert, would help make that assumption more explicit

tests/integration/nn/test_flexihelios.py
gabrieltseng: This will be fixed by https://github.com/allenai/helios/pull/101

helios/nn/flexihelios.py
gabrieltseng: Are you suggesting removing default values from these classes?

I am not against that but it might be out of scope for this PR

tests/integration/nn/test_latent_mim.py
gabrieltseng: Okay. This might be best in a different test (e.g. a full integration test which tests a forward pass as defined by our configs).

In my mind these are sort of 0.5x integration tests - they test the integration of multiple components together, but I do want to isolate the components I am testing.

tests/integration/nn/test_flexihelios.py
gabrieltseng: Discussed: https://github.com/allenai/helios/pull/106#discussion_r1970253041

helios/train/train_module/latent_mim.py
gabrieltseng: I think its probably okay to merge this in? It will be an easy conflict for Yawen to resolve if necessary (cc @yawenzzzz ?)

helios/train/callbacks/evaluator_callback.py
```diff
@@ -36,7 +36,7 @@ def __init__(
         trainer: Trainer,
         batch_size: int = 128,
         num_workers: int = 8,
-        pooling_type: PoolingType = PoolingType.MAX,
+        pooling_type: PoolingType = PoolingType.MEAN,
```
gabrieltseng, line 39: As suggested in https://github.com/allenai/helios/pull/106#discussion_r1970216538

helios/nn/flexihelios.py
gabrieltseng: Added a comment

scripts/latent_mim.py
gabrieltseng: Updated

helios/nn/flexihelios.py
Hgherzog: Yea I am suggesting removing them, I think it is confusing to have them in two places. Should definitely be in a seperate PR.

helios/train/train_module/latent_mim.py
yawenzzzz: Very sorry for missing this discussion, I was reviewing the changes in this PR but forgot to refresh the page. I’ve pushed the small update to make token exit configurable #94.

yawenzzzz: Also, this PR looks good to me, just approved it!

PR #109, opened 2025-02-26
Henryh/add token throughput

Hgherzog:
I wanted to start getting some idea of our training throughput for different batch size token budget etc configurations. 
- I am using the token budget as an estimate so this is imprecise but i think it is still helpful as an initial stab. 

https://wandb.ai/eai-ai2/helios-debug/runs/k624r9nx

### DIFF
--- a/helios/data/dataset.py
+++ b/helios/data/dataset.py
@@ -100,14 +100,6 @@ class HeliosSample(NamedTuple):
                 attribute_shape += [Modality.get(attribute).num_band_sets]
             return attribute_shape
 
-    @property
-    def batch_size(self) -> int:
-        """Get the batch size of the data."""
-        if len(self.sentinel2.shape) == 5:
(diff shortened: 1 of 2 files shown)


helios/train/callbacks/speed_monitor.py
```diff
@@ -1,19 +1,63 @@
...
+            return
+
+        # We need token budget times encoder ratio and token budget times decoder ratio
+        if isinstance(batch, HeliosSample):
+            self._step_tokens_encoded = (
+                batch.batch_size * self._encoder_ratio * self._token_budget
+            )
+            self._step_tokens_decoded = (
+                batch.batch_size * self._decoder_ratio * self._token_budget
+            )
```
yawenzzzz, line 57: add the tokens for the target encoder too? given that it can take the whole unmasked image and also impact the token throughput.

PR #112, opened 2025-02-26
Henryh/add overides support

Hgherzog:
Adds the ability to overide different fields in the configuration. This is based on https://omegaconf.readthedocs.io/en/latest/structured_config.html#merging-with-other-configs 

I added some examples of the override syntax. 

Call outs:
- going forward we want configs to be made of simple types as much as possible https://omegaconf.readthedocs.io/en/latest/structured_config.html#simple-types 
- If not, we can usually make the class work when we build it. StrEnums can be very helpful for this.

### DIFF
--- a/scripts/latent_mim.py
+++ b/scripts/latent_mim.py
@@ -133,6 +133,7 @@ def build_dataloader_config(common: CommonComponents) -> HeliosDataLoaderConfig:
     """Build the dataloader config for an experiment."""
     # things should be set during building
     # TODO: Include collate function here
+
     NUM_WORKERS = 0
     NUM_THREADS = 0
     GLOBAL_BATCH_SIZE = 32
--- a/helios/train/callbacks/speed_monitor.py
+++ b/helios/train/callbacks/speed_monitor.py
@@ -1,19 +1,65 @@
 """Speed monitor callback for the trainer for Helios."""
 
+import logging
 import time
 from typing import Any
 
 from olmo_core.train.callbacks.speed_monitor import SpeedMonitorCallback
 
+from helios.data.dataset import HeliosSample
+from helios.train.train_module.latent_mim import LatentMIMTrainModule
 
-# FOR NOW FORGET ABOUT TOKENS AND STUFF
-# TODO: update this for V2
+logger = logging.getLogger(__name__)
+
+
 class HeliosSpeedMonitorCallback(SpeedMonitorCallback):
     """Speed monitor callback for the trainer for Helios."""
+
+    _total_tokens_encoded = 0
+    _total_tokens_decoded = 0
+    _total_tokens_target_encoder = 0
+
+    def pre_train(self) -> None:
+        """Pre-train callback for the speed monitor."""
+        super().pre_train()
+        train_module = self.trainer.train_module
+
+        if isinstance(train_module, LatentMIMTrainModule):
+            # Unwrap if the model is in DDP
+            model = train_module.model
+            self._token_budget = model.token_budget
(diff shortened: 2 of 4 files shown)


helios/internal/experiment.py
```diff
@@ -188,13 +217,13 @@ def prepare_environment(self) -> None:
...
-        # if get_local_rank() == 0:
-        #     print(config)
-        #     print(
-        #         "\n"
-        #         f"[b blue]Total parameters:[/]                {config.model.num_params:,d}\n"
-        #         f"[b blue]Non-embedding parameters:[/]        {config.model.num_non_embedding_params:,d}"
-        #     )
+        if get_local_rank() == 0:
+            print(config)
+            # print(
```
favyen2, line 223: Is there any reason not to print this?

scripts/latent_mim.py
```diff
@@ -150,12 +151,12 @@ def build_dataloader_config(common: CommonComponents) -> HeliosDataLoaderConfig:
 
 def build_dataset_config(common: CommonComponents) -> HeliosDatasetConfig:
     """Build the dataset config for an experiment."""
-    TILE_PATH = UPath("/weka/dfive-default/helios/dataset/20250212/")
-    DTYPE = np.dtype("float32")
+    TILE_PATH = "/weka/dfive-default/helios/dataset/20250212/"
+    TILE_PATH = "./test_data"
```
favyen2: Remove test line. Also, will there be a configuration file to set things like the tile path in the future?

helios/data/dataset.py
```diff
@@ -564,31 +565,44 @@ def __getitem__(self, index: int) -> HeliosSample:
 class HeliosDatasetConfig(Config):
     """Configuration for the HeliosDataset."""
 
-    tile_path: UPath
-    supported_modalities: list[ModalitySpec]
```
favyen2, line 575: Is it possible to provide ModalitySpec/UPath when providing configuration file (which may be in Python) but str when overriding? Equivalent of registering a type handler in jsonargparse.

This change makes it more difficult to configure it in Python. But if it is the only way then it is fine.

helios/dataset/utils.py
```diff
@@ -4,7 +4,7 @@
```
favyen2, line 4: It looks like this specific file is renamed from util.py to utils.py. To keep it consistent, the other util.py files like `dataset_creation/util.py` should be renamed as well.

Hgherzog: I missed that one, I noticed half of our stuff was utils half was util so just wanted to standardize.

helios/data/dataset.py
Hgherzog: the python configuration would just be  I should probably not use strings to configure it directly [Modality.WORLDCOVER.name, Modality.Sentinel2.name] does this work for you? UPath should be fine as it will get wrapped before its built anyways. I will look into seeing if we could do like a =[Modality.Sentinel2] type overide.

favyen2: I looked at the omegaconf doc a bit, it sounds like this is the way they want to do it, the way you have it in the PR looks fine to me. The Modality.SENTINEL2.name is fine too.

scripts/latent_mim.py
Hgherzog: I think each experiment will just be configured in a single python script, rather than loading from yaml.

helios/internal/experiment.py
Hgherzog: I don't think these are implemented here we would need to add some calculations for this as this is before the model is built.

PR #114, opened 2025-02-27
Fix class builders

gabrieltseng:
Small fixes - otherwise we were getting errors like 

```
Uncaught TypeError: Encoder.__init__() got an unexpected keyword argument 'supported_modality_names'
```

### DIFF
--- a/tests/unit/nn/test_flexihelios.py
+++ b/tests/unit/nn/test_flexihelios.py
@@ -7,10 +7,12 @@ from einops import repeat
 from helios.data.constants import Modality, ModalitySpec
 from helios.nn.flexihelios import (
     Encoder,
+    EncoderConfig,
     FlexiHeliosBase,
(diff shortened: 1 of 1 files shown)


helios/data/dataset.py
```diff
@@ -612,4 +612,8 @@ def build(self) -> "HeliosDataset":
         """Build the dataset."""
         self.validate()
         kwargs = self.as_dict(exclude_none=True, recurse=False)
-        return HeliosDataset(tile_path=self.tile_upath, **kwargs)
+        kwargs["tile_path"] = self.tile_upath
+        kwargs.pop("supported_modality_names")
+        kwargs["supported_modalities"] = self.supported_modalities
+        logger.info(f"HeliosDataset kwargs: {kwargs}")
+        return HeliosDataset(**kwargs)
```
Hgherzog, line 619: I think we may want to replace tile_path with tile_upath as well

PR #115, opened 2025-02-27
MADOS eval set

gabrieltseng:
Adds MADOS via (segmentation) linear probing.

A todo is to improve support beyond S2 when doing the spatial pooling but this PR is otherwise functional.

### DIFF
--- a/scripts/latent_mim.py
+++ b/scripts/latent_mim.py
@@ -180,7 +180,15 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
             num_workers=8,
             pooling_type=PoolingType.MEAN,
             norm_stats_from_pretrained=True,
-        )
+        ),
+        DownstreamTaskConfig(
+            dataset="mados",
+            batch_size=128,
+            num_workers=8,
+            pooling_type=PoolingType.MEAN,
+            norm_stats_from_pretrained=False,
+            probe_lr=0.1,
+        ),
     ]
     # Let us not use garbage collector fallback
     trainer_config = (


helios/train/callbacks/evaluator_callback.py
```diff
@@ -70,7 +68,7 @@ def _get_embeddings(self, data_loader: DataLoader) -> tuple:
         return get_embeddings(
             data_loader=data_loader,
             model=self.trainer.train_module.model.encoder,
-            patch_size=self.trainer.train_module.model.encoder.max_patch_size,
+            patch_size=self.patch_size,
```
gabrieltseng, line 74: This might (partly) explain our performance on Eurosat

helios/evals/embeddings.py
```diff
@@ -42,7 +44,16 @@ def get_embeddings(
...
-            embeddings.append(batch_embeddings.pool_unmasked_tokens(pooling_type).cpu())
+            if task_type == TaskType.CLASSIFICATION:
+                averaged_embeddings = batch_embeddings.pool_unmasked_tokens(
+                    pooling_type
+                )
+            else:
+                # hackey way to take a mean across T and Band sets for S2
+                averaged_embeddings = torch.mean(
+                    batch_embeddings.sentinel2_l2a, dim=(-2, -3)
+                )
```
gabrieltseng: Highlighting this as something I'm unhappy with. Ideally there should be a `pool_unmasked_tokens(pooling_type, keep_spatial_dims=True)` syntax but this makes the `pool_unmasked_tokens` more complicated.

gabrieltseng: We can simplify a lot if we assume that when doing spatial pooling, the mask within a modality is the same (e.g. we always have the same mask for all of S2).

gabrieltseng: okay fixed following this assumption in https://github.com/allenai/helios/pull/115/commits/bebcaa7ff5aea8c99b24612888ae3c7866d1748b, run: https://beaker.allen.ai/orgs/ai2/workspaces/earth-systems/work/01JP7RXN6T28HCBJ8CY9KK89NK?taskId=01JP7RXN6ZA313GM1WQH2GGT3P&jobId=01JP7RXNC6V0Q0N0D2XEQR3RZ3

helios/nn/flexihelios.py
```diff
@@ -145,26 +145,57 @@ def flatten_tokens_and_masks(self) -> tuple[Tensor, Tensor]:
...
+                return x_for_pooling.max(dim=1).values
+            elif pooling_type == PoolingType.MEAN:
+                return x_for_pooling.sum(dim=1) / torch.sum(mask, -1, keepdim=True)
+            else:
+                raise ValueError(f"Invalid pooling type: {pooling_type}")
         else:
-            raise ValueError(f"Invalid pooling type: {pooling_type}")
+            spatial_average = []
+            for attr_name in self.modalities:
+                if Modality.get(attr_name).get_tile_resolution() > 0:
```
yawenzzzz: you can use `Modality.get(attr_name).is_spatial`, and skip the following comment

helios/train/callbacks/evaluator_callback.py
```diff
@@ -117,27 +146,31 @@ def post_step(self) -> None:
...
+    dataset: str
     batch_size: int = 128
     num_workers: int = 8
     pooling_type: PoolingType = PoolingType.MEAN
     norm_stats_from_pretrained: bool = True
+    # for MADOS and a default partition, the following lrs
+    # did best for Galileo:
+    # ViT-nano = 0.8 or 0.5
+    # ViT-tiny = 0.1
+    # ViT-base = 0.01
```
yawenzzzz, line 171: Besides lr, do we need to adjust batch size for different ViT models?

helios/evals/datasets/geobench_dataset.py
```diff
@@ -144,9 +98,7 @@ def __init__(
         self.band_indices = [
             original_band_names.index(band_name) for band_name in self.band_names
         ]
-        imputed_band_info = self._impute_normalization_stats(
-            task.band_stats, config.imputes
-        )
+        imputed_band_info = impute_normalization_stats(task.band_stats, config.imputes)
```
yawenzzzz, line 101: I got a little confused about this `impute_normalization_stats`, seems EuroSat has no imputes but this function is applied, while MADOS has imputes but the `BAND_STATS` from 13 bands are complete and this function is not applied, why is it? Also, does the imputes follow any rule or just impute using nearby band mean & std?

helios/evals/datasets/mados_dataset.py
```diff
@@ -0,0 +1,289 @@
...
+from helios.data.constants import Modality
+from helios.data.dataset import HeliosSample
+from helios.train.masking import MaskedHeliosSample
+
+from .constants import EVAL_S2_BAND_NAMES, EVAL_TO_HELIOS_S2_BANDS
+from .normalize import normalize_bands
+
+torch.multiprocessing.set_sharing_strategy("file_system")
+
+MADOS_DIR = UPath("/weka/dfive-default/presto_eval_sets/mados")
```
yawenzzzz, line 25: just fyi, I downloaded the PASTIS data here: https://weka.allen.ai/b/dfive-default/helios/eval_dataset/

helios/evals/datasets/mados_dataset.py
```diff
@@ -0,0 +1,289 @@
...
+    "08 - NIR": {"mean": 0.03185819461941719, "std": 0.03786701336503029},
+    "08A - Vegetation Red Edge": {
+        "mean": 0.0348929800093174,
+        "std": 0.04325847327709198,
+    },
+    "09 - Water vapour": {"mean": 0.0348929800093174, "std": 0.04325847327709198},
+    "10 - SWIR - Cirrus": {"mean": 0.029205497354269028, "std": 0.0350610613822937},
+    "11 - SWIR": {"mean": 0.023518012836575508, "std": 0.02930651605129242},
+    "12 - SWIR": {"mean": 0.017942268401384354, "std": 0.022308016195893288},
+}
```
yawenzzzz, line 57: the mean and std values are much smaller than our norm stats: https://github.com/allenai/helios/blob/main/data/norm_configs/computed.json, are these mean & std computed from normalized values?

helios/evals/datasets/__init__.py
```diff
@@ -1,5 +1,43 @@
...
+        raise ValueError(f"Unrecognized dataset {eval_dataset}")
+
+    if eval_dataset.startswith("m-"):
+        # m- == "modified for geobench"
+        return GeobenchDataset(
+            geobench_dir=GEOBENCH_DIR,
+            dataset=eval_dataset,
+            split=split,
+            partition=partition,
+            norm_stats_from_pretrained=norm_stats_from_pretrained,
```
yawenzzzz, line 31: Have you noticed better performance on EuroSat when using its own normalizer instead of the pretraining dataset’s?

helios/evals/datasets/mados_dataset.py
```diff
@@ -0,0 +1,289 @@
...
+        240,
+    ), "Image tensor must be of shape (13, 240, 240)"
+    assert label_tensor.shape == (240, 240), "Label tensor must be of shape (240, 240)"
+
+    tile_size = 80
+    tiles = []
+    labels = []
+
+    for i in range(3):
+        for j in range(3):
```
yawenzzzz, line 84: Suggestion: Add `tile_size` as an argument and use it to compute `new_height` and `new_width`. This could make the function useful for other tasks as well, but we can do it later.

helios/train/callbacks/evaluator_callback.py
gabrieltseng: Modifying lr is similar to modifying batch size - I think its easier to modify only one parameter for these evals.

helios/evals/datasets/mados_dataset.py
gabrieltseng: This is because the MADOS dataset uses Rayleigh reflectance values (see Section 3.1.1. of the [paper](https://www.sciencedirect.com/science/article/pii/S0924271624000625#s0025)). 

This is why I added the warning here - https://github.com/allenai/helios/blob/faf91241e3244ddd5cdc1d45c9da1486a9e6f425/helios/evals/datasets/__init__.py#L36

helios/evals/datasets/__init__.py
gabrieltseng: I mostly added this argument because of the different processing in MADOS, as noted in https://github.com/allenai/helios/pull/115#discussion_r1995899947

helios/evals/datasets/mados_dataset.py
gabrieltseng: okay - lets do that if we have another use case and want to make this function more general. For now this code was only run once to save the data into WEKA, so it would require rerunning things.

Also all the evals in Galileo assume this tile size so all the model evals would need to be rerun if we change it for MADOS.

helios/nn/flexihelios.py
gabrieltseng: Updated

helios/evals/datasets/geobench_dataset.py
gabrieltseng: This is because the GeoBench dataset is general beyond EuroSat - for example m-BigEarthNet requires imputes. This anticipates the those impute-requiring datasets.

PR #117, opened 2025-02-27
space structured masking and tests

pjreddie:
Space structured masking working but it doesn't handle non-spatial data. It seems like the same masking strategy is applied to all data in a group. Thus I'm not sure how to actually handle this since some data may not be spatial (e.g. latlon). One option is if non-spatial data is passed to apply_mask for spatial masking it just defaults back to random masking? I'm not sure what the right thing to do is here.

### DIFF
--- a/helios/train/callbacks/speed_monitor.py
+++ b/helios/train/callbacks/speed_monitor.py
@@ -60,6 +60,7 @@ class HeliosSpeedMonitorCallback(SpeedMonitorCallback):
         self._total_steps += 1
         self._total_tokens_encoded += self._step_tokens_encoded
         self._total_tokens_decoded += self._step_tokens_decoded
+        self._total_tokens_target_encoder += self._step_tokens_target_encoder
 
     def post_step(self) -> None:
         """Post-step callback for the speed monitor."""


helios/train/masking.py
```diff
@@ -221,42 +258,216 @@ def __init__(
...
+                    if temporal_mask is None:
+                        temporal_mask = self._create_temporal_mask(
+                            Modality.get(modality_name), shape, device
+                        )
+                    t, b_s = shape[-2:]
+                    b, h, w = list(shape[:-2]) + [1] * (3 - len(shape[:-2]))
+                    mask = repeat(
+                        temporal_mask, "b t -> b h w t b_s", h=h, w=w, b_s=b_s
+                    )
+                    mask = mask.view(*shape)
```
gabrieltseng, line 335: nit: what is the purpose of this line? In all cases (currently), `shape` will be `b h w t b_s` so I don't think (?) it will do anything? And if the shape is different (e.g. no `h w`) then this would fail.

Similar - I don't fully understand line 336. What is the purpose of `[1] * (3 - len(shape[:-2]))`?

pjreddie: the view call will change the dimensional layout but won't change the contents, it has to already have the correct number of elements. the repeat function call ensures that the tensor has the appropriate number of elements before it is re-viewed to the correct shape.

line 335-336 are my way of extracting dimensions regardless of input shape, replacing what we used to do with:

    if n.dim==4:
        ...
    if n.dim == 5:
        ...
    etc.

line 336 in particular infills with 1s so that if `h` and `w` don't exist in the input shape it will set them to 1. eventually (I hope) all of this will be unnecessary because all the input will have the same number of dimensions but this handles it until then.

shape is not always `b h w t b_s`, this is basically to handle that fact.

gabrieltseng: Ah - ok I understand. And then the `.view` removes those added dimensions

pjreddie: actually, I'm wrong, with our current modalities shape always is `b h w t b_s`. I'm not sure this is necessarily the case tho, other parts of our code assume the shape could be something like `b t b_s`, i.e. temporal but not spatial, so I wrote this with that in mind.

gabrieltseng: yea, that makes sense to me (to anticipate a `b t c` input). Worse case we remove 2 lines of code in the future but this does not significantly affect the readability of the code.

PR #126, opened 2025-03-03
small-fixes-to-galileo

Hgherzog:
build common components should always handle a couple of things so making sure we  use the updated version

### DIFF
--- a/helios/internal/experiment.py
+++ b/helios/internal/experiment.py
@@ -25,6 +25,7 @@ from helios.data.normalize import Normalizer, Strategy
 from helios.data.visualize import visualize_sample
 from helios.nn.latent_mim import LatentMIMConfig
 from helios.train.train_module.latent_mim import LatentMIMTrainModuleConfig
+from helios.train.train_module.train_module import HeliosTrainModuleConfig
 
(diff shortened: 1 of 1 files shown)


helios/internal/experiment.py
```diff
@@ -72,10 +72,10 @@ class HeliosExperimentConfig(Config):
 
     run_name: str
     launch: BeakerLaunchConfig
-    model: LatentMIMConfig  # TODO: make this agnostic to training setup
+    model: Any
```
favyen2: May make sense to use Config here over Any.

helios/internal/experiment.py
```diff
@@ -72,10 +72,10 @@ class HeliosExperimentConfig(Config):
...
     dataset: HeliosDatasetConfig  # will likely be fixed for us
     data_loader: HeliosDataLoaderConfig  # will likely be fixed for us
-    train_module: LatentMIMTrainModuleConfig  # we will want to support different train module model combinations
+    train_module: Any
```
yawenzzzz: similarly, Any -> HeliosTrainModuleConfig

PR #128, opened 2025-03-04
Compute offline norm stats and add data distributions

yawenzzzz:
1. Compute normalization values for each modality separately in a streaming manner
2. Use 1000 samples to generate data distribution (per modality per band), our original normalization values for B09 band in Sentinel2 L2A is incorrect (I grab it from EuroSat normalization dict), B09 band values after normalization are much larger than 1

Run with incorrect norm values for Sentinel2 L2A B09: https://wandb.ai/eai-ai2/helios-debug/runs/kwr4sw7a?nw=nwuseryawenz
Run with updated norm values for Sentinel2 L2A B09: https://wandb.ai/eai-ai2/helios-debug/runs/esdff6na?nw=nwuseryawenz

Issue #92

### DIFF
--- a/helios/data/dataset.py
+++ b/helios/data/dataset.py
@@ -540,6 +540,7 @@ class HeliosDataset(Dataset):
         sample_data: dict[str, Any] = {}
 
         # Assume samples could include different modalities and bands
+        # TODO: compute the histogram for each modality and band directly
         for i in tqdm(indices_to_sample):
             sample = self[i]
             for modality in sample.modalities:


helios/data/dataset.py
```diff
@@ -513,6 +520,41 @@ def get_geographic_distribution(self) -> np.ndarray:
         latlons = np.vstack(latlons)
         return latlons
 
+    def get_data_distribution(self, num_samples: int = 100) -> dict[str, Any]:
```
favyen2: Can you add some documentation for what the "data distribution" is? As far as I can tell, it is sampling some data and concatenating all of it, which is not usually what one thinks of for "data distribution".

Also, in plot_modality_data_distribution, it seems it is sampling just 1000 values for each band. In that case, here you could sample 1000/num_samples values from each sample or something like that.

A better alternative might be to compute the histogram in this function `get_data_distribution`. That would be closer to what one would expect "data distribution" to be, and it would substantially reduce the memory utilization. For uint8/uint16 data it is easy to store histogram for every possible value, but I guess it can be difficult for int32/float32 data since you may not know the min/max ahead of time.

tests/unit/dataset/test_compute_norm_stats.py
```diff
@@ -0,0 +1,59 @@
+"""Test the HeliosDataset class compute_norm_stats method."""
+
+import logging
+
+import numpy as np
+
+logger = logging.getLogger(__name__)
+
+
+def test_compute_norm_stats() -> None:
```
favyen2: This test does not seem to use any code in helios. I would either remove the test (since I don't see a good reason to maintain tests that do not test the actual code in helios) or refactor it to see how you can use the dataset (alternatively, you can try to re-factor out the portion that computes streaming stats given a stream of numpy arrays containing values).

yawenzzzz: Yes, this test is not quite right. As an update, I extracted the streaming stats part and changed the test to just test that algorithm.

helios/data/dataset.py
yawenzzzz: Thanks for the comment! I updated the function name to indicate that this step is only sample data for visualization purpose, I also moved get sample values per modality per band to this function, so that we can sample dataset & sample band data in the same place, and it did make the plotting faster. I added a TODO on computing the histogram directly, will do that later. With the new change (by sampling the values before concatenating them), there will be only 10K data points per modality per band, so the memory consumption should be good.

PR #130, opened 2025-03-06
Add adjusted InfoNCE loss

yawenzzzz:
Reference: https://proceedings.neurips.cc/paper_files/paper/2023/file/48aaa5ea741ae8430bd58e25917d267d-Paper-Conference.pdf

Adding weights to the negative pairs to prevent very hard negatives from getting very large weights.
Screenshot 2025-03-05 at 11 26 19 AM

Screenshot 2025-03-06 at 8 03 35 AM

Sample run: https://wandb.ai/eai-ai2/helios-debug/runs/gpzr9lih?nw=nwuseryawenz

### DIFF
--- a/helios/evals/finetune.py
+++ b/helios/evals/finetune.py
@@ -1,1 +1,0 @@
-"""Finetune the Helios model on a downstream task."""


helios/train/loss.py
```diff
@@ -122,6 +123,103 @@ def compute(
...
+        count = (all_masks == MaskValue.DECODER.value).sum(dim=-1)
+
+        if self.mask_other_samples:
+            logit_mask = torch.full_like(scores, -torch.finfo(scores.dtype).max)
+            start = 0
+            for c in count:
+                end = start + c
+                logit_mask[:, start:end, start:end] = 0
+                start += c
+            logger.info(f"logit_mask: {logit_mask.shape}")
```
favyen2: If this will print on every batch, it may be better to make it logger.debug.

helios/train/loss.py
```diff
@@ -122,6 +123,103 @@ def compute(
...
+            pred = (pred - pred_mu) / (pred_std + 1e-4)
+
+        pred = F.normalize(pred, p=2, dim=-1)
+        target = F.normalize(target, p=2, dim=-1)
+
+        scores = torch.einsum("npd,nqd->npq", pred, target) / self.tau
+        count = (all_masks == MaskValue.DECODER.value).sum(dim=-1)
+
+        if self.mask_other_samples:
+            logit_mask = torch.full_like(scores, -torch.finfo(scores.dtype).max)
```
favyen2: If there is a reason to use -torch.finfo(...).max here and -float("inf") below, then it may be useful to add a comment.

Otherwise, it could be better to be consistent (choose one to use in both places).

helios/train/loss.py
Hgherzog: we should switch to debug in the other loss function as well where this was copied from

yawenzzzz: Previously I ran into OOM with our patch discrimination loss implementation, because we're creating a big matrix of similarity scores between any two samples across the whole batch, which is not necessary (this is needed by AllDisc loss though). I updated the implementation to compute similarity matrix per sample and then apply weights per sample, the `logger.info` has been removed.

helios/train/loss.py
yawenzzzz: Same as above, the function has been updated.

Issue #135, opened 2025-03-09
Skip batch when loss goes to nan


yawenzzzz:
There could be very similar patches that lead to loss goes to nan, change `train_batch` to stop gradient updates on those anomalous batches.

PR #136, opened 2025-03-09
Skip gradient updates for bad batches

yawenzzzz:
Issue #135

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -27,7 +27,7 @@ jobs:
         pip install .
         git clone https://github.com/allenai/OLMo-core.git
         cd OLMo-core
-        git checkout v2 && pip install .
+        git checkout main && pip install .
 
 
     - name: Run tests


helios/internal/experiment.py
```diff
@@ -198,7 +198,7 @@ def launch(config: HeliosExperimentConfig) -> None:
     """Launch an experiment."""
     logger.info("Launching the experiment")
     logger.info(config)
-    config.launch.launch(follow=True)
+    config.launch.launch(follow=False)
```
gabrieltseng: what is the change for?

scripts/galileo.py
```diff
@@ -35,7 +35,7 @@
 
 logger = logging.getLogger(__name__)
 # TODO: Need to use the dynamic computation from trainer for this
-STEPS_PER_EPOCH = 25
+STEPS_PER_EPOCH = 100
```
gabrieltseng: is this accurate? it would be nice to calculate this on the fly, but this is out of scope for this PR

scripts/galileo.py
```diff
@@ -205,6 +205,13 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
             pooling_type=PoolingType.MEAN,
             norm_stats_from_pretrained=True,
         ),
+        DownstreamTaskConfig(
```
gabrieltseng: I think it might be not necessary to run these evals every N steps, especially as we add more. We may want to be able to run evals at varying frequencies (e.g. a quick one like Eurosat more frequently, and then others like Pastis less often).

helios/internal/experiment.py
yawenzzzz: This change is for preventing the "stream the logs to your console until you cancel", this is mainly for sweeping runs as I don't want to see logs from all runs. Let me change it back to True, as for single run it still make sense to stream the logs and make sure it's working smoothly.

scripts/galileo.py
yawenzzzz: This is inaccurate. Yes, I updated this part to calculate steps per epoch automatically, now we only need to set `eval_duration` in epoch unit in the scripts, by default, it's set to 5 epochs.

scripts/galileo.py
yawenzzzz: Yes, the frequencies has been reduced for evaluation set, now default to 5 epochs, but as you mentioned, it will be nice to have a varying `eval_duration` per task, which requires changes in `DownstreamEvaluatorCallback`, will do that in a later PR.

PR #137, opened 2025-03-10
Band masking

gabrieltseng:
Mask out entire modalities. In Galileo we masked out band sets instead of modalities, but I think this is much simpler and also makes sense as an approach? 

In general I think there are pros and cons; modalities are much coarser (i.e. larger sets of tokens are masked or unmasked) but it makes more intuitive sense to me to (e.g.) reconstruct S1 from S2 than to reconstruct 20m S2 from 10m S2.

### DIFF
--- a/helios/train/masking.py
+++ b/helios/train/masking.py
@@ -475,25 +475,28 @@ class ModalityMaskingStrategy(MaskingStrategy):
         """
         output_dict: dict[str, ArrayTensor | None] = {"timestamps": batch.timestamps}
 
-        present_bands = list(batch.as_dict(ignore_nones=True).keys())
-        present_bands = [b for b in present_bands if b != "timestamps"]
+        present_modalities = list(batch.as_dict(ignore_nones=True).keys())
+        present_modalities = [b for b in present_modalities if b != "timestamps"]
 
-        num_present_bands = len(present_bands)
-        encode_bands = max(1, int(self.encode_ratio * num_present_bands))
-        decode_bands = max(1, int(self.decode_ratio * num_present_bands))
-        target_bands = num_present_bands - encode_bands - decode_bands
+        num_present_modalities = len(present_modalities)
+        encode_modalities = max(1, int(self.encode_ratio * num_present_modalities))
+        decode_modalities = max(1, int(self.decode_ratio * num_present_modalities))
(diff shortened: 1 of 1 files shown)


helios/train/masking.py
```diff
@@ -447,6 +447,72 @@ def apply_mask(
...
+        Args:
+            batch: Input data of type HeliosSample
+            **kwargs: Additional arguments for maskings
+
+        Returns:
+            MaskedHeliosSample containing the masked data and mask
+        """
+        output_dict: dict[str, ArrayTensor | None] = {"timestamps": batch.timestamps}
+
+        present_bands = list(batch.as_dict(ignore_nones=True).keys())
```
yawenzzzz: `present_modalities` maybe more accurate than `present_bands`, as we're random masking modalities here.

helios/train/masking.py
```diff
@@ -447,6 +447,72 @@ def apply_mask(
...
+        present_bands = [b for b in present_bands if b != "timestamps"]
+
+        num_present_bands = len(present_bands)
+        encode_bands = max(1, int(self.encode_ratio * num_present_bands))
+        decode_bands = max(1, int(self.decode_ratio * num_present_bands))
+        target_bands = num_present_bands - encode_bands - decode_bands
```
yawenzzzz: Same comment as above, regarding `bands`.

helios/train/masking.py
```diff
@@ -475,6 +541,44 @@ def apply_mask(
             return self.time_strategy.apply_mask(batch, **kwargs)
 
 
+@MASKING_STRATEGY_REGISTRY.register("modality_space_time")
+class ModalitySpaceTimeMaskingStrategy(MaskingStrategy):
+    """Randomly select space or time masking and apply it to the input data."""
```
yawenzzzz: Update this docstring to reflect the selection from space, time and modality.

helios/train/masking.py
gabrieltseng: updated

helios/train/masking.py
gabrieltseng: updated

helios/train/masking.py
gabrieltseng: Updated

PR #140, opened 2025-03-12
Speed up iteration

yawenzzzz:
A few changes to speed up iteration: 

- Per dataset & filtering rule, it will generate one set of h5 files at epoch 1, and at later epochs, the h5 can be directly loaded. Here's the ~100K h5 files on WEKA: https://weka.allen.ai/b/dfive-default/helios/dataset/presto/, the folder is under the `tile_path` and named as `len(self.samples)` so that when different filtering rule applies, the folder name can be differentiated. Each h5 file (contains one HeliosSample) is 25MB, I also tried copying h5 to local disk, but soon ran into "no space on disk" error, this approach may speed up the data loading, but once the job is preempted, we will need to generate another copy, which is not very good.  

- Based on the logging information, each h5 file took about 0.02s to load from WEKA, but it can depends on the network connection etc. 

- Tried different combinations of [`GLOBAL_BATCH_SIZE`, `MICRO_BATCH_SIZE`, `NUM_WORKERS`], and the most stable and fast one I got is [512, 64, 4] with `num_gpus` = 8, with this set up, one epoch will take about 18min on Jupiter. Previously, with 1 GPU and `GLOBAL_BATCH_SIZE` = 128, it took ~ 1 to 2 hours. Our model is small right now, so data loading currently is still using the majority of time, some TODOs include: (1) exploring ways to reduce memory overhead so that we can use more workers, e.g., move subsetting to collate function, (2) saving worker information for the next epoch (`persistent_worker = True` is not actually working)

- Updated the PatchDisc loss to support that when `mask_other_samples` is True, we only compute similarity scores within a sample instead of producing a large tensor of similarity scores across all tokens within a batch.

### DIFF
--- a/helios/data/dataloader.py
+++ b/helios/data/dataloader.py
@@ -407,7 +407,7 @@ class HeliosDataLoaderConfig(Config):
     num_threads: int | None = None
     num_workers: int = 0
     prefetch_factor: int | None = None
-    target_device_type: str = get_default_device().type
+    target_device_type: str | None = None
     drop_last: bool = True
 
     def validate(self) -> None:
@@ -444,7 +444,7 @@ class HeliosDataLoaderConfig(Config):
             num_threads=self.num_threads,
             num_workers=self.num_workers,
             prefetch_factor=self.prefetch_factor,
-            target_device_type=self.target_device_type,
+            target_device_type=self.target_device_type or get_default_device().type,
             collator=collator,
             drop_last=self.drop_last,
         )


helios/data/dataset.py
```diff
@@ -666,24 +684,77 @@ def normalize_image(self, modality: ModalitySpec, image: np.ndarray) -> np.ndarr
...
             sample_modality = sample.modalities[modality]
             image = self.load_sample(sample_modality, sample)
-            # Convert Sentinel1 data to dB
+
             if modality == Modality.SENTINEL1:
                 image = convert_to_db(image)
-            # Normalize data and convert to dtype
+
             if self.normalize:
                 image = self.normalize_image(modality, image)
```
gabrieltseng: I think it would be better to save the unnormalized data in the h5py files. This way, if we want to modify the normalization method (or update the stats) we don't need to re-save the whole dataset.

yawenzzzz: Right, I was going to change it to unnormalized this morning, mark this one as WIP as I was testing if it works or not.

helios/data/dataset.py
```diff
@@ -666,24 +686,77 @@ def normalize_image(self, modality: ModalitySpec, image: np.ndarray) -> np.ndarr
         except Exception:
             return self.normalizer_predefined.normalize(modality, image)
 
-    def __getitem__(self, index: int) -> HeliosSample:
-        """Get the item at the given index."""
-        sample = self.samples[index]
+    # def __getitem__(self, index: int) -> HeliosSample:
```
pbeukema: remove commented code.

helios/data/dataset.py
```diff
@@ -666,24 +686,77 @@ def normalize_image(self, modality: ModalitySpec, image: np.ndarray) -> np.ndarr
...
+            h5_path = os.path.join(self.h5py_dir, f"{sample_id}.h5")
+
+            if sample_id in self.h5py_files:
+                continue  # Skip if already saved
+
+            with h5py.File(h5_path, "w") as h5file:
+                sample_data = self._load_sample(sample)
+                for key, value in sample_data.as_dict().items():
+                    if value is not None:
+                        h5file.create_dataset(key, data=value, compression="gzip")
```
pbeukema: not sure if this is already achieving the necessary speed gains, but one thing might be worth doing is chunking at the same level as the windows, so that you can retrieve chunks faster.

helios/train/loss.py
```diff
@@ -94,31 +94,61 @@ def compute(
         pred = F.normalize(pred, p=2, dim=-1)
         target = F.normalize(target, p=2, dim=-1)
 
-        scores = torch.einsum("npd,nqd->npq", pred, target) / self.tau
+        # scores = torch.einsum("npd,nqd->npq", pred, target) / self.tau
         count = (all_masks == MaskValue.DECODER.value).sum(dim=-1)
 
+        # if self.mask_other_samples:
```
pbeukema: commented code.

scripts/latent_mim.py
```diff
@@ -132,9 +132,9 @@ def build_dataloader_config(common: CommonComponents) -> HeliosDataLoaderConfig:
     # things should be set during building
     # TODO: Include collate function here
 
-    NUM_WORKERS = 8
+    NUM_WORKERS = 16
```
pbeukema: are we actually constrained by num_workers? 16 feels like very few.

helios/data/dataset.py
yawenzzzz: Fixed.

helios/data/dataset.py
yawenzzzz: We need to load the whole sample first and then apply random subsetting, not sure if chunking will be very useful or not.

helios/train/loss.py
yawenzzzz: Fixed.

scripts/latent_mim.py
yawenzzzz: In DDP setup, the total number of workers is num_workers * num_gpus, so it's 8 fold. Larger number of workers will also require a larger shared memory, we set it to 256GB but still sometimes ran into out of memory when the num_workers is larger, right now, num_workers = 4 and num_gpus = 8 seems quite stable.

helios/data/dataset.py
yawenzzzz: Fixed.

helios/data/dataloader.py
```diff
@@ -407,7 +407,7 @@ class HeliosDataLoaderConfig(Config):
     num_threads: int | None = None
     num_workers: int = 0
     prefetch_factor: int | None = None
-    target_device_type: str = "cpu"
```
Hgherzog, line 410: why changing this default?

helios/data/dataset.py
```diff
@@ -667,23 +672,39 @@ def normalize_image(self, modality: ModalitySpec, image: np.ndarray) -> np.ndarr
...
-                image = convert_to_db(image)
-            # Normalize data and convert to dtype
-            if self.normalize:
-                image = self.normalize_image(modality, image)
-            sample_dict[modality.name] = image.astype(self.dtype)
-            # Get latlon and timestamps from Sentinel2 data
-            if modality == Modality.SENTINEL2_L2A:
-                sample_dict["latlon"] = self.get_latlon(sample).astype(self.dtype)
-                sample_dict["timestamps"] = self._get_timestamps(sample)
+        h5_file_path = self.h5py_dir / f"sample_{index}.h5"
```
Hgherzog: nit: write a helper function for this

helios/data/dataset.py
```diff
@@ -667,23 +672,39 @@ def normalize_image(self, modality: ModalitySpec, image: np.ndarray) -> np.ndarr
...
+                sample_dict[modality.name] = image
+                # Get latlon and timestamps from Sentinel2 data
+                if modality == Modality.SENTINEL2_L2A:
+                    sample_dict["latlon"] = self.get_latlon(sample).astype(self.dtype)
+                    sample_dict["timestamps"] = self._get_timestamps(sample)
+            with h5py.File(h5_file_path, "w") as f:
+                for modality_name, image in sample_dict.items():
+                    f.create_dataset(modality_name, data=image)
+        else:
+            # took about 0.02 seconds to read h5 data
```
Hgherzog: Can you add the script where you got this benchmarking?

helios/train/loss.py
```diff
@@ -94,31 +95,45 @@ def compute(
         pred = F.normalize(pred, p=2, dim=-1)
         target = F.normalize(target, p=2, dim=-1)
 
-        scores = torch.einsum("npd,nqd->npq", pred, target) / self.tau
         count = (all_masks == MaskValue.DECODER.value).sum(dim=-1)
 
         if self.mask_other_samples:
-            logit_mask = torch.full_like(scores, -torch.finfo(scores.dtype).max)
+            # Compute scores per sample to reduce memory usage
+            losses = []
```
Hgherzog, line 103: nit: there is a non for loop implementation which you cna get by skipping the unsqueeze feel free to just add a TODO and I can update implemenation

helios/data/dataset.py
```diff
@@ -325,25 +328,28 @@ def __init__(
...
-        # Note: if samples are provided, use them, if not, get them from the tile directory
+        # If samples are provided, use them, if not, get them from the tile directory
         if not samples:
             samples = self._get_samples()  # type: ignore
         if len(samples) == 0:
             raise ValueError("No samples provided")
         self.samples = self._filter_samples(samples)  # type: ignore
         self.dtype = dtype
         self.normalize = normalize
+        self.h5py_dir = self.tile_path / h5py_folder / str(len(self.samples))
```
Hgherzog: As more modalities are added data per sample may change, ie with SRTM. Add supported Modalities based str to this dir as well

helios/data/dataloader.py
yawenzzzz: We should set it to "cuda" if we're using GPUs, `pin_memory=self.target_device_type == "cuda" and self.num_workers > 0,`, if this is default to "cpu" and we're not passing `target_device_type` via data loader config, then it will make `pin_memory` false. I tried setting `pin_memory` as False and don't think it bring major improvement to the memory usage, so setting it back to True is better since we're training with GPUs. Here's how OLMO-core set this `target_device_type`:
https://github.com/allenai/OLMo-core/blob/56e06ee289e3dfb42cab2b42c6ee21a4b73f5fc5/src/olmo_core/data/data_loader.py#L1068
https://github.com/allenai/OLMo-core/blob/main/src/olmo_core/utils.py#L126

helios/data/dataset.py
yawenzzzz: This number is from my local run, I added that at index = 0, logging the time taken to read h5.

helios/train/loss.py
yawenzzzz: Cool, I added a TODO for this. Thanks!

helios/data/dataset.py
yawenzzzz: Fixed.

helios/data/dataset.py
yawenzzzz: make sense! I added the supported modalities to the folder name, now it's tile_path / h5py_folder / supported_modalities / len(samples), under the same set of supported_modalities, we can apply different filtering rules.

yawenzzzz: I also added it into dataset fingerprint

helios/data/dataloader.py
Hgherzog: why not self.target_device_type or get_default_device().type, as in olmo-core it makes sense to me to use the get_default always so that this will be simple for tests/local dev

yawenzzzz: make sense, I changed to `get_default_device().type`

PR #143, opened 2025-03-14
Restore some defaults

yawenzzzz:


helios/internal/common.py
```diff
@@ -70,7 +70,7 @@ def build_launch_config(
         weka_buckets=weka_buckets,
         beaker_image=f"henryh/{OLMoCoreBeakerImage.stable}",  # we can all use the same image for now
```
Hgherzog, line 71: Can you add an additional script saving the base configurations of your experiments that had good speed so it is easy to reproduce, saving it in a dated folder

yawenzzzz: Sure, I added a script for it: helios/scripts/ddp_experiments/20250314_galileo_base_ddp.py

PR #145, opened 2025-03-18
Add PASTIS evaluation dataset

yawenzzzz:
- [x] Add PASTIS dataset
- [x] Process PASTIS-R (S2 + S1) dataset
- [x] Add PASTIS-R dataset
- [x] Make evaluation epoch configurable per task
- [x] Add test
- [x] Run experiment

NOTE: right now, I processed the PASTIS data similarly as in Presto, i.e., getting the monthly mean of Sentinel2, but this may not be great given the potential cloud coverage, but there's no metadata on cloud coverage, so I don't have a very good solution for it. For comparison purpose, I think using mean is fine.

### DIFF
--- a/helios/evals/datasets/constants.py
+++ b/helios/evals/datasets/constants.py
@@ -23,6 +23,7 @@ EVAL_S1_BAND_NAMES = [
     "vh",
 ]
 
+
 def _eval_s2_band_index_from_helios_name(helios_name: str) -> int:
     for idx, band_name in enumerate(EVAL_S2_BAND_NAMES):
         if helios_name.endswith(band_name.split(" ")[0][-2:]):
--- a/helios/internal/experiment.py
+++ b/helios/internal/experiment.py
@@ -323,7 +323,7 @@ def main(
 
     """
     usage = f"""
-[yellow]Usage:[/] [i blue]python[/] [i cyan]{sys.argv[0]}[/] [i b magenta]{'|'.join(SubCmd)}[/] [i b]RUN_NAME CLUSTER[/] [i][OVERRIDES...][/]
+[yellow]Usage:[/] [i blue]python[/] [i cyan]{sys.argv[0]}[/] [i b magenta]{"|".join(SubCmd)}[/] [i b]RUN_NAME CLUSTER[/] [i][OVERRIDES...][/]
 If running command on a local machine ie from a session, you can use the [b]local[/b] cluster name.
 [b]Subcommands[/]
 [b magenta]launch:[/]     Not Implemented. Launch the script on Beaker with the [b magenta]train[/] subcommand.


helios/evals/datasets/constants.py
```diff
@@ -18,24 +18,30 @@
...
 
+EVAL_S1_BAND_NAMES = [
+    "vv",
+    "vh",
+]
 
-EVAL_S1_BAND_NAMES = ["vv", "vh"]
+def _eval_s2_band_index_from_helios_name(helios_name: str) -> int:
+    for idx, band_name in enumerate(EVAL_S2_BAND_NAMES):
+        if helios_name.endswith(band_name.split(" ")[0][-2:]):
```
favyen2, line 29: Suggest to add comment here to explain why this is necessary perhaps with an example of Helios band vs band name in EVAL_S2_BAND_NAMES (and why can't just change EVAL_S2_BAND_NAMES).

helios/evals/datasets/pastis_dataset.py
```diff
@@ -0,0 +1,433 @@
...
+        """Process the PASTIS-R dataset."""
+        with open(self.data_dir / "metadata.geojson") as f:
+            meta_data = json.load(f)
+
+        all_data: dict[str, dict[str, list[torch.Tensor]]] = {
+            f"fold_{i}": {"s2_images": [], "s1_images": [], "months": [], "targets": []}
+            for i in range(1, 6)
+        }
+
+        doesnt_have_twelve = 0
```
favyen2, line 220: Would be helpful to have a comment here saying that some PASTIS examples don't have Sentinel-2 images from 12 distinct months.

helios/evals/datasets/pastis_dataset.py
```diff
@@ -0,0 +1,433 @@
...
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        """Aggregate images into monthly averages."""
+        months_dict = dict[str, list[torch.Tensor]]()
+        for m in self.all_months:
+            months_dict[m] = []
+
+        for idx, date in dates.items():
+            month = str(date)[:6]
+            img = torch.tensor(images[int(idx)], dtype=torch.float32)
+            if images.shape[1] == 10:
```
favyen2: If it is always expected to have 10 bands here, then suggest verifying that (throw ValueError) and always imputing instead of like this.

If some images have 10 bands but others have 13 bands, then a comment explaining it would be helpful.

helios/evals/datasets/pastis_dataset.py
```diff
@@ -0,0 +1,433 @@
...
+            s2_image = self.normalizer_computed.normalize(
+                Modality.SENTINEL2_L2A, s2_image
+            )
+            if self.is_multimodal:
+                s1_image = self.normalizer_computed.normalize(
+                    Modality.SENTINEL1, s1_image
+                )
+
+        timestamps = []
+        # A little hack to get the correct year for the timestamps
```
favyen2: Why not store both year and month in the months.pt instead of just month? It seems there is some overlap in the months.

Otherwise suggest to all_months to 2018-09 to 2019-08 so there is no overlap and this code here will always work. But it seems that storing the year should be preferred so there is no need to guess when reading it.

helios/evals/datasets/pastis_dataset.py
```diff
@@ -0,0 +1,433 @@
...
+            masked_sample = MaskedHeliosSample.from_heliossample(
+                HeliosSample(
+                    sentinel2_l2a=torch.tensor(s2_image).float(),
+                    timestamps=timestamps,
+                )
+            )
+        return masked_sample, labels.long()
+
+
+# if __name__ == "__main__":
```
favyen2: I think either uncomment or remove.

helios/train/callbacks/evaluator_callback.py
```diff
@@ -77,10 +78,7 @@ def _get_embeddings(
 
     def val(self) -> float:
         """Validate the model on the downstream task."""
-        # try:
```
favyen2, line 80: Remove the commented code added here.

helios/train/callbacks/evaluator_callback.py
```diff
@@ -128,33 +126,29 @@ def val(self) -> float:
             raise ValueError(f"Unrecognized task type: {self.config.task_type}")
         logger.info(f"Downstream evaluator {self.dataset} score: {val_result}")
         return val_result
-        # except Exception as e:
```
favyen2, line 131: Remove the commented code added here.

yawenzzzz: Fixed.

helios/train/callbacks/evaluator_callback.py
yawenzzzz: Fixed.

helios/evals/datasets/pastis_dataset.py
yawenzzzz: Fixed, I removed it.

helios/evals/datasets/constants.py
yawenzzzz: These names are adopted from Geobench datasets, their band stats are saved in json files, for example, /weka/dfive-default/presto-geobench/dataset/geobench/classification_v1.0/m-eurosat/band_stats.json, and the Geobench dataset class will load the json file get the normalization stats. We use the same S2 band names for all eval datasets that require S2 data for consistency. I added a comment on this.

helios/evals/datasets/pastis_dataset.py
yawenzzzz: Sure, added a comment on this, and the results is 0 meaning that all samples have the 12 months' data.

helios/evals/datasets/pastis_dataset.py
yawenzzzz: It always has 10 bands, I added a check there to verify if it's sentinel2 and if the band numbers is equal to 10.

helios/evals/datasets/pastis_dataset.py
yawenzzzz: OH nice catch! Yea, this is unnecessary as I can easily get the year from the year-month str. I updated this part.

PR #147, opened 2025-03-18
Fix PatchDisc loss

yawenzzzz:
1. Change random masking on latlon, ensure number of decoder tokens are the same for samples within a batch
2. Fix the PatchDisc loss when `mask_other_samples = True`

TODO:

### DIFF
--- a/helios/train/masking.py
+++ b/helios/train/masking.py
@@ -212,8 +212,7 @@ class MaskingStrategy:
             decode_tokens = int(num_tokens * self.decode_ratio)
             target_tokens = int(num_tokens - (encode_tokens + decode_tokens))
         else:
-            # If there's only one token, we assign it to online encoder or decoder based on their ratios
-            # In this way, latlon can be handled similarly as other modalities
+            # If there's only one token, we assign it to online encoder or decoder based on ratios
             encoder_only = self.generator.random() < self.encode_ratio
             decoder_only = self.generator.random() < self.decode_ratio
             if encoder_only:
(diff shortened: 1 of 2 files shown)


helios/train/masking.py
```diff
@@ -198,17 +198,35 @@ def _create_random_mask(
...
 
-        encode_tokens = int(num_tokens * self.encode_ratio)
-        decode_tokens = int(num_tokens * self.decode_ratio)
-        target_tokens = int(num_tokens - (encode_tokens + decode_tokens))
+        if num_tokens > 1:
+            encode_tokens = int(num_tokens * self.encode_ratio)
+            decode_tokens = int(num_tokens * self.decode_ratio)
+            target_tokens = int(num_tokens - (encode_tokens + decode_tokens))
+        else:
+            # If there's only one token, we assign it to online encoder or decoder based on ratios
```
gabrieltseng, line 215: When does this happen? it shouldn't

The loss doesn't really make sense if there is only one token

yawenzzzz: When Modality is latlon, since it's number of band_sets is 1 and also no spatial and temporal, then number of tokens become one, unless you pad the batch size dimension to it. I change this latlon part because: 1. for other modalities, we apply the same mask, so maybe we should do it similarly to latlon, and use the encoder and decoder ratio in a similar way, 2. when fixing the PatchDisc loss (we were trying to avoid the large matrix created when mask_other_samples is True, this may have issues when we have a large batch size), I found that the number of Decoder tokens varies only by one, and it's introduced by the latlon token, so I did this fix. But happy to chat on what is a better approach.

PR #150, opened 2025-03-19
Henry/faster worker spawns

Hgherzog:
1. Enabled faster startup times 30 seconds to 6 seconds per worker with 4 workers (gets better with more workers) by moving the samples list out of the dataset class so it doesn't get pickled into every worker.
2. Subsetting in get item. Subsetting after collation means that really big tensors are in the worker queue which uses up a ton of memory and was causing a lot of problems. Subsetting in getitem creates a big speed up for normalization (s2 normalization was actually dominating get-item because we normalized a ton of pixels we are throwing away). It does add a bit more complexity as we have to pass patch size around with a batch but the result is we need way way fewer workers to saturate. 
3. Debug script and additional settings for faster start up of experiments.

### DIFF
--- a/helios/data/dataset.py
+++ b/helios/data/dataset.py
@@ -10,7 +10,7 @@ from collections.abc import Sequence
 from dataclasses import dataclass
 from math import floor
 from pathlib import Path
-from typing import Any, NamedTuple
+from typing import Any, NamedTuple, cast
 
 import h5py
 import numpy as np
@@ -51,12 +51,11 @@ logger = logging.getLogger(__name__)
 class HeliosSample(NamedTuple):
     """A sample of the data from the Helios dataset.
 
-    We always require sentinel2 data.
     This is a namedtuple that contains the data of a single sample or a batch of samples from the Helios dataset.
     For each modality, we have an ArrayTensor named by the modality, along with the latlon and timestamps.
     """
 
-    sentinel2_l2a: ArrayTensor  # [B, H, W, T, len(S2_bands)]
+    sentinel2_l2a: ArrayTensor | None = None  # [B, H, W, T, len(S2_bands)]
     latlon: ArrayTensor | None = None  # [B, 2]
     timestamps: ArrayTensor | None = None  # [B, T, D=3], where D=[day, month, year]
     sentinel1: ArrayTensor | None = None  # [B, H, W, T, len(S1_bands)]
@@ -88,15 +87,13 @@ class HeliosSample(NamedTuple):
                 # timestamps is a special case which is not in Modality
                 raise ValueError("Timestamps are not maskable")
         else:
-            if self.sentinel2_l2a is None:
-                raise ValueError("Sentinel2 L2A is not present in the sample")
             attribute_shape = []
             if Modality.get(attribute).get_tile_resolution() > 0:
                 # Add batch size (if has), height, width
-                attribute_shape += self.sentinel2_l2a.shape[:-2]
+                attribute_shape += [self.height, self.width]
             if Modality.get(attribute).is_multitemporal:
                 # Add number of timesteps
-                attribute_shape += [self.sentinel2_l2a.shape[-2]]
+                attribute_shape += [self.time]
             if not mask:
                 # Add number of bands
                 attribute_shape += [Modality.get(attribute).num_bands]
@@ -159,42 +156,53 @@ class HeliosSample(NamedTuple):
     @property
     def batch_size(self) -> int:
         """Get the batch size of the data."""
-        if len(self.sentinel2_l2a.shape) == 5:
-            return self.sentinel2_l2a.shape[0]
+        vals = [
+            cast(ArrayTensor, x).shape[0]
+            for x in self.as_dict(ignore_nones=True).values()
+        ]
+        if len(set(vals)) == 1:
+            return vals[0]
         else:
             return 1
 
     @property
     def height(self) -> int:
         """Get the height of the data."""
-        if len(self.sentinel2_l2a.shape) == 5:
-            return self.sentinel2_l2a.shape[1]
-        elif len(self.sentinel2_l2a.shape) == 4:
-            return self.sentinel2_l2a.shape[0]
-        else:
-            raise ValueError(f"Invalid shape: {self.sentinel2_l2a.shape}")
+        height_width_time_modalities = ["sentinel2_l2a", "sentinel1", "worldcover"]
+        for modality in height_width_time_modalities:
+            x = getattr(self, modality)
+            if x is not None:
+                if len(x.shape) == 5:
+                    return x.shape[1]
+                else:
+                    # no batch dimension
+                    if len(x.shape) != 4:
+                        raise ValueError(f"Unexpected shape {x.shape} for {modality}")
+                    return x.shape[0]
+        raise ValueError("No modality with height or width present")
 
     @property
     def width(self) -> int:
-        """Get the width of the data."""
-        if len(self.sentinel2_l2a.shape) == 5:
-            return self.sentinel2_l2a.shape[2]
-        elif len(self.sentinel2_l2a.shape) == 4:
-            return self.sentinel2_l2a.shape[1]
-        else:
-            raise ValueError(f"Invalid shape: {self.sentinel2_l2a.shape}")
+        """Get the height of the data."""
+        height_width_time_modalities = ["sentinel2_l2a", "sentinel1", "worldcover"]
+        for modality in height_width_time_modalities:
+            x = getattr(self, modality)
+            if x is not None:
+                if len(x.shape) == 5:
+                    return x.shape[2]
+                else:
+                    # no batch dimension
+                    if len(x.shape) != 4:
+                        raise ValueError(f"Unexpected shape {x.shape} for {modality}")
+                    return x.shape[1]
+        raise ValueError("No modality with height or width present")
 
     @property
     def time(self) -> int:
         """Get the number of time steps in the data."""
         if self.timestamps is None:
             raise ValueError("Timestamps are not present in the sample")
-        if self.timestamps.ndim == 3:
-            return self.timestamps.shape[1]
-        elif self.timestamps.ndim == 2:
-            return self.timestamps.shape[0]
-        else:
-            raise ValueError(f"Invalid timestamps shape: {self.timestamps.shape}")
+        return self.timestamps.shape[-2]
 
     def _get_max_t_within_token_budget(
         self, h_w_p: int, max_tokens_per_instance: int
--- a/scripts/base_debug_scripts/galileo.py
+++ b/scripts/base_debug_scripts/galileo.py
@@ -224,6 +224,14 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
             norm_stats_from_pretrained=False,
             probe_lr=0.1,
         ),
+        DownstreamTaskConfig(
+            dataset="sen1floods11",
+            batch_size=128,
+            num_workers=8,
+            pooling_type=PoolingType.MEAN,
+            norm_stats_from_pretrained=True,
+            probe_lr=0.1,
+        ),
     ]
     # Let us not use garbage collector fallback
     trainer_config = (
--- a/.github/workflows/labeler.yaml
+++ b/.github/workflows/labeler.yaml
@@ -1,0 +1,32 @@
+name: labeler
+
+on: [pull_request]
+
+jobs:
+  labeler:
+    permissions:
+      pull-requests: write
+      contents: read
+      issues: write
+    runs-on: ubuntu-latest
+    name: Label the PR size
+    steps:
+      - uses: codelytv/pr-size-labeler@v1
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          xs_label: 'size/xs'
+          xs_max_size: '10'
+          s_label: 'size/s'
+          s_max_size: '100'
+          m_label: 'size/m'
+          m_max_size: '500'
+          l_label: 'size/l'
+          l_max_size: '1000'
+          xl_label: 'size/xl'
+          fail_if_xl: 'false'
+          message_if_xl: >
+            This PR exceeds the recommended size of 1000 lines.
+            Please make sure you are NOT addressing multiple issues with one PR.
+            Note this PR might be rejected due to its size.
+          github_api_url: 'https://api.github.com'
+          files_to_ignore: ''
--- a/helios/evals/datasets/__init__.py
+++ b/helios/evals/datasets/__init__.py
@@ -5,6 +5,7 @@ import logging
 from torch.utils.data import Dataset
 
 from .configs import ALL_DATASETS
+from .floods_dataset import FLOODS_DIR, Sen1Floods11Dataset
 from .geobench_dataset import GEOBENCH_DIR, GeobenchDataset
 from .mados_dataset import MADOS_DIR, MADOSDataset
 
@@ -41,3 +42,12 @@ def get_eval_dataset(
             partition=partition,
             norm_stats_from_pretrained=norm_stats_from_pretrained,
         )
+    elif eval_dataset == "sen1floods11":
+        return Sen1Floods11Dataset(
+            path_to_splits=FLOODS_DIR,
+            split=split,
+            partition=partition,
+            norm_stats_from_pretrained=norm_stats_from_pretrained,
+        )
+    else:
+        raise ValueError(f"Unrecognized eval_dataset {eval_dataset}")
--- a/helios/evals/datasets/configs.py
+++ b/helios/evals/datasets/configs.py
@@ -47,6 +47,13 @@ DATASET_TO_CONFIG = {
         is_multilabel=False,
         height_width=80,
     ),
+    "sen1floods11": EvalDatasetConfig(
+        task_type=TaskType.SEGMENTATION,
+        imputes=[],
+        num_classes=2,
+        is_multilabel=False,
+        height_width=64,
+    ),
 }
 
 ALL_DATASETS = list(DATASET_TO_CONFIG.keys())
--- a/helios/evals/datasets/constants.py
+++ b/helios/evals/datasets/constants.py
@@ -19,8 +19,11 @@ EVAL_S2_BAND_NAMES = [
 ]
 
 
-def _eval_band_index_from_helios_name(helios_name: str) -> int:
-    for idx, band_name in enumerate(EVAL_S2_BAND_NAMES):
+EVAL_S1_BAND_NAMES = ["vv", "vh"]
+
+
+def _eval_band_index_from_helios_name(eval_bands: list[str], helios_name: str) -> int:
+    for idx, band_name in enumerate(eval_bands):
         if helios_name.endswith(band_name.split(" ")[0][-2:]):
             return idx
     raise ValueError(f"Unmatched band name {helios_name}")
@@ -28,5 +31,11 @@ def _eval_band_index_from_helios_name(helios_name: str) -> int:
 
 # For now, with Sentinel2 L2A, we will drop the B10 band in the eval datasets
 EVAL_TO_HELIOS_S2_BANDS = [
-    _eval_band_index_from_helios_name(b) for b in Modality.SENTINEL2_L2A.band_order
+    _eval_band_index_from_helios_name(EVAL_S2_BAND_NAMES, b)
+    for b in Modality.SENTINEL2_L2A.band_order
+]
+
+EVAL_TO_HELIOS_S1_BANDS = [
+    _eval_band_index_from_helios_name(EVAL_S1_BAND_NAMES, b)
+    for b in Modality.SENTINEL1.band_order
 ]
--- a/helios/evals/datasets/floods_dataset.py
+++ b/helios/evals/datasets/floods_dataset.py
@@ -1,0 +1,287 @@
+"""Floods eval dataset, based on Sen1Floods11."""
+
+import json
+from pathlib import Path
+
+import numpy as np
+import pandas as pd
+import rioxarray
+import torch
+from einops import rearrange, repeat
+from torch.utils.data import Dataset
+from tqdm import tqdm
+from upath import UPath
+
+from helios.data.constants import Modality
+from helios.data.dataset import HeliosSample
+from helios.train.masking import MaskedHeliosSample
+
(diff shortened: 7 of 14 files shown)


helios/data/dataloader.py
```diff
@@ -66,6 +68,10 @@ def __init__(
         )
         self.dataset = dataset
         assert isinstance(self.dataset, HeliosDataset)  # type: ignore
+        if not self.dataset.is_dataset_prepared:
+            raise RuntimeError("Dataset must be prepared before creating a dataloader")
+        self.patch_sizes = np.arange(1, 8)
```
Hgherzog: These need to be configured and passed in instead of hadcoded

helios/data/dataloader.py
```diff
@@ -76,6 +82,10 @@ def __init__(
         self.drop_last = drop_last
         self._global_indices: np.ndarray | None = None
         self.persistent_workers = persistent_workers
+        self.multiprocessing_context = multiprocessing_context
+        if self.num_workers > 0 and self.multiprocessing_context == "forkserver":
+            # Overhead of loading modules on import by preloading them
+            mp.set_forkserver_preload(["torch", "rasterio"])
```
Hgherzog, line 95: only torch though this doesn't really matter anymore so we could delete

helios/data/dataloader.py
```diff
@@ -279,7 +290,8 @@ def get_mock_batch(self) -> HeliosSample:
         timestamps = torch.cat([days, months, years], dim=1)
         timestamps = rearrange(timestamps, "b t c -> b c t")
         output_dict["timestamps"] = timestamps
-        return HeliosSample(**output_dict)
+        #  TODO: Temporary hardcoded fix later
```
Hgherzog: Make sure mock batch data actually works

helios/data/dataloader.py
```diff
@@ -331,6 +343,22 @@ def iter_batched(
         yield tuple(batch)
 
 
+def _get_batch_item_params_iterator(indices: np.ndarray, patch_size: list[int], sampled_hw_p: list[int], rank_batch_size: int) -> Iterator[tuple[int, int, int]]:
```
Hgherzog: Add a unit test for this

helios/data/dataloader.py
```diff
@@ -361,9 +389,11 @@ def __iter__(self) -> Iterator[HeliosSample]:
             # try to guess a good number of threads.
             num_threads = 4
 
+
         # Potentially slice by threads.
         instance_iterator: Iterator[HeliosSample]
         if num_threads:
+            raise NotImplementedError("Threading is not supported yet")
```
Hgherzog: Write an issue for this to be reimplemented only when needed

helios/data/dataloader.py
```diff
@@ -382,8 +412,14 @@ def __iter__(self) -> Iterator[HeliosSample]:
             instance_iterator = roundrobin(*thread_generators)
         else:
             indices = self.data_loader._get_local_instance_indices(global_indices)
+            # I want to create a generator that yields a tuple of (idx, patch_size, sampled_hw_p)
```
Hgherzog: Remove comments

helios/data/dataset.py
```diff
@@ -167,19 +167,34 @@ def batch_size(self) -> int:
     @property
     def height(self) -> int:
         """Get the height of the data."""
-        return self.sentinel2_l2a.shape[1]
```
Hgherzog: Add a comment about why these implementations need to deal with batch and non batched helios samples

helios/data/dataset.py
```diff
@@ -220,7 +235,7 @@ def _get_max_t_within_token_budget(
         return min(floor(max_t_within_budget), self.time)
 
     def subset(
```
Hgherzog, line 245: Should this still be a helios sample method it creates that extra wrap and unwrap?

helios/data/dataset.py
```diff
@@ -239,14 +254,14 @@ def subset(
...
         """
-        max_height_width = max(self.height, self.width)
-        max_height_width_tokens = int(max_height_width / patch_size)
-        hw_to_sample = [x for x in hw_to_sample if x <= max_height_width_tokens]
-        if len(hw_to_sample) == 0:
-            raise ValueError(
-                "max height/width allowed by sample smaller than values in hw_to_sample"
-            )
-        sampled_hw_p = random.choice(hw_to_sample)
+        # max_height_width = max(self.height, self.width)
```
Hgherzog: Remove this

helios/data/dataset.py
```diff
@@ -239,14 +254,14 @@ def subset(
         of timesteps allowable so that the total tokens (per instance) is >=
```
Hgherzog, line 262: Add documentation that this no longer works on batched samples but individual samples instead

helios/data/dataset.py
```diff
@@ -258,48 +273,49 @@ def subset(
...
 
     # TODO: Gets all non-None modalities ASSUMES ALL SAMPLES HAVE THE SAME MODALITIES
-    sample_fields = batch[0].modalities
+    patch_size, batch_zero = batch[0]
+    sample_fields = batch_zero.modalities
 
     # Create a dictionary of stacked tensors for each field
     collated_dict = {field: stack_or_none(field) for field in sample_fields}
-    return HeliosSample(**collated_dict)
+    return patch_size, HeliosSample(**collated_dict)
```
Hgherzog, line 317: Open to other ideas as this is quite annoying I would rather return a single named tuple but don't want patch size to be confused with HeliosSample

helios/data/dataset.py
```diff
@@ -312,9 +328,10 @@ def __init__(
         tile_path: UPath,
         supported_modalities: list[ModalitySpec],
         dtype: DType,
-        samples: list[SampleInformation] | None = None,
         normalize: bool = True,
+        token_budget: int = 1500,
         h5py_folder: str = "h5py_data",
+        attempt_to_create_h5_files: bool = False,
```
Hgherzog: There should be something added to this where you can pass the path to an h5 dataset directly and that lets you skip over all the loading steps

helios/data/dataset.py
```diff
@@ -326,41 +343,34 @@ def __init__(
...
-            / h5py_folder
-            / "_".join(
-                sorted([modality.name for modality in self.supported_modalities])
-            )
-            / str(len(self.samples))
-        )
-        os.makedirs(self.h5py_dir, exist_ok=True)
-
         if self.normalize:
+            # We want to delay the creation of the normalizers
```
Hgherzog: Remove this comment it doesnt matter

helios/data/dataset.py
```diff
@@ -408,9 +420,103 @@ def work_dir_set(self) -> bool:
...
+        os.makedirs(self._h5py_dir, exist_ok=True)
+
+    def prepare(self, samples: list[SampleInformation] | None = None) -> None:
+        """Prepare the dataset.
+
+        THIS SHOULD BE CALLED BY THE MAIN PROCESS ONLY and should happen
+        before any other process tries to use the dataset
+        """
+        logger.info("Preparing dataset...")
+        # if samples is None:
```
Hgherzog: Uncomment this and deal with the logic needed to set this up correctly

helios/data/dataset.py
```diff
@@ -706,34 +816,37 @@ def _create_h5_file(
...
-                    f"Time taken to read h5 data: {end_time - start_time} seconds"
-                )
+            raise FileNotFoundError(
+                f"H5 file {h5_file_path} does not exist, Be Sure to run prepare before starting Training"
+            )
+
+        with h5py.File(h5_file_path, "r") as f:
+            sample_dict = {k: v[()] for k, v in f.items()}
+
+        sample = HeliosSample(**sample_dict)
```
Hgherzog, line 881: don't love this pack and unpack

helios/data/dataset.py
```diff
@@ -312,9 +328,10 @@ def __init__(
         tile_path: UPath,
         supported_modalities: list[ModalitySpec],
         dtype: DType,
-        samples: list[SampleInformation] | None = None,
         normalize: bool = True,
+        token_budget: int = 1500,
```
Hgherzog: Add these to the config and remove token budget from model

helios/data/dataset.py
```diff
@@ -706,34 +816,37 @@ def _create_h5_file(
...
-            if index == 0:
-                end_time = time.time()
-                logger.info(
-                    f"Time taken to read h5 data: {end_time - start_time} seconds"
-                )
+            raise FileNotFoundError(
+                f"H5 file {h5_file_path} does not exist, Be Sure to run prepare before starting Training"
+            )
+
+        with h5py.File(h5_file_path, "r") as f:
```
Hgherzog, line 878: make a comment about what () is doing so we can use it for future optimization if needed

helios/data/normalize.py
Hgherzog, line 1: Revert this entire file

helios/internal/experiment.py
```diff
@@ -26,7 +26,7 @@
 from helios.nn.latent_mim import LatentMIMConfig
 from helios.train.train_module.latent_mim import LatentMIMTrainModuleConfig
 from helios.train.train_module.train_module import HeliosTrainModuleConfig
-
+from torch.utils.viz._cycles import warn_tensor_cycles
```
Hgherzog: remove the tensor cycles stuff

helios/internal/experiment.py
```diff
@@ -199,7 +199,34 @@ def launch(config: HeliosExperimentConfig) -> None:
     logger.info("Launching the experiment")
     logger.info(config)
     # Set follow=False if you don't want to stream the logs to the terminal
-    config.launch.launch(follow=True)
+    config.launch.launch(follow=False)
+
+
+def prep(config: HeliosExperimentConfig) -> None:
```
Hgherzog, line 206: Double check these commands work

helios/train/train_module/galileo.py
```diff
@@ -16,7 +16,8 @@
 from olmo_core.train.train_module.transformer import (
     TransformerActivationCheckpointingConfig,
 )
-
+import sys
```
Hgherzog: Move all this memory info stuff to a utils file or something to save it

helios/train/train_module/galileo.py
```diff
@@ -230,17 +296,40 @@ def train_batch(self, batch: HeliosSample, dry_run: bool = False) -> None:
 
         NOTE: For contrastive losses, the loss is invariant to the global batch size across GPUS as well
         """
+
+        # # # Collect all tensors
+        # tensors = [obj for obj in gc.get_objects() if isinstance(obj, torch.Tensor)]
+
+        # # Sort tensors by size (bytes)
```
Hgherzog: delete

helios/train/train_module/galileo.py
```diff
@@ -230,17 +296,40 @@ def train_batch(self, batch: HeliosSample, dry_run: bool = False) -> None:
...
+        # #     if i < 5:
+        # #         print(f"Tensor {i}: Shape {tensor.shape}, Dtype {tensor.dtype}, Size {size_in_mb:.2f} MB")
+        # #     total_size += size_in_mb
+        # # logger.info(f"Total size of top 100 tensors: {total_size:.2f} MB")
+        # logger.info("Logging total virtual memory usage")
+        # logger.info(f"Total virtual memory usage: {psutil.virtual_memory().total / (1024 * 1024 * 1024):.2f} GB")
+        # total_pss = log_total_memory_usage()
+
+        # self.trainer.record_metric("worker_memory/total_pss", total_pss, ReduceType.mean)
+        if dry_run:
```
Hgherzog: move dry run back to before the step

helios/train/train_module/galileo.py
```diff
@@ -309,8 +390,8 @@ def train_batch(self, batch: HeliosSample, dry_run: bool = False) -> None:
         if dry_run:
             return
 
-        del batch  # In case this helps with memory utilization.
```
Hgherzog, line 312: re add these stuff

scripts/single_gpu_data_loading_speed_sweep/20250317_galileo_data_speed_sweep.py
Hgherzog, line 1: remove these from the PR was a dead end

helios/data/dataloader.py
```diff
@@ -361,9 +389,11 @@ def __iter__(self) -> Iterator[HeliosSample]:
             # try to guess a good number of threads.
             num_threads = 4
 
+
         # Potentially slice by threads.
```
pbeukema: Maybe I am missing something big here, but I thought we had abundance of CPUs (for multi-processing), and would generally try to avoid multi-threading logic unless we have a specific configuration that we are sure will work.

Hgherzog: I got this implementation from olmo-core and it did work and has slightly different memory tradeoffs then multiprocessing because you don't need to re pickle the dataset to each thread. Hopefully this is obsolete now.

helios/train/train_module/galileo.py
```diff
@@ -251,22 +340,14 @@ def train_batch(self, batch: HeliosSample, dry_run: bool = False) -> None:
                 # Gallileo does this subsetting at the microbatch level so we follow that for now
                 # Smallest h /w must be bigger than the smallest patch size
 
-                patch_size = np.random.choice(
-                    np.arange(
-                        self.model.encoder.min_patch_size,
```
Hgherzog, line 256: We never actually use patch size 8 here

helios/data/dataloader.py
```diff
@@ -331,6 +348,27 @@ def iter_batched(
...
+    patch_size_array = np.array(patch_size)
+    sampled_hw_p_array = np.array(sampled_hw_p)
+    instances_processed = 0
+    # TODO: we should use a generator here
+    for idx in indices:
+        if instances_processed % rank_batch_size == 0:
+            patch_size = np.random.choice(patch_size_array)
+            sampled_hw_p = np.random.choice(sampled_hw_p_array)
+        yield idx, int(patch_size), int(sampled_hw_p)
+        instances_processed += 1
```
yawenzzzz: One potential issue is that `sampled_hw_p` should has an upper bound set by `patch_size`, the max value is `max(height, width) // patch_size`.

helios/data/dataloader.py
```diff
@@ -262,24 +290,29 @@ def get_mock_batch(self) -> HeliosSample:
...
         if Modality.LATLON in self.dataset.supported_modalities:
-            mock_latlon = torch.rand(1, 2)
+            mock_latlon = np.random.rand(2).astype(np.float32)
             output_dict["latlon"] = mock_latlon
-        days = torch.randint(0, 25, (1, 1, 12), dtype=torch.long)
-        months = torch.randint(0, 12, (1, 1, 12), dtype=torch.long)
-        years = torch.randint(2018, 2020, (1, 1, 12), dtype=torch.long)
-        timestamps = torch.cat([days, months, years], dim=1)
-        timestamps = rearrange(timestamps, "b t c -> b c t")
+        days = np.random.randint(0, 25, (12, 1))
```
pbeukema: for reproducibility it might be useful to set an optional seed here and for all the np.random commands.

helios/data/dataset.py
```diff
@@ -258,109 +265,139 @@ def subset(
         for attribute, modality in self.as_dict(ignore_nones=True).items():
             assert modality is not None
             if attribute == "timestamps":
-                new_data_dict[attribute] = modality[:, start_t : start_t + max_t]
+                new_data_dict[attribute] = modality[start_t : start_t + max_t]
                 continue
-
+            # remember to add the batching back in
```
pbeukema: remove?

helios/data/dataset.py
```diff
@@ -408,9 +447,106 @@ def work_dir_set(self) -> bool:
...
+            )
+            / str(num_samples)
+        )
+        logger.info(f"Setting h5py_dir to {self.h5py_dir}")
+        os.makedirs(self.h5py_dir, exist_ok=True)
+
+    def prepare(self, samples: list[SampleInformation] | None = None) -> None:
+        """Prepare the dataset.
+
+        THIS SHOULD BE CALLED BY THE MAIN PROCESS ONLY and should happen
```
pbeukema, line 513: are you saying there is race condition? or is the conditional "is_dataset_prepared" handling it already?

helios/data/dataloader.py
```diff
@@ -66,16 +71,28 @@ def __init__(
         )
         self.dataset = dataset
         assert isinstance(self.dataset, HeliosDataset)  # type: ignore
+        if not self.dataset.is_dataset_prepared:
```
pbeukema, line 74: this is a ominous/unclear. If all this is doing is checking if the indices are none, then I might just say that instead.

helios/data/dataset.py
Hgherzog: Essentially yes. The reason is if you run prepare on every gpu it will try to launch a ton of grandchild processes in all the child processes to create the same h5 files. So potentially many different processes could be trying to open close and write the same files.

helios/data/dataset.py
```diff
@@ -220,9 +235,9 @@ def _get_max_t_within_token_budget(
         return min(floor(max_t_within_budget), self.time)
 
     def subset(
-        self, patch_size: int, max_tokens_per_instance: int, hw_to_sample: list[int]
+        self, patch_size: int, max_tokens_per_instance: int, sampled_hw_p: int
     ) -> "HeliosSample":
```
yawenzzzz, line 247: Update the docstring of this function since now we're passing `sampled_hw_p` instead of `sampled_hw_p`

helios/data/dataset.py
```diff
@@ -258,109 +265,139 @@ def subset(
...
+class GetItemArgs(NamedTuple):
+    """Arguments for the __getitem__ method of the HeliosDataset."""
+
+    idx: int
+    patch_size: int
+    sampled_hw_p: int
+    token_budget: int | None = None
 
 
 class HeliosDataset(Dataset):
```
yawenzzzz, line 329: The dataset class is getting bigger, we don't need to resolve in this PR, but one TODO is to remove some non dataset-related functions to `utils.py` or `dataset_utils.py`, such as `_filter_samples`, `compute_normalization_values` etc.

helios/data/visualize.py
```diff
@@ -57,16 +57,16 @@ def visualize_sample(
         wc_classes_sorted[-1] + 1
     ]  # e.g., up to 101 if last is 100
     wc_norm = mcolors.BoundaryNorm(wc_bounds, wc_cmap.N)
-
-    sample = dataset.samples[sample_index]
+    samples = dataset._get_samples()
+    samples = dataset._filter_samples(samples)
```
yawenzzzz, line 61: Given that we are using `get_samples` and `filter_samples` outside the `dataset` class, we would want to remove the prefix `_` as it is for functions that are private to that class.

helios/data/dataloader.py
```diff
@@ -43,18 +44,22 @@ def __init__(
         dataset: HeliosDataset,
         work_dir: UPath,
         global_batch_size: int,
+        min_patch_size: int,
+        max_patch_size: int,
+        sampled_hw_p_list: list[int],
+        token_budget: int = BASE_TOKEN_BUDGET,
```
yawenzzzz: To support passing `None`, we need to change this to: `token_budget: int | None = BASE_TOKEN_BUDGET`. I'm wondering whether or not we should set the default to None instead of `BASE_TOKEN_BUDGET`, as our scripts always pass the token budget.

helios/data/dataloader.py
```diff
@@ -262,24 +290,29 @@ def get_mock_batch(self) -> HeliosSample:
...
+        years = np.random.randint(2018, 2020, (12, 1))
+        timestamps = np.concatenate([days, months, years], axis=1)  # shape: (12, 3)
+
         output_dict["timestamps"] = timestamps
-        return HeliosSample(**output_dict)
+
+        patch_size = 1
+        collated_sample = self.collator(
+            [(patch_size, HeliosSample(**output_dict).subset(patch_size, 1500, 6))]
+        )
```
yawenzzzz: Suggestion: add `max_tokens_per_instance = 1500, sampled_hw_p = 6` to be more clear

helios/data/dataloader.py
```diff
@@ -331,6 +365,33 @@ def iter_batched(
...
+    instances_processed = 0
+    # TODO: We need to maintain state and reproducibility here
+    # DO we want this to differ by rank?
+    for idx in indices:
+        if instances_processed % rank_batch_size == 0:
+            patch_size = np.random.choice(patch_size_array)
+            max_height_width_tokens = int(IMAGE_TILE_SIZE / patch_size)
+            filtered_hw_p_to_sample_array = hw_p_to_sample_array[
+                hw_p_to_sample_array <= max_height_width_tokens
+            ]
```
yawenzzzz, line 396: Suggestion: it will be safer to add `hw_p_to_sample_array > 0` as well as we don't want to get `hw_p_sample = 0`

helios/data/dataloader.py
```diff
@@ -402,9 +442,12 @@ class HeliosDataLoaderConfig(Config):
 
     work_dir: str
     global_batch_size: int
+    min_patch_size: int
+    max_patch_size: int
+    sampled_hw_p_list: list[int]
     seed: int
+    token_budget: int = BASE_TOKEN_BUDGET
```
yawenzzzz: Change this to `token_budget: int | None = BASE_TOKEN_BUDGET` so that None can pass through

scripts/base_debug_scripts/galileo.py
```diff
@@ -35,15 +42,12 @@
 
 logger = logging.getLogger(__name__)
 
+MAX_PATCH_SIZE = 8  # NOTE: actual patch_size <= max_patch_size
+MIN_PATCH_SIZE = 1
+
```
yawenzzzz, line 47: I'm fine with these, but an alternative option is to move `min_patch_size` to `build_data_loader_config()` as it's only used by the dataloader, and move `max_patch_size` to `build_model_config()` and then pass its `max_patch_size` to dataloader in the `train` function of `experiment.py`.

helios/data/dataset.py
Hgherzog: agreed issue created

helios/data/visualize.py
Hgherzog: Adding this to that issue

scripts/base_debug_scripts/galileo.py
Hgherzog: I think I will leave this for here for now. I check that these align in on_attach in the helios train module

PR #151, opened 2025-03-19
Floods

gabrieltseng:


helios/data/dataset.py
```diff
@@ -159,27 +156,53 @@ def maybe_move_to_device(tensor: torch.Tensor | None) -> torch.Tensor | None:
...
+            x = getattr(self, modality)
+            if x is not None:
+                if len(x.shape) == 5:
+                    return x.shape[2]
+                else:
+                    # no batch dimension
+                    if len(x.shape) != 4:
+                        raise ValueError(f"Unexpected shape {x.shape} for {modality}")
+                    return x.shape[1]
+        raise ValueError("No modality with height or width present")
```
yawenzzzz, line 198: This looks fine, but since the `shape` method is for a specific modality, why don't we just use instance method and pass `attribute` so that we get the height and width of that specific modality?

helios/evals/datasets/floods_dataset.py
```diff
@@ -0,0 +1,287 @@
...
+FLOODS_DIR = UPath("/weka/dfive-default/presto_eval_sets/floods")
+
+
+class Sen1Floods11Processor:
+    """Class for preprocessing floods dataset."""
+
+    input_hw = 512
+    output_tile_size = 64
+
+    s1_bands = ("VV", "VH")
```
yawenzzzz, line 37: Have the S1 bands in Sen1Floods11 been terrain corrected?

helios/evals/datasets/floods_dataset.py
```diff
@@ -0,0 +1,287 @@
...
+        # target is shape (N, H, W)
+
+        new_s1, new_target = [], []
+        for i in range(s1.shape[0]):
+            if torch.any(torch.isnan(s1[i])) or torch.any(torch.isinf(s1[i])):
+                continue
+            new_s1.append(s1[i])
+            new_target.append(target[i])
+
+        return torch.stack(new_s1), torch.stack(new_target)
```
yawenzzzz, line 246: Oh I don't have this for PASTIS-R, is nan a common issue for S1 data?

helios/evals/datasets/floods_dataset.py
gabrieltseng: hmm I am not sure - the [paper](https://openaccess.thecvf.com/content_CVPRW_2020/papers/w11/Bonafilia_Sen1Floods11_A_Georeferenced_Dataset_to_Train_and_Test_Deep_Learning_CVPRW_2020_paper.pdf) doesn't specify

helios/data/dataset.py
gabrieltseng: This handles cases where we want a shape for a modality we don't have (e.g. to make a mask).

helios/evals/datasets/floods_dataset.py
gabrieltseng: No I think its just for the floods data

PR #155, opened 2025-03-21
Mae

pjreddie:
Adding in MAE training module and script and relevant classes to enable it

### DIFF
--- a/scripts/base_debug_scripts/mae.py
+++ b/scripts/base_debug_scripts/mae.py
@@ -202,6 +202,7 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
             num_workers=8,
             pooling_type=PoolingType.MEAN,
             norm_stats_from_pretrained=True,
+            eval_duration=Duration.epochs(EVAL_INTERVAL_EPOCHS),
         ),
         DownstreamTaskConfig(
             dataset="mados",
@@ -210,6 +211,7 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
             pooling_type=PoolingType.MEAN,
             norm_stats_from_pretrained=False,
             probe_lr=0.1,
+            eval_duration=Duration.epochs(EVAL_INTERVAL_EPOCHS),
         ),
     ]
     trainer_config = (
@@ -230,7 +232,6 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
             "downstream_evaluator",
             DownstreamEvaluatorCallbackConfig(
                 tasks=EVAL_TASKS,
-                eval_duration=Duration.epochs(EVAL_INTERVAL_EPOCHS),
             ),
         )
         .with_callback("garbage_collector", garbage_collector_callback)


helios/nn/mae.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class MAE(nn.Module, DistributedMixins):
+    """Masked Auto-Encoder Module."""
+
+    def __init__(
+        self,
+        encoder: nn.Module,
+        decoder: nn.Module,
+        reconstructor: nn.Module,
```
gabrieltseng, line 26: nit: This is not exactly analogous to (e.g.) SatMAE since there is now an additional linear layer in the decoder before the reconstructor.
I am not sure this matters but just making a note.

pjreddie: noted, complexity-wise, an extra linear layer with no norm doesn't change theoretical model complexity (it could be merged into the next weighted layer), but it may change training dynamics (probably not too much)

scripts/base_debug_scripts/galileo.py
pjreddie, line 1: this file shouldn't actually be changed, I'll roll this back, not sure where this came from

helios/train/loss.py
```diff
@@ -320,6 +322,62 @@ def compute(
...
+        for modality in predictions.modalities:
+            if getattr(predictions, modality) is not None:
+                masked_name = targets.get_masked_modality_name(modality)
+                valid_dict[modality] = getattr(targets, modality)
+                valid_dict[masked_name] = getattr(targets, masked_name)
+        valid_targets = TokensAndMasks(**valid_dict)
+        labels, label_masks = self._flatten_helios_data(valid_targets)
+        data = data * (masks == MaskValue.DECODER.value)
+        labels = labels * (masks == MaskValue.DECODER.value)
+        return F.mse_loss(data, labels)
```
gabrieltseng: Fix averaging

helios/train/loss.py
```diff
@@ -320,6 +322,62 @@ def compute(
...
+        """
+        data, masks = self._flatten_helios_data(predictions)
+        valid_dict = {}
+        for modality in predictions.modalities:
+            if getattr(predictions, modality) is not None:
+                masked_name = targets.get_masked_modality_name(modality)
+                valid_dict[modality] = getattr(targets, modality)
+                valid_dict[masked_name] = getattr(targets, masked_name)
+        valid_targets = TokensAndMasks(**valid_dict)
+        labels, label_masks = self._flatten_helios_data(valid_targets)
```
gabrieltseng, line 375: Use label masks (& / or make sure encoder masks are unchanged)

helios/train/callbacks/speed_monitor.py
```diff
@@ -26,6 +27,14 @@ def pre_train(self) -> None:
         train_module = self.trainer.train_module
 
         self._token_budget = self.trainer.data_loader.token_budget
+        if isinstance(train_module, MAETrainModule):
```
gabrieltseng: This can be in line 38 isinstance(train_module, (MAETrainModule, Lat...))

helios/train/loss.py
gabrieltseng: Add to encoder integration test: https://github.com/allenai/helios/issues/161

pjreddie: added this test for Mae module, it's harder for encoder since you'd have to reconstruct the original mask (as we do in MAE). but the encoder in MAE doesn't modify the mask according to the tests

PR #156, opened 2025-03-21
20250321 LatentMIM Runs

yawenzzzz:
Run link: https://wandb.ai/eai-ai2/2025_03_21_latentmim_tiny_sweep

- Fix broken modality masking
- allow us to set the eval interval for different tasks to different value. Also, switches the evals from a list to a dict to make overrides more straightforward
- de couples  all disc and a lower memory patch_discrimination_new loss

### DIFF
--- a/helios/train/callbacks/evaluator_callback.py
+++ b/helios/train/callbacks/evaluator_callback.py
@@ -32,7 +32,6 @@ class DownstreamEvaluator:
         batch_size: int = 128,
         num_workers: int = 8,
         patch_size: int = 4,
-        eval_duration: Duration = field(default_factory=lambda: Duration.epochs(1)),
         pooling_type: PoolingType = PoolingType.MEAN,
         norm_stats_from_pretrained: bool = True,
         device: torch.device | None = None,
@@ -50,7 +49,6 @@ class DownstreamEvaluator:
         self.norm_stats_from_pretrained = norm_stats_from_pretrained
         self.probe_lr = probe_lr
         self.patch_size = patch_size
-        self.eval_duration = eval_duration
 
     def _get_data_loader(self, split: str) -> DataLoader:
         """Get the data loader for the given split."""
@@ -144,7 +142,6 @@ class DownstreamEvaluatorCallback(Callback):
             )
             if self.step <= 1 or self.step % eval_interval_steps != 0:
                 continue
-            evaluator.eval_duration
             logger.info(f"Running {evaluator.dataset} evaluations...")
             start_time = time.monotonic()
             val_result = evaluator.val()


helios/internal/experiment.py
```diff
@@ -200,7 +200,7 @@ def launch(config: HeliosExperimentConfig) -> None:
     logger.info("Launching the experiment")
     logger.info(config)
     # Set follow=False if you don't want to stream the logs to the terminal
-    config.launch.launch(follow=True)
+    config.launch.launch(follow=False)
```
favyen2, line 203: Should this be configurable or always False is good?

helios/train/callbacks/evaluator_callback.py
```diff
@@ -173,24 +173,24 @@ class DownstreamTaskConfig:
     # ViT-base = 0.01
     probe_lr: float | None = None
     patch_size: int = 4
+    eval_interval: Duration = field(default_factory=lambda: Duration.epochs(1))
```
favyen2: Just wondering, is there an advantage to using `field(default_factory=lambda: X)` instead of `X` for immutable types? (Anyway `Duration` may not be immutable here, I'm not sure.)

helios/train/masking.py
```diff
@@ -507,7 +507,7 @@ def apply_mask(
             MaskedHeliosSample containing the masked data and mask
         """
         if patch_size is not None:
-            raise ValueError("patch_size must not be provided for modality masking")
+            patch_size = None
```
favyen2: As far as I can tell, `patch_size` is not used in this function so it seems to make sense to either keep the ValueError or to remove the `if patch_size is not None:` block.

scripts/base_debug_scripts/galileo.py
```diff
@@ -207,32 +207,31 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
     garbage_collector_callback = GarbageCollectorCallback(gc_interval=1)
     logger.warning("WANDB Distribution Uploads are disabled for Debugging")
     EVAL_INTERVAL_EPOCHS = 1
```
favyen2: I thought the eval interval is moved to DownstreamTaskConfig. Does it still work to pass it to DownstreamEvaluatorCallbackConfig?

favyen2: This seems to be fixed in the PASTIS PR, seems there are some mixed changes between the two PRs.

helios/internal/experiment.py
Hgherzog: I feel like I don't get a ton of value from streaming the logs as it takes time to actually start the experiment anyway as you are often queued

scripts/base_debug_scripts/galileo.py
Hgherzog: yes this is out of date

helios/train/callbacks/evaluator_callback.py
Hgherzog: direct access should be preferred for immutable types. Duration is immutable

helios/train/callbacks/evaluator_callback.py
```diff
@@ -138,18 +139,19 @@ class DownstreamEvaluatorCallback(Callback):
...
         """Run the evaluators."""
         for evaluator in self.evaluators:
-            # Convert epochs to steps
             eval_interval_steps = self.trainer.convert_duration_to_steps(
-                evaluator.eval_duration
+                evaluator.eval_interval
+            )
+            if self.step <= 1 or self.step % eval_interval_steps != 0:
+                continue
+            evaluator.eval_duration
```
favyen2: remove

favyen2: (I mean the `evaluator.eval_duration` line)

PR #159, opened 2025-03-21
ripping out float8

pjreddie:
Olmo-core changed Float8 api, we aren't using it anyway at the moment. Started an experiment but GPUs are pretty full up, I'll comment when it finds a home


helios/train/train_module/galileo.py
Hgherzog, line 1: Can we pin to 2.0.1 in this PR?

PR #163, opened 2025-03-26
FSDP-first-Pass

Hgherzog:
Initial Implementation of FSDP

- Enables FSDP with a single not configurable wrapping strategy TODO: is to test and try different wrapping strategies
- Enables Batch Size 32 per GPU whereas DDP only allows 16 for a 1.1 billion parameter model
- Currently still very slow to do forward pass with 1B model
- Uses FSDP 2 https://github.com/pytorch/torchtitan/blob/main/docs/fsdp.md 
- Includes scripts for testing these FSDP setting for each of the model variants
- Bumps olmo-core to most recent commit so that we can use the distribute_like function

Right now we are still too slow so I wrote out some TODOs on how we get faster https://github.com/allenai/helios/issues/167

### DIFF
--- a/helios/train/train_module/train_module.py
+++ b/helios/train/train_module/train_module.py
@@ -37,7 +37,6 @@ from torch.optim import Optimizer
 
 from helios.data.transform import TransformConfig
 
-
 logger = getLogger(__name__)
 
 
@@ -475,7 +474,6 @@ class HeliosTrainModule(TrainModule):
             for param, target_param in zip(
                 self.model.encoder.parameters(), self.model.target_encoder.parameters()
             ):
-
                 # TODO: Make this an in place operation
                 target_param.data = cur_ema_value * get_full_tensor(
                     target_param.data


helios/train/callbacks/speed_monitor.py
```diff
@@ -141,3 +145,9 @@ def post_step(self) -> None:
         self.trainer.record_metric("throughput/device/data loading (%)", data_pct)
         self.trainer.record_metric("throughput/device/BPS", bps)
         self.trainer.record_metric("throughput/device/BPS (estimated avg)", bps_avg)
+        self.trainer.record_metric(
+            "throughput/device/model duration (s)", self.model_duration
+        )
+        self.trainer.record_metric(
+            "throughput/device/model duration (%)", self.model_duration / step_time
+        )
```
yawenzzzz, line 153: Nice metrics!

helios/nn/flexi_patch_embed.py
```diff
@@ -152,6 +153,8 @@ def resize_patch_embed(
             )
         pinv = self.pinvs[new_patch_size]
         pinv = pinv.to(patch_embed.device)
+        # # I don't know if this will work without FSDP
```
yawenzzzz: there's a check on source dtype, if it's not a DTensor, then non-sharded target will be returned, so it should work even without FSDP

https://github.com/allenai/OLMo-core/blob/ecb14e03f9e5f7374b81cc9f924fd13a4b2482f2/src/olmo_core/distributed/utils.py#L442

helios/nn/flexihelios.py
```diff
@@ -1181,6 +1192,14 @@ def forward(
             )
         return TokensAndMasks(**patchified_tokens_and_masks)
 
+    def apply_fsdp(self, **fsdp_kwargs: Any) -> None:
+        """Apply FSDP to the model."""
+        super().apply_fsdp(**fsdp_kwargs)
+        fully_shard(self.patch_embeddings, **fsdp_kwargs)
+        register_fsdp_forward_method(self.patch_embeddings, "forward")
+        # fully_shard(self.norm, **fsdp_kwargs)
```
yawenzzzz: suggestion: either remove it or add a comment to this commented code

scripts/fsdp_comparisons/base_fsdp_galileo.py
```diff
@@ -0,0 +1,322 @@
...
+
+ENCODER_EMBEDDING_SIZE = 1024
+DECODER_EMBEDDING_SIZE = 768
+ENCODER_DEPTH = 24
+DECODER_DEPTH = 12
+ENCODER_NUM_HEADS = 16
+DECODER_NUM_HEADS = 12
+MLP_RATIO = 4.0
+
+# Vit-L
```
yawenzzzz, line 68: Since now we have ViT-T, ViT-B, ViT-L, ViT-H, for easier experiments, we can add them directly into model config, like https://github.com/allenai/OLMo-core/blob/main/src/olmo_core/nn/transformer/config.py#L394. However, no need to resolve in this PR, once we finalize the decoder depth, we can set up this.

scripts/base_debug_scripts/galileo.py
```diff
@@ -46,14 +46,32 @@
...
+    # DECODER_DEPTH = 4
+    # ENCODER_NUM_HEADS = 8
+    # DECODER_NUM_HEADS = 8
+    # MLP_RATIO = 4.0
+    ENCODER_EMBEDDING_SIZE = 1024
+    DECODER_EMBEDDING_SIZE = 768
+    ENCODER_DEPTH = 24
+    DECODER_DEPTH = 12
+    ENCODER_NUM_HEADS = 16
+    DECODER_NUM_HEADS = 12
```
yawenzzzz: Suggestion: bring defaults back for debugging

scripts/fsdp_comparisons/base_fsdp_galileo.py
Hgherzog: https://github.com/allenai/helios/issues/173

PR #168, opened 2025-03-28
Obvious simplifications

Hgherzog:
- Move transform to base helios train module
- move eval batch to helios train module and remove eval loss
- rmeove update target encoder to helios train module


scripts/base_debug_scripts/mae.py
pjreddie, line 1: nit: Since we're using flip_and_rotate for everything maybe we push this into the defaults at some point?

Hgherzog: I moved the default to flip_and_rotate in the train module

PR #169, opened 2025-03-31
make sure we don't pass encodings to the exit zero output for token d…

Hgherzog:
- In Galileo, encodings are not applied when target exit depth is zero because this allows the model to more trivially predict the encoding that it has also seen.
- In Helios, we previously were applying encodings to exit depth zero because we did not also set the exit_after_n to zero

I removed this extra parameter and am deriving whether to do attention  and add encodings based solely on the exit config values.

https://wandb.ai/eai-ai2/helios-debug/runs/q2j0d82e?nw=nwuserhenryhzog

### DIFF
--- a/helios/nn/galileo.py
+++ b/helios/nn/galileo.py
@@ -6,7 +6,6 @@ from dataclasses import dataclass
 import torch.nn as nn
 from olmo_core.config import Config
 
-from helios.data.transform import Transform, TransformConfig
 from helios.nn.flexihelios import EncoderConfig, PredictorConfig, TokensAndMasks
 from helios.nn.utils import DistributedMixins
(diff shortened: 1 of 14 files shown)


helios/nn/flexihelios.py
```diff
@@ -1084,13 +1083,30 @@ def apply_attn(
         )
         # exited tokens are just the linear projection
         exited_tokens, _ = self.collapse_and_combine_hwtc(x)
+        # Get the keys of the modalities that should exit at depth 0
```
favyen2: Suggest to include comment here explaining the motivation for skipping the composite_encodings for cases where user specifies depth 0 exit.

PR #170, opened 2025-04-01
KoLeo regularizer

gabrieltseng:
Implements the KoLeo regularizer

Screenshot 2025-04-01 at 12 50 10 PM

Example run: https://wandb.ai/eai-ai2/helios-debug/runs/vj4k4bgo

### DIFF
--- a/helios/train/train_module/galileo.py
+++ b/helios/train/train_module/galileo.py
@@ -19,7 +19,7 @@ from helios.data.constants import Modality
 from helios.data.dataset import HeliosSample
 from helios.data.transform import TransformConfig
 from helios.nn.flexihelios import TokensAndMasks
-from helios.nn.latent_mim import LatentMIM
+from helios.nn.galileo import Galileo
 from helios.train.loss import LossConfig
 from helios.train.masking import MaskedHeliosSample, MaskingConfig
 from helios.train.train_module.train_module import (
@@ -65,7 +65,7 @@ class GalileoTrainModuleConfig(HeliosTrainModuleConfig):
 
     def build(
         self,
-        model: LatentMIM,
+        model: Galileo,
         device: torch.device | None = None,
     ) -> "GalileoTrainModule":
         """Build the corresponding :class:`LatentMIMTrainModule`.
@@ -74,7 +74,7 @@ class GalileoTrainModuleConfig(HeliosTrainModuleConfig):
             model: The model to train.
             device: The device to train on.
         """
-        kwargs = self.as_dict(exclude_none=True, recurse=False)
+        kwargs = self.prepare_kwargs()
         return GalileoTrainModule(
             model=model,
             device=device,
@@ -112,7 +112,7 @@ class GalileoTrainModule(HeliosTrainModule):
 
     def __init__(
         self,
-        model: LatentMIM,
+        model: Galileo,
         optim_config: OptimConfig,
         transform_config: TransformConfig,
         masking_config_a: MaskingConfig,
@@ -242,22 +242,20 @@ class GalileoTrainModule(HeliosTrainModule):
                     )
 
                     # Run Encoder and decoder on the augmented input
-                    latent, decoded, target_output = self.model_forward_a(
+                    loss, latent, decoded, target_output = self.model_forward_a(
                         masked_batch, patch_size, self.token_exit_cfg_a
                     )
-                    loss = self.loss_fn_a(decoded, target_output)
                 else:
                     masked_batch = self.masking_strategy_b.apply_mask(
                         microbatch, patch_size=patch_size
                     )
 
                     # Run Encoder and decoder on the augmented input
-                    latent, decoded, target_output = self.model_forward_b(
+                    loss, latent, decoded, target_output = self.model_forward_b(
                         masked_batch, patch_size, self.token_exit_cfg_b
                     )
-                    loss = self.loss_fn_b(decoded, target_output)
+
                 # Scale loss by number of microbatches
-                reg_term = self.compute_regularization(latent)
                 reg_term = self.compute_regularization(latent)
                 if reg_term is not None:
                     loss = loss + reg_term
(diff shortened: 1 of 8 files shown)


helios/train/loss.py
```diff
@@ -517,6 +517,73 @@ def compute(
...
+                _online encoder outputs_, not decoder outputs.
+            targets: Unused, and only kept for consistency.
+            **kwargs: Additional keyword arguments.
+
+        Returns:
+            The computed loss value.
+        """
+        all_preds, all_masks = predictions.flatten_tokens_and_masks()
+        online_encodings = all_preds[all_masks == MaskValue.ONLINE_ENCODER.value]
+
```
gabrieltseng, line 596: Currently this is computed against **all patch embeddings**. Could this be too expensive? It runs at the moment but I believe Dinov2 does it against "backbone embeddings" - one thing we could do is compute an instance average before this regularizer is applied.

gabrieltseng: This OOMs on Galileo with a microbatch size of 128 but works with 64
https://wandb.ai/eai-ai2/helios-debug/runs/8t4ktbhi

Screenshot 2025-04-01 at 2 04 35 PM

Hgherzog: I think an instance average makes sense and aligns more closely with what is done in dino and aligns with using a mean pooling of the features

helios/train/train_module/train_module.py
```diff
@@ -484,3 +488,12 @@ def eval_batch(
...
         raise NotImplementedError("eval batch not implemented")
+
+    def add_regularizer_to_loss(
+        self, loss: torch.Tensor, latent: TokensAndMasks
+    ) -> torch.Tensor:
+        """If a regularizer is present, add it to the loss."""
+        regularizer = getattr(self, "regularizer", None)
+        if regularizer is None:
+            return loss
+        return loss + regularizer.compute(latent, None)
```
Hgherzog: Suggestion: log the regularizer metric separetly as well

helios/train/train_module/latent_mim.py
```diff
@@ -166,6 +169,13 @@ def __init__(
         self.token_exit_cfg = token_exit_cfg
         self.base_loss = loss_config.build()
         self.masking_strategy = masking_config.build()
+        self.regularizer = (
+            regularizer_config.build() if regularizer_config is not None else None
+        )
+
+        self.metric_name = self.base_loss.name
```
Hgherzog: Suggestion: metric name is unclear, perhaps total_loss_name?

helios/train/loss.py
```diff
@@ -517,6 +517,73 @@ def compute(
...
+        """
+        # parwise dot products (= inverse distance)
+        dots = torch.mm(x, x.t())
+        n = x.shape[0]
+        dots.view(-1)[:: (n + 1)].fill_(-1)  # Trick to fill diagonal with -1
+        # max inner prod -> min distance
+        _, indices = torch.max(dots, dim=1)
+        return indices
+
+    def compute(
```
Hgherzog, line 575: we should add some way to weight the regularizer here.  See page 29 https://arxiv.org/pdf/2304.07193 DinoV2 uses a weight of 0.1

helios/train/train_module/mae.py
```diff
@@ -160,6 +162,12 @@ def __init__(
         self.token_exit_cfg = token_exit_cfg
         self.base_loss = loss_config.build()
         self.masking_strategy = masking_config.build()
+        self.regularizer = (
+            regularizer_config.build() if regularizer_config is not None else None
+        )
+        self.metric_name = self.base_loss.name
+        if self.regularizer is not None:
```
Hgherzog, line 169: Suggestion: Another option instead of having to bind the names here is to either make a meta loss class that allows for arbitrary weighted combinations of losses. This also might be more amenable to the case where we want to add multiple regularizing terms

helios/train/loss.py
gabrieltseng: updated

helios/train/loss.py
gabrieltseng: added

helios/train/train_module/mae.py
gabrieltseng: I agree and this is what I originally intended to do, but a challenge is that different losses receive different inputs (e.g. this regularizer is computed against the latents), and its generally a bit module specific (e.g. Galileo's losses being branch specific)

I guess we could have all loss functions receive *three* inputs (latents, preds, targets) but this seems like it would be good to do when we have an example of at least one other loss which requires the latents to establish the pattern

gabrieltseng: Actually on reflection this is possible but I'd prefer to do it in a separate PR to keep the changes contained. 

Issue created: https://github.com/allenai/helios/issues/175

helios/train/train_module/latent_mim.py
gabrieltseng: updated

helios/train/train_module/mae.py
```diff
@@ -223,7 +231,7 @@ def train_batch(
                 loss.backward()
 
         self.trainer.record_metric(
-            f"train/{self.base_loss.name}",
+            f"train/{self.total_loss_name}",
             total_batch_loss / get_world_size(self.dp_process_group),
             ReduceType.mean,
```
gabrieltseng, line 250: MAE is the only module to divide the `total_batch_loss` by `get_world_size(self.dp_process_group)` - probably the other modules should be doing this or MAE shouldn't?

helios/train/train_module/galileo.py
```diff
@@ -212,13 +221,9 @@ def train_batch(
         # Set the model to train mode
         self.model.train()
 
-        # This is a clear loss buffer
-        if not hasattr(self, "total_batch_loss"):
-            self.total_batch_loss = torch.zeros(1, device=self.device)
-        else:
-            self.total_batch_loss.fill_(0.0)
```
gabrieltseng, line 219: I assume this is outdated code and safe to remove? no other train module has it.

helios/train/train_module/train_module.py
gabrieltseng: okay - added but I'd appreciate if you double checked my implementation (plus two other comments)

helios/train/train_module/mae.py
Hgherzog: Yes MAE shouldn't Reduce_Type.mean already averages the metric across the gpus

Hgherzog: by yes I mean that MAE shouldn't

helios/train/train_module/galileo.py
Hgherzog: The idea here was to remove a host device sync, it ended up not really making a big difference though because we have to sync again immediately after. Option to either remove or add methods to helios train module. This would be like a set and clear loss buffers method (I think the trainer already calls clear loss buffers)

PR #172, opened 2025-04-01
Enable training on multiple Helios datasets

favyen2:
Support training on multiple Helios datasets where they will be treated as one merged dataset.

The Helios dataset doesn't need to be copied but a different set of h5s will be generated.

Note: this may cause issues with checkpoints since dataset fingerprint is being changed, and also the arguments to HeliosDataset are changed so scripts need update.

### DIFF
--- a/scripts/base_debug_scripts/galileo.py
+++ b/scripts/base_debug_scripts/galileo.py
@@ -179,7 +179,9 @@ def build_dataset_config(common: CommonComponents) -> HeliosDatasetConfig:
     h5py_dir = "/weka/dfive-default/helios/dataset/presto/h5py_data/latlon_sentinel1_sentinel2_l2a_worldcover/98856"
     return HeliosDatasetConfig(
         h5py_dir=h5py_dir,
-        tile_path=None,
+        tile_paths=[
+            "/weka/dfive-default/helios/dataset/presto/",
+        ],
         supported_modality_names=common.supported_modality_names,
         dtype=DType.float32,
     )
--- a/scripts/base_debug_scripts/latent_mim.py
+++ b/scripts/base_debug_scripts/latent_mim.py
@@ -154,7 +154,9 @@ def build_dataset_config(common: CommonComponents) -> HeliosDatasetConfig:
     h5py_dir = "/weka/dfive-default/helios/dataset/presto/h5py_data/latlon_sentinel1_sentinel2_l2a_worldcover/98856"
     return HeliosDatasetConfig(
         h5py_dir=h5py_dir,
-        tile_path=None,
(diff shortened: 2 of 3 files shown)


helios/internal/common.py
```diff
@@ -124,6 +124,11 @@ def build_launch_config(
             #  "pip install --upgrade --pre torch==2.6.0.dev20241112+cu121 --index-url https://download.pytorch.org/whl/nightly/cu121",
             "pip freeze",
         ],
+        # The default in olmo-core is to enable host networking on Augusta but disable
+        # it on other clusters. Since we often use just one GPU, we need to disable it
+        # on Augusta too, otherwise if multiple jobs are assigned to the same
+        # hypervisor they will fail due to binding to the same port for torchrun.
+        host_networking=False,
```
Hgherzog, line 131: Can you add logic that handles if this is multi-gpu or not

helios/data/dataset.py
```diff
@@ -495,7 +490,7 @@ def set_h5py_dir(self, num_samples: int) -> None:
             return
 
         self.h5py_dir = (
-            self.tile_path
+            self.tile_paths[0]
```
Hgherzog, line 493: What happens if the tile_paths are in a different order? Should we sort the tile paths initially?

helios/data/dataset.py
Hgherzog, line 1: with this implementation, it seems to me like we have to recreate the h5s to use both datasets, perhaps there is a way we could concatenate two existing datasets and just use the sample index modulo total number of samples to figure out which base path to use?

helios/data/dataset.py
```diff
@@ -345,8 +345,8 @@ def __init__(
         self,
         supported_modalities: list[ModalitySpec],
         dtype: DType,
+        tile_paths: list[UPath],
```
Hgherzog, line 348: tile_paths should still be optional. The intended path during training is to pass in just the h5py dir that you want to use.

helios/internal/common.py
pjreddie, line 131: Do we have scripts where this needs to be on?

helios/data/dataset.py
```diff
@@ -854,13 +849,6 @@ def validate(self) -> None:
         Raises:
             ValueError: If any arguments are invalid
         """
-        # Validate tile_path
```
pjreddie, line 857: Do we need to validate the list? could loop over and validate each path or is this no longer necessary?

favyen2: The validation was to check that one of tile_path or h5py_dir is set, I removed it since now tile_paths is required and h5py_dir is optional.

PR #179, opened 2025-04-03
Reduce syncs

pjreddie:
Reduce syncs in:
- masking
- patch resizing
need to test:
- always running embedding instead of testing if tokens are seen by encoder

### DIFF
--- a/helios/nn/flexihelios.py
+++ b/helios/nn/flexihelios.py
@@ -976,7 +976,6 @@ class Encoder(FlexiHeliosBase):
             updated_mask: [B, T]
             where T is the max number of unmasked tokens for an instance
         """
-        # At this point when we flip the mask 1 means keep 0 means remove
         sorted_mask, indices = torch.sort(mask, dim=1, descending=True, stable=True)
         # Now all the places where we want to keep the token are at the front of the tensor
         x = x.gather(1, indices[:, :, None].expand_as(x))


helios/nn/flexihelios.py
```diff
@@ -975,8 +963,8 @@ def remove_masked_tokens(x: Tensor, mask: Tensor) -> tuple[Tensor, Tensor, Tenso
         Implementation from https://stackoverflow.com/a/68621610/2332296
 
         On Input:
-        1 means this token should be removed
```
pjreddie, line 978: FYI this is a significant change, idk why it was the other way before but there were a lot of ~mask.bool() that I could get rid of by reversing it. We now assume the masked tensor is a boolean tensor with 1 for keeping

helios/nn/flexihelios.py
```diff
@@ -988,12 +976,8 @@ def remove_masked_tokens(x: Tensor, mask: Tensor) -> tuple[Tensor, Tensor, Tenso
             updated_mask: [B, T]
             where T is the max number of unmasked tokens for an instance
         """
-        org_mask_dtype = mask.dtype
-        mask = mask.bool()
         # At this point when we flip the mask 1 means keep 0 means remove
```
yawenzzzz: nit: remove this comment to avoid confusion

helios/train/masking.py
```diff
@@ -414,23 +418,28 @@ def _create_spatial_mask(
...
+                    MaskValue.TARGET_ENCODER_ONLY.value,
+                    device=device,
+                ),
+            ]
         )
 
-        # numpy to for permuted function
-        batch_mask = repeat(flat_mask, "x -> b x", b=b)
-        random_batch_mask = self.generator.permuted(batch_mask, axis=1)
+        masks = [flat_mask[torch.randperm(patches, device=device)] for i in range(b)]
```
yawenzzzz, line 435: Since `seed_all` is applied, this one will be deterministic based on `patches` so different spatial modalities will have the same region being masked, and for microbatches with the same `patch_size`, this spatial mask will always be the same. I'm wondering if it will be helpful to introduce some randomness, either within modalities, and/or within microbatches. No need to add anything in this PR.

PR #181, opened 2025-04-03
Henryh/apply model compile

Hgherzog:
add ability to compile the transformer layers, we should see more performance increases on better hardware and larger models

UPDATES base debug scripts to use bflaot16

torch tensor cycle warning does not work with torch compile

### DIFF
--- a/helios/nn/latent_mim.py
+++ b/helios/nn/latent_mim.py
@@ -72,7 +72,7 @@ class LatentMIM(nn.Module, DistributedMixins):
         self.decoder.apply_compile()
         logger.info("Applied torch.compile to the decoder")
         self.target_encoder.apply_compile()
-        # logger.info("Applied torch.compile to the target encoder")
+        logger.info("Applied torch.compile to the target encoder")
 
 
 @dataclass


helios/train/train_module/train_module.py
```diff
@@ -212,8 +212,13 @@ def __init__(
 
         # Maybe compile.
         if compile_model:
-            self.model.apply_compile()
-            logger.info("Applied torch.compile() to the model")
+            if torch.cuda.is_available():
```
favyen2, line 215: Should this just raise error if cuda is not available but compile_model flag is set?

helios/nn/latent_mim.py
```diff
@@ -61,6 +64,16 @@ def apply_fsdp(
...
 
+    def apply_compile(self) -> None:
+        """Apply torch.compile to the model."""
+        logger.info("Applying torch.compile to the model")
+        self.encoder.apply_compile()
+        logger.info("Applied torch.compile to the encoder")
+        self.decoder.apply_compile()
+        logger.info("Applied torch.compile to the decoder")
+        self.target_encoder.apply_compile()
+        # logger.info("Applied torch.compile to the target encoder")
```
favyen2: Should this be uncommented (to keep consistency with the other lines above)?

helios/internal/experiment.py
```diff
@@ -163,7 +162,6 @@ def train(config: HeliosExperimentConfig) -> None:
     config_dict = config.as_config_dict()
     cast(WandBCallback, trainer.callbacks["wandb"]).config = config_dict
     cast(ConfigSaverCallback, trainer.callbacks["config_saver"]).config = config_dict
-    warn_tensor_cycles()
```
favyen2, line 166: Should this still run if compile_model flag is false or it's fine to remove it now?

Hgherzog: I think it is fine to remove and we can just add it back for specific debugging. I left it in because I didn;t realize it would conflict with anything but it breaks compile and it was a weird bug so would want to avoid in future.

helios/train/train_module/train_module.py
Hgherzog: I think a warning is fine because it likely was a mistake while debugging. It wouldnt make sense to debug compile without a gpu, so the debugging would be unrelated.

PR #184, opened 2025-04-08
Contrastive loss term

gabrieltseng:
See here for motivation: https://docs.google.com/presentation/d/1zeDlEqFMxBhOujxtFMDDfKEe6v07n2HpDhNWKT-fgv8/edit?usp=sharing

### DIFF
--- a/tests/unit/train/test_loss.py
+++ b/tests/unit/train/test_loss.py
@@ -8,6 +8,7 @@ from helios.nn.flexihelios import TokensAndMasks
 from helios.train.loss import (
     AdjustedPatchDiscriminationLoss,
     CrossEntropyLoss,
+    InfoNCELoss,
     L1Loss,
     L2Loss,
     PatchDiscriminationLoss,
@@ -222,3 +223,18 @@ def test_cross_entropy_loss() -> None:
     loss_value = loss.compute(preds, targets)
     # loss for BCE, prediction of .5 for both classes
     assert torch.isclose(loss_value, -torch.log(torch.tensor(0.5)), 0.0001)
+
+
+def test_infonce_loss() -> None:
+    """Just test that it runs as expected."""
+    b, d = 16, 128
+
+    loss = InfoNCELoss()
+    loss_value = loss.compute(torch.ones((b, d)), torch.zeros((b, d)))
+    # not very good! since they are all the same
+    # predictions and values
+    assert loss_value > 0.5
+    # check the weight
+    loss = InfoNCELoss(weight=0.1)
+    w_loss_value = loss.compute(torch.ones((b, d)), torch.zeros((b, d)))
+    assert 0.1 * loss_value == w_loss_value


helios/train/train_module/galileo.py
```diff
@@ -236,30 +244,41 @@ def train_batch(
...
+                        get_local_tensor((reg_term_a + reg_term_b) / 2)
+                        / num_microbatches
                     )
 
-                    # Run Encoder and decoder on the augmented input
-                    loss, latent, decoded, target_output = self.model_forward_b(
-                        masked_batch, patch_size, self.token_exit_cfg_b
+                if self.contrastive_loss is not None:
+                    contrastive_loss = self.contrastive_loss.compute(latent_a, latent_b)
+                    loss += contrastive_loss
```
gabrieltseng, line 281: Strictly speaking there should be a linear layer between the pooled latents and the contrastive loss. See the CROMA paper below

Screenshot 2025-04-08 at 12 50 27 PM

helios/train/train_module/galileo.py
```diff
@@ -236,30 +244,41 @@ def train_batch(
...
-                    )
+                loss_a, latent_a = self.apply_masks_and_compute_losses_and_latents(
+                    microbatch,
+                    self.masking_strategy_a.apply_mask,
+                    self.model_forward_a,
+                    patch_size,
+                    self.token_exit_cfg_a,
+                )
+                loss_b, latent_b = self.apply_masks_and_compute_losses_and_latents(
+                    microbatch,
```
gabrieltseng: This applies different masks to the same microbatch, which better mirrors Galileo.

Also it might speed training up? Since we get more learning signal for the same data loading cost.

helios/train/train_module/galileo.py
gabrieltseng: added: https://wandb.ai/eai-ai2/helios-debug/runs/o38klaxi?nw=nwusergabrieltseng

pjreddie: Now you run the model twice per batch with both masking strategies etc instead of alternating right? Are we at all worried about backward compatibility since this changes from previous experiments? Probably doesn't matter really though. Is this how Galileo (NASA version) works?

helios/train/loss.py
```diff
@@ -514,6 +514,52 @@ def compute(
...
+class InfoNCELoss(Loss):
+    """Loss function for InfoNCE."""
+
+    name = "InfoNCE"
+
+    def __init__(
+        self,
+        tau: float = 0.1,
+    ):
+        """Initialize adjusted patch discrimination loss.
```
pjreddie: Nit: probably should be `"""Initialize InfoNCE loss"""`

helios/train/loss.py
pjreddie, line 1: probably should add a test for the new loss

helios/train/train_module/galileo.py
gabrieltseng: yes. it does affect compatibility but not massively i think (?)

and sort of - Galileo did use this approach (multiple re-uses of a same minibatch) but (i) did it 4 times, and (ii) subsampled afresh each time.

helios/train/loss.py
gabrieltseng: added!

PR #188, opened 2025-04-09
Titan training

pjreddie:


helios/internal/common.py
```diff
@@ -118,10 +125,12 @@ def build_launch_config(
             "git submodule update --init --recursive",
             # Setup python environment.
             "conda shell.bash activate base",
+            # Install torch==2.7 if we're targetting titan
             "pip install -e '.[all]'",
             "pip install --upgrade beaker-py",
             # Quickly try a new version of PyTorch like this
             #  "pip install --upgrade --pre torch==2.6.0.dev20241112+cu121 --index-url https://download.pytorch.org/whl/nightly/cu121",
+            pytorch_upgrade,
```
Hgherzog, line 133: pytorch upgrade should default to something that does nothing if we are not using titan

pjreddie: yeah the default is an empty string

Issue #154, opened 2025-03-21
Move non dataset needed functions out of dataset class


Hgherzog:
Dataset class is getting big filtering samples and computing normalization values don't really fit here so this should be refactored

PR #190, opened 2025-04-09
Henryh/add srtm naip landsat dataset

Hgherzog:
This PR accomplishes 3 main things: 

1.  Enables training with landsat and srtm data
2. Enables us to use the same set of files for training with different combinations of modalities (allowing us to reduce the amount of data we need in hot storage) 
3. Separates creating the h5 dataset and training on the h5 dataset into different modules

Notes on 1:
- I computed new normalization values for landsat srtm and naip so they could be trained on
Notes on 2:

-  supported_modalities now is used to say what modalities the dataset contains
- training_modalities says what you want to use during training
- we are able to start trying stuff with naip because it is in the dataset 

presto: `/weka/dfive-default/helios/dataset/presto/h5py_data/landsat_naip_openstreetmap_raster_sentinel1_sentinel2_l2a_srtm_worldcover/118861`

osm_sampling: `/weka/dfive-default/helios/dataset/osm_sampling/h5py_data/landsat_naip_openstreetmap_raster_sentinel1_sentinel2_l2a_srtm_worldcover/324192`

- we can now add filtering logic for the samples based on a sample_metadata.csv that indicates which samples have which modalities (e.g exclude all samples with naip from training or don't use samples with only osm and worldcover) 

Notes on 3: addresses #154 
- tile_path is discontinued from the dataset class
- **There is now a single way to use the dataset class based on h5py dirs** 

For each dataset we are using we will want to delete the old copies and just filter using training modalities going forward to reduce our data storage overhead.

### DIFF
--- a/helios/data/dataset.py
+++ b/helios/data/dataset.py
@@ -732,10 +732,6 @@ class HeliosDatasetConfig(Config):
         # Validate supported_modalities
         if not isinstance(self.training_modalities, list):
             raise ValueError("training_modalities must be a list")
-        # if not all(isinstance(m, ModalitySpec) for m in self.training_modalities):
-        #     raise ValueError(
-        #         "All elements in supported_modalities must be ModalitySpec"
-        #     )
 
     @property
     def h5py_dir_upath(self) -> UPath:


helios/internal/run_h5_conversion.py
```diff
@@ -0,0 +1,53 @@
...
+            Modality.SRTM.name,
+            Modality.NAIP.name,
+            Modality.LANDSAT.name,
+            Modality.OPENSTREETMAP_RASTER.name,
+        ],
+        multiprocessed_h5_creation=True,
+    )
+
+
+def main(config_builder: Callable = build_default_config, *args: Any) -> None:
```
favyen2, line 41: It would be useful to have documentation somewhere about an example way to use this script (with the overrides and such).

helios/data/dataset.py
```diff
@@ -447,97 +407,87 @@ def fingerprint(self) -> str:
...
-                )
-        else:
-            for i, sample in enumerate(samples):
-                self.process_sample_into_h5((i, sample))
-
-    def set_h5py_dir(self, num_samples: int) -> None:
-        """Set the h5py directory.
+    def is_dataset_prepared(self) -> bool:
+        """Check if the dataset is prepared."""
+        return self.sample_indices is not None and self.h5py_dir.exists()
```
favyen2: prepare() would not create the h5py_dir so it may make sense to move the `self.h5py_dir.exists` check to `__init__`.

helios/data/dataset.py
```diff
@@ -924,30 +727,13 @@ def validate(self) -> None:
...
-        """Get the supported modalities."""
-        return get_modality_specs_from_names(self.supported_modality_names)
-
-    @property
-    def tile_upath(self) -> UPath:
-        """Get the tile path."""
-        return UPath(self.tile_path)
+        if not isinstance(self.training_modalities, list):
+            raise ValueError("training_modalities must be a list")
+        # if not all(isinstance(m, ModalitySpec) for m in self.training_modalities):
```
favyen2: cleanup commented code

PR #191, opened 2025-04-09
Train on multiple datasets v2

favyen2:
Updated to use a dataset wrapper around list of existing HeliosDatasets.

here is the previous closed pr for reference:
https://github.com/allenai/helios/pull/172

### DIFF
--- a/scripts/2025_04_09_multidataset/sweep.py
+++ b/scripts/2025_04_09_multidataset/sweep.py
@@ -126,7 +126,7 @@ for token_exit_args, model_args in itertools.product(
     MODEL_ARGS,
 ):
     # Construct run name indicating hyperparameters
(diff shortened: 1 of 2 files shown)


helios/data/concat.py
```diff
@@ -0,0 +1,126 @@
...
+from torch.utils.data import ConcatDataset, Dataset
+
+from helios.data.constants import ModalitySpec
+
+from .dataset import GetItemArgs
+
+logger = logging.getLogger(__name__)
+
+
+class HeliosConcatDataset(ConcatDataset):
```
Hgherzog, line 20: Can you add some unit tests for this? Also are all the work_dir and fs_local_rank methods in helios dataset not actually being used? From a brief glance at olmo_core it seems like it is only required to have these in the Dataloader. Would be great to remove them if so

favyen2: added test

PR #192, opened 2025-04-14
Pjreddie/latent mae

pjreddie:
experiments with latent mim + mae

### DIFF
--- a/tests/integration/nn/test_latent_mim.py
+++ b/tests/integration/nn/test_latent_mim.py
@@ -87,7 +87,7 @@ def test_latentmim_with_loss(
     )
     latentmim = LatentMIM(encoder, predictor)
 
-    _, output, _ = latentmim.forward(x, patch_size)
+    _, output, _, _ = latentmim.forward(x, patch_size)
     output = predictor.forward(output, x.timestamps, patch_size, input_res=1)
     patched_H = H // patch_size
     patched_W = W // patch_size


helios/train/train_module/latent_mim.py
pjreddie, line 1: don't merge these changes in, just for experiment

helios/nn/galileo.py
```diff
@@ -126,7 +136,9 @@ def build(self) -> "Galileo":
         self.validate()
         encoder = self.encoder_config.build()
         decoder = self.decoder_config.build()
+        reconstructor = self.reconstructor_config and self.reconstructor_config.build()
```
favyen2: I think this would be more clear as an if statement because it's not immediately obvious that the result is None as opposed to say False if reconstructor_config is None.

Unless we are already using this pattern everywhere.

Actually I just checked and it seems elsewhere we use this pattern which I think is more clear:

```python
self.regularizer = regularizer_config.build() if regularizer_config is not None else None
```

helios/nn/galileo.py
```diff
@@ -30,38 +28,47 @@ def __init__(
...
         self.target_encoder = deepcopy(self.encoder)
+        self.reconstructor = reconstructor
         for p in self.target_encoder.parameters():
             p.requires_grad = False
 
     def forward_a(
         self, x: MaskedHeliosSample, patch_size: int
-    ) -> tuple[TokensAndMasks, TokensAndMasks, torch.Tensor]:
+    ) -> tuple[TokensAndMasks, TokensAndMasks, torch.Tensor, TokensAndMasks | None]:
         """Forward pass for the Latent MIM Style."""
```
favyen2: There are many return values now so it may be worth documenting it in the doc string.

helios/train/train_module/mae.py
```diff
@@ -40,9 +39,12 @@ class MAETrainModuleConfig(HeliosTrainModuleConfig):
         masking_config: The masking configuration for the model.
     """
 
-    loss_config: LossConfig = field(
-        default_factory=lambda: LossConfig(loss_config={"type": "mae"})
+    mae_loss_config: LossConfig | None = field(
```
favyen2, line 42: Does this mean the MAE scripts have to be updated to use this name?

pjreddie: yes I'll do this

helios/nn/flexihelios.py
```diff
@@ -410,29 +410,38 @@ class Reconstructor(nn.Module):
...
         for modality in self.supported_modalities:
             self.per_modality_reconstructions[modality.name] = (
                 self._get_patch_reconstruction_module_for_modality(modality)
             )
 
+    def apply_compile(self) -> None:
+        """Apply torch.compile to the model."""
+        self.decoder.apply_compile()
+
+    def apply_fsdp(self, **fsdp_kwargs: Any) -> None:
```
pjreddie, line 441: @Hgherzog can you double check this?

Hgherzog: I think mae.py may also may need fsdp added to it as well. I would recommend running and debugging this to make sure every part is added. I would also recommend doing this in a separate PR that hasn't already been approved. It is okay to use DDP for this for now.

PR #196, opened 2025-04-16
Fix training when some samples don't have any multitemporal modalities

favyen2:


helios/data/dataset.py
```diff
@@ -471,22 +471,30 @@ def _filter_sample_indices_for_training(self) -> None:
             self.naip_indices = np.array([])
         logger.info(f"NAIP indices: {self.naip_indices}")
 
-        # Get the indices of samples that don't have any training modalities
-        no_training_indices = metadata_df[
```
Hgherzog, line 475: I think we still want to filter samplers where none of the requested training modalities are present, ie if you wnated to train only on a subset of modalities, samples that don't include any of those modalities should be removed

favyen2: I changed it to filter out samples without training modalities that are multi-temporal. If none of the training modalities are present, then none of the multi-temporal training modalities should be present, and therefore the sample should be filtered. Does that make sense or might there be an issue with the code?

Hgherzog: Should we raise an error then if there are no multitemporal training modalities?

PR #201, opened 2025-04-18
Masking experiments

favyen2:
This adds various masking strategies:

- RandomSpaceMaskingStrategy: combine space masking with random masking.
- RandomRangeMaskingStrategy: randomly pick a different encode ratio for each sample between user-provided low/high values.
- SelectableModalityMaskingStrategy: modality masking but support configuring certain modalities to always become decoder only, and others to always become missing.
- SelectableRandomRangeModalityMaskingStrategy: combine SelectableModalityMaskingStrategy with RandomRangeMaskingStrategy.

Of these only RandomSpaceMaskingStrategy seems good but the benefit is small so I'm not sure if it is actually helping. At https://wandb.ai/eai-ai2/2025_04_15_random_range compare train_randspace1_50 and train_randspace2_50 to train_old_40.

The two modality ones seem to reduce performance. RandomRangeMaskingStrategy has no impact on performance.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -45,7 +45,7 @@ repos:
         additional_dependencies:
           - "pydantic>=2.7.1,<3"
           - "types-protobuf"
-        exclude: ./.*_pb2_.*.py
+        exclude: (\./.*_pb2_.*\.py|^scripts/)
   - repo: https://github.com/PyCQA/bandit
     rev: "1.7.5"
     hooks:
--- a/tests/fixtures/__init__.py
+++ b/tests/fixtures/__init__.py
@@ -1,0 +1,1 @@
+"""Fixtures for Helios tests."""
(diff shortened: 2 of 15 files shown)


helios/train/masking.py
```diff
@@ -722,6 +758,228 @@ def apply_mask(
...
+class RandomRangeMaskingStrategy(MaskingStrategy):
+    """Randomly masks the input data."""
+
+    def __init__(
+        self,
+        min_encode_ratio: float = 0.1,
+        max_encode_ratio: float = 0.5,
+        min_decode_ratio: float = 0.1,
+        max_decode_ratio: float = 0.5,
+    ) -> None:
```
yawenzzzz: Can we also support the case where the user only provides the min/max encoder ratio? Then we can randomly pick a value in that range, and set the decoder ratio to 1 minus the encoder ratio.

favyen2: Yeah that's a good idea, added it plus a test for it!

PR #203, opened 2025-04-21
inference throughput benchmarking

cmwilhelm:
This changeset introduces a new module
for performing inference throughput benchmarking.

In the `helios` package, an encoder can be
loaded, and made to generate embeddings
in successively larger batch sizes. The benchmarking
run is parameterized on included modalities,
number of timestamps, bit precision, patch size,
image size, and batch sizes. Results are reported to `wandb`.

The update includes scripts to launch benchmarking
jobs to beaker based on desired GPU type; as well as one to
scrape results from wandb's API and prepare
a summary spreadsheet.

Initial results plotting max tokens-per-second on 1xA100
per model size as a function of instance sequence length:

Screenshot 2025-04-21 at 11 59 48 AM

Note: the above results required forcing flashattention by
dropping attn_masks in `attention.py`, a change not included
in this PR

### DIFF
--- a/helios/inference_benchmarking/run_throughput_benchmark.py
+++ b/helios/inference_benchmarking/run_throughput_benchmark.py
@@ -126,10 +126,11 @@ class ThroughputBenchmarkRunnerConfig(Config):
             raise ValueError("Either sweep_dict or sweep_keys must be set")
         if self.sweep_dict is not None and self.sweep_keys is not None:
             raise ValueError("Only one of sweep_dict or sweep_keys can be set")
-        if self.sweep_dict is None:
-            self.sweep_dict = {}
+        if self.sweep_dict is None and self.sweep_keys is not None:
+            sweep_dict: dict[str, Any] = {}
             for sweep_key in self.sweep_keys:
-                self.sweep_dict[sweep_key] = constants.SWEEPS[sweep_key]
+                sweep_dict[sweep_key] = constants.SWEEPS[sweep_key]
+            self.sweep_dict = sweep_dict
 
         return ThroughputBenchmarkRunner(
             default_run_params=self.default_run_params,
@@ -161,7 +162,7 @@ class ThroughputBenchmarkRunner:
         training_modalities: list[str],
         work_dir: Path,
         save_folder: Path | None = None,
-        sweep_dict: dict[str, str] = {},
+        sweep_dict: dict[str, Any] = {},
         cross_product_sweep: bool = False,
     ):
         """Initializes the throughput benchmarking runner."""


helios/inference_benchmarking/data_models.py
```diff
@@ -0,0 +1,141 @@
...
+            [
+                item
+                for item in [
+                    self.model_size,
+                    self.gpu_type,
+                    "bf16" if self.bf16 else None,
+                    "s1" if self.use_s1 else None,
+                    "s2" if self.use_s2 else None,
+                    "ls" if self.use_landsat else None,
+                    f"is{self.image_size}",
```
Hgherzog, line 45: Nit: use full words for image_size, patch_size and landsat for readability. (s1 and s2 are okay as they are called that in abbreviation.

helios/inference_benchmarking/run_throughput_benchmark.py
```diff
@@ -0,0 +1,225 @@
...
+import wandb
+from olmo_core.config import Config
+from olmo_core.distributed.checkpoint import load_model_and_optim_state
+
+from helios.inference_benchmarking import constants
+from helios.inference_benchmarking.data_models import RunParams
+from helios.nn.flexihelios import TokensAndMasks
+from helios.train.masking import MaskedHeliosSample, MaskValue
+
+NUM_S1_BANDS = 4
```
Hgherzog: in helios/data/constants.py we have constants for each modality that will give you the number of bands

helios/inference_benchmarking/run_throughput_benchmark.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def __init__(self, checkpoint_path: str):
+        """Loads the checkpoint, keeps only the encoder."""
+        super().__init__()
+
+        # Load the model config and initialize it.
+        # We avoid loading the train module here because it depends on running within
+        # olmo_core.
+        with open(f"{checkpoint_path}/config.json") as f:
+            config_dict = json.load(f)
```
Hgherzog: Suggestion: upload this config to wandb maybe under an additional "training_config" key, so that it easy to get a sense for any configurations during training that could affect inference throughput

helios/inference_benchmarking/run_throughput_benchmark.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+if __name__ == "__main__":
+    checkpoint_path = os.getenv(constants.PARAM_KEYS["checkpoint_path"], "/artifacts")
+    run_params = RunParams.from_env_vars()
+    project = os.getenv(constants.PARAM_KEYS["project"], constants.PROJECT_NAME)
+    owner = os.getenv(constants.PARAM_KEYS["owner"], constants.ENTITY_NAME)
+    name = os.getenv(constants.PARAM_KEYS["name"], "test")
+
+    print("Initializing wandb...")
```
Hgherzog: We generallly use prepare_cli_environment from olmo_core.utils to set up our logger and then we can log instead of print.

helios/inference_benchmarking/run_throughput_benchmark.py
```diff
@@ -0,0 +1,225 @@
...
+    wandb_dir = "/wandb"
```
Hgherzog: If the user does wandb login is setting up this dir still needed?

scripts/inference_benchmarking/launch_on_beaker.py
```diff
@@ -0,0 +1,99 @@
...
+        gpu_type="A100",
+        bf16=True,
+    )
+    for image_size in [64]
+    # for image_size in [1, 8, 16, 32, 64, 128]
+]
+
+
+if __name__ == "__main__":
+    b = Beaker.from_env(default_workspace=constants.BEAKER_WORKSPACE)
```
Hgherzog: I think this is fine for now, but we have been writing scripts in this style https://github.com/allenai/helios/blob/main/helios/internal/experiment.py https://github.com/allenai/helios/blob/main/helios/internal/run_h5_conversion.py for consistency. with this beaker launch config https://github.com/allenai/helios/blob/main/helios/internal/common.py. The main benefit is that configuration and overriding are done in a consistent way across the codebase.

helios/inference_benchmarking/run_throughput_benchmark.py
cmwilhelm: 👍

helios/inference_benchmarking/data_models.py
cmwilhelm: Agree on this nit, generally; but wandb allegedly has a 128-character limit on the names of runs, and we're already at ~60 chars with these terse abbreviations.

helios/inference_benchmarking/run_throughput_benchmark.py
cmwilhelm: 👍

scripts/inference_benchmarking/launch_on_beaker.py
cmwilhelm: 👀

helios/inference_benchmarking/run_throughput_benchmark.py
cmwilhelm: I'm not sure, I just replicated what I saw going on in the bowels of olmo-core. I assumed wandb was using this under the hood to cache things locally before making API calls home.

helios/inference_benchmarking/data_models.py
Hgherzog: makes sense. Maybe just use constants for these strings then?

PR #206, opened 2025-04-22
Miscellaneous dataset creation changes

favyen2:
- Add NAIP full config for full coverage of 10 m/pixel tiles.
- Improve documentation in various places.
- Add code to copy files quickly.

### DIFF
--- a/helios/data/constants.py
+++ b/helios/data/constants.py
@@ -94,7 +94,16 @@ class TimeSpan(str, Enum):
 
 @dataclass(frozen=True)
 class ModalitySpec:
-    """Modality specification."""
+    """Modality specification.
+
+    Args:
+        name: the name of the modality.
+        tile_resolution_factor: the factor of how much more ground area is covered by the tile compared with a tile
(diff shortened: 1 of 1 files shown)


helios/data/constants.py
```diff
@@ -175,6 +175,14 @@ class Modality:
         ignore_when_parsing=False,
     )
 
+    NAIP_10 = ModalitySpec(
```
Hgherzog, line 189: I would add a comment about what NAIP_10 means, my understanding it is that the data covers the 10m 256 by 256 grid (ie is a 4096 by 4096 grid) not that it is NAIP resampled to 10m.

helios/data/constants.py
```diff
@@ -175,6 +175,14 @@ class Modality:
...
+        name="naip_10",
+        tile_resolution_factor=16,
```
Hgherzog, line 191: Could we add a doc string to the Modality spec to define tile_resolution_factor. As it not exactly related to the resolution we are storing the data at but rather how much area we are covering compared to a 256 by 256 tile at base resolution.

PR #207, opened 2025-04-22
Henryh/fix growin process memory

Hgherzog:
Fixes issue where process memory grows over time when using torch 2.6, goes away in torch 2.7
- Run where process memory grows https://wandb.ai/eai-ai2/2025_04_18_galileo_contrastive/runs/8se3fiy5?nw=nwuserhenryhzog 
- Same run with torch 2.7 https://wandb.ai/eai-ai2/2025_04_18_galileo_contrastive/runs/ks4yzgdf?nw=nwuserhenryhzog&panelDisplayName=Process+Memory+In+Use+%28MB%29&panelSectionName=System 

Small Patches
- fixes small memory build up from not detaching losses before logging 
- allows cache dir to be None for dataset 

Runs on saturn and jupiter with this new setup https://wandb.ai/eai-ai2/helios-debug?nw=cd5dligrb8r

### DIFF
--- a/helios/data/dataset.py
+++ b/helios/data/dataset.py
@@ -809,6 +809,8 @@ class HeliosDatasetConfig(Config):
         self.validate()
         kwargs = self.as_dict(exclude_none=True, recurse=False)
         kwargs["h5py_dir"] = self.h5py_dir_upath
-        kwargs["cache_dir"] = self.cache_dir_upath if self.cache_dir else None
+        kwargs["cache_dir"] = (
+            self.cache_dir_upath if self.cache_dir is not None else None
+        )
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -15,7 +15,7 @@ pre-commit
 pytest
 rasterio
 rioxarray
-torch>=2.5
+torch>=2.6 # We use 2.7 for training on everything now
```
favyen2, line 18: should this set to >=2.7?

Hgherzog: 2.7 hasn't been officially released yet, I will bump it to 2.7 when it officially releases tomorrow

Hgherzog: the docker image is built with torch 2.7

PR #210, opened 2025-04-23
Updated implementation of model that alternates between spatial and temporal attention

favyen2:


helios/nn/st_model.py
```diff
@@ -0,0 +1,1269 @@
...
+                    input=tokens,
+                    other=exited_tokens,
+                )
+
+            # On even blocks, do temporal attention.
+            # On odd blocks, do spatial attention.
+            if i_blk % 2 == 0:
+                attention_mode = AttentionMode.TEMPORAL
+            else:
+                attention_mode = AttentionMode.SPATIAL
```
gabrieltseng, line 790: Another interesting experiment is to alternate between temporal, spatial and full attentions.

favyen2: moved to issue to explore later

PR #211, opened 2025-04-23
Breizhcrops

gabrieltseng:
https://wandb.ai/eai-ai2/helios-debug/runs/yh647hpk?nw=nwusergabrieltseng

Results are not comparable with the Galileo paper (since those are linear probing and these are KNN) but this will still be informative.

### DIFF
--- a/helios/evals/datasets/breizhcrops.py
+++ b/helios/evals/datasets/breizhcrops.py
@@ -162,6 +162,7 @@ class BreizhCropsDataset(Dataset):
 
     @staticmethod
     def _average_over_month(x: np.ndarray) -> np.ndarray:
+        # doa == date of acquisition
         x[:, SELECTED_BANDS[LEVEL].index("doa")] = np.array(
             [
                 t.month - 1


scripts/base_debug_scripts/galileo.py
```diff
@@ -220,6 +220,15 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
             norm_stats_from_pretrained=True,
             eval_interval=Duration.epochs(5),
         ),
+        "breizhcrops": DownstreamTaskConfig(
```
gabrieltseng, line 223: This change is repeated in `latent_mim.py` and `mae.py` - do we want an `EVAL_TASKS` list stored somewhere to avoid this?

yawenzzzz: yes, we can have sth like `scripts/eval_tasks_config.py` and import from there directly.

helios/evals/datasets/breizhcrops.py
```diff
@@ -0,0 +1,184 @@
...
+from einops import repeat
+from torch.utils.data import ConcatDataset, Dataset
+from upath import UPath
+
+from helios.train.masking import HeliosSample, MaskedHeliosSample, Modality
+
+from .constants import EVAL_S2_BAND_NAMES, EVAL_TO_HELIOS_S2_BANDS
+from .normalize import normalize_bands
+
+LEVEL = "L1C"
```
yawenzzzz, line 19: ah, then we will see the potential impact of L2A -> L1C

helios/evals/datasets/breizhcrops.py
```diff
@@ -0,0 +1,184 @@
...
+        masked_sample = MaskedHeliosSample.from_heliossample(
+            HeliosSample(
+                sentinel1=torch.tensor(image).float(), timestamps=timestamp.long()
+            )
+        )
+        return masked_sample, y_true.long()
+
+    @staticmethod
+    def _average_over_month(x: np.ndarray) -> np.ndarray:
+        x[:, SELECTED_BANDS[LEVEL].index("doa")] = np.array(
```
yawenzzzz, line 166: what is `doa`?

gabrieltseng: date of acquisition - I will add a comment noting this

scripts/base_debug_scripts/galileo.py
gabrieltseng: I don't love importing things from the scripts folder - I'll keep thinking about this

PR #212, opened 2025-04-23
Henryh/enable compression

Hgherzog:
Enable creating compressed h5 datasets with different options.

### DIFF
--- a/helios/dataset/convert_to_h5py.py
+++ b/helios/dataset/convert_to_h5py.py
@@ -99,9 +99,9 @@ class ConvertToH5py:
 
         Use for folder naming.
         """
-        compression_str = "_"
+        compression_str = ""
         if self.compression is not None:
-            compression_str = self.compression
+            compression_str = f"_{self.compression}"
         if self.compression_opts is not None:
             compression_str += f"_{self.compression_opts}"
         if self.shuffle is not None:


helios/dataset/convert_to_h5py.py
```diff
@@ -234,14 +282,25 @@ def set_h5py_dir(self, num_samples: int) -> None:
             logger.warning("h5py_dir is already set, ignoring new value")
             return
 
-        self.h5py_dir = (
-            self.tile_path
-            / self.h5py_folder
-            / "_".join(
-                sorted([modality.name for modality in self.supported_modalities])
+        if self.compression_settings_str:
```
favyen2: Suggestion: if you change `compression_settings_str` to `compression_settings_suffix` that starts with "_" when compression is set, then you can avoid the redundancy here and always have it as `f"{self.h5py_folder}{self.compression_settings_suffix}"`.

helios/dataset/convert_to_h5py.py
```diff
@@ -94,12 +94,12 @@ def __init__(
...
 
     @property
-    def compression_settings_str(self) -> str:
+    def compression_settings_suffix(self) -> str:
         """String representation of the compression settings.
 
         Use for folder naming.
         """
-        compression_str = ""
+        compression_str = "_"
```
favyen2: I think this should be empty string if compression is not set, otherwise same as before but with an extra "_" prefix.

Hgherzog: thanks

PR #213, opened 2025-04-25
Add SICKLE dataset

yawenzzzz:
Test run: https://beaker.org/ex/01JSSNN73BP1TKRBDTDPYFVVKM, https://wandb.ai/eai-ai2/helios-debug/runs/6wevswr3

### DIFF
--- a/helios/internal/common.py
+++ b/helios/internal/common.py
@@ -92,7 +92,7 @@ def build_launch_config(
         workspace=workspace,
         clusters=clusters,
         weka_buckets=weka_buckets,
-        beaker_image=f"petew/{OLMoCoreBeakerImage.stable_cu126}",  # we can all use the same image for now trying petes to see if it works or we need a copy in our workspace
+        beaker_image=f"petew/{OLMoCoreBeakerImage.stable_cu128}",  # we can all use the same image for now trying petes to see if it works or we need a copy in our workspace
         num_nodes=1,
         num_gpus=1,
         shared_memory="256GiB",


scripts/base_debug_scripts/latent_mim.py
```diff
@@ -230,6 +230,24 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
...
+        ),
+        "sickle-r": DownstreamTaskConfig(
+            dataset="sickle-r",
+            batch_size=8,
+            num_workers=2,
+            pooling_type=PoolingType.MEAN,
+            norm_stats_from_pretrained=True,
+            probe_lr=0.1,
+            eval_interval=Duration.epochs(5),
+        ),
```
yawenzzzz: Reduce this to 20 epochs, also add into galileo script

helios/evals/datasets/configs.py
```diff
@@ -69,5 +69,19 @@ class EvalDatasetConfig(NamedTuple):
...
+        imputes=[],
+        num_classes=2,
+        is_multilabel=False,
+        height_width=32,
+    ),
+    "sickle-r": EvalDatasetConfig(
+        task_type=TaskType.SEGMENTATION,
+        imputes=[],
+        num_classes=2,
+        is_multilabel=True,
```
gabrieltseng: If sickle-r is just the multimodal version of sickle, then should the `is_multilabel` attribute be consistent between both of these ?

helios/evals/datasets/constants.py
```diff
@@ -18,11 +18,40 @@
     "12 - SWIR",
 ]
 
+EVAL_S2_L2A_BAND_NAMES = [
```
gabrieltseng: do we need a separate list for L2A? or can we just use a single S2 list?

helios/evals/datasets/constants.py
```diff
@@ -31,18 +60,41 @@ def _eval_s2_band_index_from_helios_name(helios_name: str) -> int:
     raise ValueError(f"Unmatched band name {helios_name}")
 
 
+def _eval_s2_l2a_band_index_from_helios_name(helios_name: str) -> int:
```
gabrieltseng: Do we need this function? It's functionality is identical to the one above

helios/evals/datasets/sickle_dataset.py
```diff
@@ -0,0 +1,626 @@
...
+
+logger = logging.getLogger(__name__)
+
+CSV_PATH = UPath(
+    "/weka/dfive-default/helios/evaluation/SICKLE/sickle_dataset_tabular.csv"
+)
+DATA_DIR = UPath("/weka/dfive-default/helios/evaluation/SICKLE")
+SICKLE_DIR = UPath("/weka/dfive-default/presto_eval_sets/sickle")
+
+MONTH_TO_INT = {
```
gabrieltseng, line 45: nit: instead of defining month to int this way, and then commenting below that its zero indexed (L604) could we just 0-index this dict (and perhaps have that column here?)

helios/evals/datasets/sickle_dataset.py
```diff
@@ -0,0 +1,626 @@
...
+
+class SICKLEDataset(Dataset):
+    """SICKLE dataset class."""
+
+    def __init__(
+        self,
+        path_to_splits: Path = SICKLE_DIR,
+        split: str = "train",
+        partition: str = "default",
+        is_multimodal: bool = True,
```
gabrieltseng: instead of being boolean, could this be a list of input modalities, allowing (up to) S2, S1, L8? 

This seems to reflect what was done in the SICKLE paper (Table 3)

helios/evals/datasets/sickle_dataset.py
```diff
@@ -0,0 +1,626 @@
...
+    "aug": 8,
+    "sep": 9,
+    "oct": 10,
+    "nov": 11,
+    "dec": 12,
+}
+
+# Minimum number of months of data required for each modality
+# The growing season is usually 5 or 6 months
+MIN_MONTHS = 5
```
gabrieltseng, line 63: This is treated more like a max months, rightt? Since we fix the sampling to 5 months in L375 onwards? 

What is the motivation for doing this? (as opposed to just taking all the months of data). Is it to have consistent tensor sizes?

scripts/base_debug_scripts/latent_mim.py
yawenzzzz: Fixed.

helios/evals/datasets/configs.py
yawenzzzz: Ah, yes, it should be `False`, just fixed it. sickle - just Landsat, sickle-r - Landsat + S1 + S2.

helios/evals/datasets/constants.py
yawenzzzz: Yes, because the original SICKLE S2 images are all L2A, with only 12 bands, and `_eval_s2_band_index_from_helios_name` will extract the corresponding indices of Helios bands in the S2 image, we need a separate band list to get the correct band indices.

helios/evals/datasets/constants.py
yawenzzzz: I modified the functions so that we can input either the S2 L1C or L2A bands, and get the band indices via the same function.

helios/evals/datasets/sickle_dataset.py
yawenzzzz: I don't think we should subtract 1 in `MONTH_TO_INT`, this is mainly used to get `start_date` and `end_date` (L268-L269), and it should be the actual month so that we can find the satellite images within this time range and get their monthly aggregations.

helios/evals/datasets/sickle_dataset.py
yawenzzzz: Yes, this is mainly because months are either 5 or 6, and there's no less than 5 actually. So I make them all into 5 to have consistent tensor sizes. I will add a TODO here, once we have the missing timestep implementation, we can make them as 6, and marked the missing month.

helios/evals/datasets/sickle_dataset.py
yawenzzzz: I think this is a good idea, I removed the boolean, and added input modalities, which allows us to specify the modalities to use for PASTIS and SICKLE, it's only for multimodal tasks.

PR #219, opened 2025-04-30
Add remaining S2 geobench datasets

gabrieltseng:
These datasets are already on WEKA so we should definitely be using them


scripts/base_debug_scripts/galileo.py
```diff
@@ -220,6 +220,30 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
             norm_stats_from_pretrained=True,
             eval_interval=Duration.epochs(5),
         ),
+        "m-bigearthnet": DownstreamTaskConfig(
+            dataset="m-bigearthnet",
+            batch_size=128,
+            num_workers=8,
+            pooling_type=PoolingType.MEAN,
+            norm_stats_from_pretrained=True,
+            eval_interval=Duration.epochs(1),
```
gabrieltseng: note - just for testing. Change when merging

Issue #216, opened 2025-04-29
Move slow evaluations out of training loop and run as separate jobs


yawenzzzz:
Some of our in-loop evaluations, especially the multitemporal ones like PASTIS, are taking a long time to run. We can either reduce their evaluation frequency or move them out of the training loop and run them as separate jobs.

```
Finished m-eurosat evaluations in 151.3 seconds.
Finished mados evaluations in 91.7 seconds.
Finished sen1floods11 evaluations in 112.9 seconds.
Finished pastis evaluations in 820.1 seconds.
Finished pastis-r evaluations in 1247.8 seconds.
```

PR #220, opened 2025-05-01
Add GeoBench segmentation tasks

gabrieltseng:


requirements.txt
```diff
@@ -1,5 +1,5 @@
 ai2-olmo-core==2.1.0
-beaker-py>=1.32.0
+beaker-py==1.34.1
```
gabrieltseng: beaker-py 2.0.1 fails with the following error from olmo-core (in this [run](https://beaker.allen.ai/orgs/ai2/workspaces/earth-systems/work/01JT3R7TGHHPQ5052EDT4HCQCX/logs?jobId=01JT4R5A0H6BVSE20ENEHZ0WSM)):

`ImportError: cannot import name 'BeakerError' from 'beaker'`

We'll need to update both together. For now I pinned it to what seems to be in the docker image by default.

helios/internal/common.py
```diff
@@ -127,7 +127,9 @@ def build_launch_config(
             "conda shell.bash activate base",
             # Install torch==2.7 if we're targetting titan
             "pip install -e '.[all]'",
-            "pip install --upgrade beaker-py",
+            # commenting out the pip install beaker-py line, since
+            # it is incomaptible with olmo-core
+            # "pip install --upgrade beaker-py",
```
gabrieltseng: Highlighting this change, related to https://github.com/allenai/helios/pull/220/files#r2069933223

scripts/base_debug_scripts/galileo.py
```diff
@@ -242,7 +242,23 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
...
+            norm_stats_from_pretrained=True,
+            eval_interval=Duration.epochs(1),
+        ),
+        "m-cashew-plant": DownstreamTaskConfig(
+            dataset="m-cashew-plant",
+            batch_size=128,
+            num_workers=8,
+            pooling_type=PoolingType.MEAN,
+            norm_stats_from_pretrained=True,
+            eval_interval=Duration.epochs(1),
```
gabrieltseng: just for testing, to be changed

gabrieltseng: These are very slow and probably shouldn't be included in the training loop until we close https://github.com/allenai/helios/issues/216

yawenzzzz: yea, I'm working on #216 to support launching evaluations with checkpoints.

helios/internal/common.py
yawenzzzz: This gets fixed also in #223.

scripts/base_debug_scripts/galileo.py
gabrieltseng: okay - I will remove them from this script for now.

PR #222, opened 2025-05-01
Add projection before the pooling and optionally make it nonlinear

gabrieltseng:
Following our conversation on Monday's meeting. 

3 runs:

1. Project, then aggregate (although this should be equivalent to the existing run since this is just a linear projection): https://wandb.ai/eai-ai2/helios-debug/runs/3c1ab3wn
2. Aggregate, then project for 2 layers: https://wandb.ai/eai-ai2/helios-debug/runs/fr2cb4f4
3. Project, then aggregate for 2 layers: https://wandb.ai/eai-ai2/helios-debug/runs/i0gpk5sk

### DIFF
--- a/helios/nn/flexihelios.py
+++ b/helios/nn/flexihelios.py
@@ -10,7 +10,7 @@ import torch
 from einops import rearrange, repeat
 from olmo_core.config import Config
 from torch import Tensor, nn
-from torch.distributed.fsdp import fully_shard, register_fsdp_forward_method
+from torch.distributed.fsdp import fully_shard
 
 from helios.data.constants import Modality, ModalitySpec
 from helios.dataset.utils import get_modality_specs_from_names
@@ -1243,10 +1243,10 @@ class Encoder(FlexiHeliosBase):
     def apply_fsdp(self, **fsdp_kwargs: Any) -> None:
         """Apply FSDP to the model."""
         super().apply_fsdp(**fsdp_kwargs)
-        fully_shard(self.patch_embeddings, **fsdp_kwargs)
-        register_fsdp_forward_method(self.patch_embeddings, "forward")
-        fully_shard(self.project_and_aggregate, **fsdp_kwargs)
-        register_fsdp_forward_method(self.project_and_aggregate, "forward")
+        # fully_shard(self.patch_embeddings, **fsdp_kwargs)
+        # register_fsdp_forward_method(self.patch_embeddings, "forward")
+        # fully_shard(self.project_and_aggregate, **fsdp_kwargs)
+        # register_fsdp_forward_method(self.project_and_aggregate, "forward")
         fully_shard(self, **fsdp_kwargs)
 
 


helios/nn/flexihelios.py
```diff
@@ -1170,13 +1237,16 @@ def forward(
                 input_res=input_res,
                 token_exit_cfg=token_exit_cfg,
             )
-        return TokensAndMasks(**patchified_tokens_and_masks)
+        output = TokensAndMasks(**patchified_tokens_and_masks)
+        return output, self.project_and_aggregate(output)
```
gabrieltseng, line 1241: This PR moves the "project and aggregate" step into the encoder. This means the output of the encoder's forward function now returns the `TokensAndMasks` and the aggregated tensor. Happy to move it back to how it was if that is preferable.

helios/nn/flexihelios.py
```diff
@@ -212,6 +216,58 @@ def pool_unmasked_tokens(
...
+                x_modality = getattr(x, modality)
+                # Are these normalizations masked correctly?
+                x_modality = self.projection(x_modality)
+                masked_modality_name = x.get_masked_modality_name(modality)
+                decoder_emedded_dict[modality] = x_modality
+                decoder_emedded_dict[masked_modality_name] = getattr(
+                    x, masked_modality_name
+                )
+                x_projected = TokensAndMasks(**decoder_emedded_dict)
+                return x_projected.pool_unmasked_tokens(
```
Hgherzog: right now this is only projecting the first modality and leaving the other modalities untouched. Lines 265 and 266 should be unindented.

helios/nn/flexihelios.py
```diff
@@ -1170,13 +1237,16 @@ def forward(
...
-        return TokensAndMasks(**patchified_tokens_and_masks)
+        output = TokensAndMasks(**patchified_tokens_and_masks)
+        return output, self.project_and_aggregate(output)
 
     def apply_fsdp(self, **fsdp_kwargs: Any) -> None:
         """Apply FSDP to the model."""
         super().apply_fsdp(**fsdp_kwargs)
         fully_shard(self.patch_embeddings, **fsdp_kwargs)
         register_fsdp_forward_method(self.patch_embeddings, "forward")
+        fully_shard(self.project_and_aggregate, **fsdp_kwargs)
```
Hgherzog: This is a tiny layer, so I don't think sharding it separately is needed unless you profiled it and saw performance improvement for some reason. Calling fully shard will basically add an extra bit of communication without significant memory savings

Hgherzog: We likely should not  shard self.patch_embeddings either fwiw

helios/nn/flexihelios.py
gabrieltseng: 🤦 thanks for catching this. Fixed

helios/nn/flexihelios.py
gabrieltseng: okay. I have removed both

PR #223, opened 2025-05-01
Don't upgrade beaker-py

yawenzzzz:


helios/internal/common.py
```diff
@@ -127,7 +127,7 @@ def build_launch_config(
             "conda shell.bash activate base",
             # Install torch==2.7 if we're targetting titan
             "pip install -e '.[all]'",
-            "pip install --upgrade beaker-py",
```
favyen2, line 130: I think a comment here would be useful. Or remove the code entirely and rely on requirements.txt.

Do we need something in requirements.txt to ensure the newer version is not installed? This is because of some compatibility issue right?

yawenzzzz: OK, I added a comment for this. Also, update the requirements.txt to make sure the tests are good as well.

PR #226, opened 2025-05-07
Add Latlons as supported modalities

gabrieltseng:
https://wandb.ai/eai-ai2/helios-debug/runs/nw5ryjsg

### DIFF
--- a/helios/data/constants.py
+++ b/helios/data/constants.py
@@ -103,6 +103,7 @@ class ModalitySpec:
         band_sets: the band sets of the modality, ie the units of tokenization.
         is_multitemporal: whether the modality is multitemporal.
         ignore_when_parsing: whether to ignore the modality when parsing the data form the csv file.
+        image_tile_size_factor: the factor of how much bigger the dimensions of the image tile are compared with the base tile size.
     """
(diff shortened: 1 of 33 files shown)


scripts/base_debug_scripts/galileo.py
```diff
@@ -184,7 +184,7 @@ def build_dataloader_config(common: CommonComponents) -> HeliosDataLoaderConfig:
 
 def build_dataset_config(common: CommonComponents) -> HeliosDatasetConfig:
     """Build the dataset config for an experiment."""
-    h5py_dir = "/weka/dfive-default/helios/dataset/presto/h5py_data/landsat_naip_openstreetmap_raster_sentinel1_sentinel2_l2a_srtm_worldcover/118861"
```
favyen2: Should this be replicated across the base_debug_scripts for consistency?

gabrieltseng: I believe (?) these will change in the next bit as we update the datasets (and e.g. add NAIP). I've made an issue to make sure we do get to this: https://github.com/allenai/helios/issues/235

PR #227, opened 2025-05-07
Henryh/create 4k naip dataset

Hgherzog:
-adapt h5 and sample creation to allow 4k by 4k Naip imagery 
-introduce a tile_size_scale_factor attribute to modalities and grid tiles to indicate the expected dimensions of that modality (should allow us to do expected shape for filling missing modality and things like this) 

DOES NOT enable training on naip_10

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -130,6 +130,7 @@ def prepare_samples_and_supported_modalities() -> (
                             ): data_path / "s2_l2a_20m.tif",
                             BandSet(["B01", "B09"], 64): data_path / "s2_l2a_40m.tif",
                         },
+                        modality=Modality.SENTINEL2_L2A,
                     ),
                     Modality.SENTINEL1: ModalityTile(
                         grid_tile=GridTile(
@@ -140,6 +141,7 @@ def prepare_samples_and_supported_modalities() -> (
                         band_sets={
                             BandSet(["VV", "VH"], 16): data_path / "s1_10m.tif",
                         },
+                        modality=Modality.SENTINEL1,
                     ),
                     Modality.WORLDCOVER: ModalityTile(
                         grid_tile=GridTile(
@@ -148,6 +150,7 @@ def prepare_samples_and_supported_modalities() -> (
                         images=images,
                         center_time=datetime(2020, 6, 30),
                         band_sets={BandSet(["B1"], 16): data_path / "worldcover.tif"},
+                        modality=Modality.WORLDCOVER,
                     ),
                     Modality.OPENSTREETMAP_RASTER: ModalityTile(
                         grid_tile=GridTile(
@@ -192,6 +195,7 @@ def prepare_samples_and_supported_modalities() -> (
                                 4,
                             ): data_path / "openstreetmap.tif",
                         },
+                        modality=Modality.OPENSTREETMAP_RASTER,
                     ),
                 },
             )
--- a/tests/unit/dataset/test_sample.py
+++ b/tests/unit/dataset/test_sample.py
@@ -56,6 +56,7 @@ def create_image_tiles(tmp_path: Path) -> Callable:
                         ): data_path / "s2_20m.tif",
                         BandSet(["B01", "B09", "B10"], 64): data_path / "s2_40m.tif",
                     },
+                    modality=Modality.SENTINEL2,
                 ),
             ]
         }
@@ -70,6 +71,7 @@ def create_image_tiles(tmp_path: Path) -> Callable:
                     band_sets={
                         BandSet(["VV", "VH"], 16): data_path / "s1_10m.tif",
                     },
+                    modality=Modality.SENTINEL1,
                 ),
             ]
         }


helios/data/constants.py
```diff
@@ -192,6 +194,7 @@ class Modality:
         band_sets=[BandSet(["R", "G", "B", "IR"], 1)],
         is_multitemporal=False,
         ignore_when_parsing=False,
+        image_tile_size_factor=16,
```
gabrieltseng, line 197: I think this can be calculated using the existing explicit attributes.

We already have the resolution_factor of the band set, and the tile_resolution_factor of the modality spec.

So the tile size factor is ModalitySpec.tile_resolution_factor / BandSet.resolution_factor

Hgherzog: Not exactly. Some modalities have different resolutions for different bandsets and the image tile size should be the factor for the dimensions things are resampled to for a HeliosSample. It could be a property though instead that is ModalitySpec.tile_resolution_factor(min( BandSet.resolution_factor for band_set in bandsets))

helios/data/constants.py
```diff
@@ -98,7 +98,7 @@ class ModalitySpec:
 
     Args:
         name: the name of the modality.
-        tile_resolution_factor: the factor of how much more ground area is covered by the tile compared with a tile
+        tile_resolution_factor: the factor of how many more pixels are in the tile for the same ground area compared with a tile
```
favyen2: I think the original wording is correct -- a tile with `tile_resolution_factor=1` covers IMAGE_TILE_SIZE * BASE_RESOLUTION x IMAGE_TILE_SIZE * BASE_RESOLUTION meters, while a tile with `tile_resolution_factor=4` covers IMAGE_TILE_SIZE * BASE_RESOLUTION * 4 x IMAGE_TILE_SIZE  *BASE_RESOLUTION * 4 meters (4x the physical area). The number of pixels is not being controlled here, it is just controlling the ground area.

helios/data/constants.py
```diff
@@ -129,6 +129,16 @@ def bandsets_as_indices(self) -> list[list[int]]:
             offset += num_bands
         return indices
 
+    @property
+    def image_tile_size_factor(self) -> int:
```
favyen2: Deriving this automatically from the band sets may have unintended consequences, like resulting in openstreetmap_raster being input at 1024x1024 instead of the current 256x256 which I think has been working fine. Suggest to make it a configurable attribute instead, which defaults to 1 but can be set higher for NAIP_10.

helios/dataset/parse.py
```diff
@@ -53,6 +53,9 @@ class GridTile:
     col: int
     row: int
 
+    # The factor of how much bigger the dimensions of the image tile are compared with the base tile size.
+    image_tile_size_factor: int = 1
```
favyen2: I think this is a bit unintuitive since GridTile just specifies a projection, resolution, and tile within that projection/resolution. GridTile is not intended to be modality-specific, but this change makes it somewhat modality-specific.

I would suggest accessing the `image_tile_size_factor` directly from `load_image_for_sample` without changes to other functions/classes, by seeing how to access the ModalitySpec. If it's not possible, then this field would make more sense in ModalityTile since that is the modality-specific one.

helios/dataset/sample.py
```diff
@@ -112,15 +112,22 @@ def image_tiles_to_samples(
...
     # case we add it as both year and two-week, since currently all data at the base
     # resolution is static. (The intention here is to avoid adding a two-week tile
     # based on WorldCover being available if Sentinel-2 and others are only available
     # for one-year, but to still add NAIP or Maxar tiles.)
     unique_image_tiles: set[tuple[GridTile, TimeSpan]] = set()
-    for _, grid_tile, time_span in image_tile_index.keys():
+    for modality, grid_tile, time_span in image_tile_index.keys():
         if time_span == TimeSpan.STATIC:
-            if grid_tile.resolution_factor > 1:
+            # Now that we have NAIP_10, we want to include it in both datasets always
```
favyen2: The changes to `image_tiles_to_samples` do not seem necessary. The exception for static modalities at resolution_factor=1 (0.625 m/pixel) was added because there are no multitemporal modalities at that resolution, and so when discovering which tiles we should use, for 0.625 m/pixel we do need to use the static modalities. But at coarser resolutions, we can just use the multitemporal modalities. The original reason to do this is in case some tiles only contain one-year data or only contain two-week data. I think this is explained in the comment in the code above.

helios/dataset/sample.py
```diff
@@ -147,7 +154,12 @@ def image_tiles_to_samples(
                 continue
             # We only use modalities that are at an equal or coarser resolution.
             if modality.tile_resolution_factor < sample.grid_tile.resolution_factor:
+                logger.info(
+                    f"ignoring modality {modality.name} with resolution factor {modality.tile_resolution_factor} because it is coarser than the sample grid tile resolution factor {sample.grid_tile.resolution_factor}"
+                )
                 continue
+
+            # Why are we downscaling here?
```
favyen2: Some modalities like ERA5 may be stored at a coarser resolution in the dataset. This code is performing a lookup to discover modalities at equal or courser resolution, so for the ones at courser resolution it needs to adjust the lookup parameters in `modality_grid_tile` accordingly.

helios/dataset/sample.py
```diff
@@ -112,15 +112,22 @@ def image_tiles_to_samples(
...
-    for _, grid_tile, time_span in image_tile_index.keys():
+    for modality, grid_tile, time_span in image_tile_index.keys():
         if time_span == TimeSpan.STATIC:
-            if grid_tile.resolution_factor > 1:
+            # Now that we have NAIP_10, we want to include it in both datasets always
+            if (
+                grid_tile.resolution_factor > 1
+                and modality.name != Modality.NAIP_10.name
+            ):
+                logger.info(
```
favyen2: There may be millions of tiles captured here, suggest to make this debug.

helios/dataset/sample.py
```diff
@@ -235,10 +253,17 @@ def load_image_for_sample(
...
                 logger.debug(f"reading window={rasterio_window} from {fname}")
                 image: npt.NDArray = raster.read(window=rasterio_window)
+                logger.debug(f"image.shape={image.shape}")
 
                 # And then for now resample it to the grid resolution.
                 # The difference in resolution should always be a power of 2.
-                desired_subtile_size = IMAGE_TILE_SIZE // factor
+                # If the factor is less than 1 we want the desired size to be multiplied by the thing
+                # If the tile size is greater we want to keep that extent
+                desired_subtile_size = int(
```
favyen2, line 259: I think this is the only change within the `helios.dataset` module that is really needed. It would be better to avoid adding the GridTile.image_tile_size_factor and other changes. Here you can get the image_tile_size_factor either from ModalityTile (by adding a field there), or find a way to access the ModalitySpec (either by adding a field to ModalityTile or by adding it as an argument to this `load_image_for_sample` function) directly.

helios/dataset/parse.py
Hgherzog: The reason I did this was because `load_image_for_sample` does not know what modality is dealing with as the modality is not passed through there. I think it would probably make the most sense to store the modality_spec within the modality tile as well so that any modality specific information can be easily accessed for a modality tile.

Hgherzog: Though the resolution factor of the grid tile is already modality specific a it is used in `image_tiles_to_sample` because the resolution factor is based on the modality and the downscale factor is also specific to the modality.

favyen2: I think it makes sense to pass the ModalitySpec to `image_tiles_to_sample` in addition to or in lieu of the ModalityTile, since in SampleInformation the `modalities` field is already `dict[ModalitySpec, ModalityTile]` (and the SampleInformation is already being passed to `image_tiles_to_sample`).

I agree the grid is somewhat modality-specific since different modalities may be stored on a 0.625 m/pixel based grid, 10 m/pixel based grid, etc. but I think it makes sense to have GridTile related to that grid instead of even more modality-specific.

favyen2: So it would involve:
* Change `image = self.load_sample(sample_modality, sample)` in `convert_to_h5py.py` to pass `modality` instead of `sample_modality`.
* Change `load_sample` in `convert_to_h5py.py` to match the new API for `load_image_to_sample`, i.e. Callable[[ModalitySpec, SampleInformation], npt.NDArray].

favyen2: but yeah I think storing the ModalitySpec in ModalityTile makes sense too. it's just that it's already keyed in the SampleInformation modalities dict from ModalitySpec -> ModalityTile.

helios/dataset/sample.py
Hgherzog: I went with adding the ModalitySpec to the ModalityTile

helios/dataset/sample.py
Hgherzog: When I changed added the property to the grid tiles, this addition was neccesary but are not needed anymore after removing the property from the grid tiles.

PR #231, opened 2025-05-08
NAIP

gabrieltseng:
Opening a PR to leave comments

### DIFF
--- a/helios/dataset/sample.py
+++ b/helios/dataset/sample.py
@@ -112,7 +112,7 @@ def image_tiles_to_samples(
 
     # Enumerate all the (grid_tile, time_span) tuples present in the dataset.
     # Each of these identifies a training example.
-    # We ignore static time span here, unless it is at the base resolution or if it is NAIP_10, in which
+    # We ignore static time span here, unless it is at the base resolution, in which
     # case we add it as both year and two-week, since currently all data at the base
     # resolution is static. (The intention here is to avoid adding a two-week tile
     # based on WorldCover being available if Sentinel-2 and others are only available
@@ -120,13 +120,10 @@ def image_tiles_to_samples(
     unique_image_tiles: set[tuple[GridTile, TimeSpan]] = set()
     for modality, grid_tile, time_span in image_tile_index.keys():
         if time_span == TimeSpan.STATIC:
-            # Now that we have NAIP_10, we want to include it in both datasets always
-            if (
-                grid_tile.resolution_factor > 1
-                and modality.name != Modality.NAIP_10.name
-            ):
-                logger.info(
-                    f"ignoring static tile {grid_tile.resolution_factor} because it is coarser than the base resolution for modality {modality.name}"
+            if grid_tile.resolution_factor > 1:
+                logger.debug(
+                    f"ignoring static tile {grid_tile.resolution_factor} "
+                    f"because it is coarser than the base resolution for modality {modality.name}"
                 )
                 continue
             else:
@@ -154,12 +151,13 @@ def image_tiles_to_samples(
                 continue
             # We only use modalities that are at an equal or coarser resolution.
             if modality.tile_resolution_factor < sample.grid_tile.resolution_factor:
-                logger.info(
-                    f"ignoring modality {modality.name} with resolution factor {modality.tile_resolution_factor} because it is coarser than the sample grid tile resolution factor {sample.grid_tile.resolution_factor}"
+                logger.debug(
+                    f"ignoring modality {modality.name} with resolution factor "
+                    f"{modality.tile_resolution_factor} because it is coarser than "
+                    f"the sample grid tile resolution factor {sample.grid_tile.resolution_factor}"
                 )
                 continue
 
-            # Why are we downscaling here?
             downscale_factor = (
                 modality.tile_resolution_factor // sample.grid_tile.resolution_factor
             )
@@ -180,7 +178,6 @@ def image_tiles_to_samples(
                 resolution_factor=modality.tile_resolution_factor,
                 col=grid_tile.col // downscale_factor,
                 row=grid_tile.row // downscale_factor,
-                image_tile_size_factor=modality.image_tile_size_factor,
             )
 
             index_key = (modality, modality_grid_tile, lookup_time_span)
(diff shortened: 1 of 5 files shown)


helios/data/constants.py
```diff
@@ -129,6 +129,16 @@ def bandsets_as_indices(self) -> list[list[int]]:
             offset += num_bands
         return indices
 
+    @property
+    def image_tile_size_factor(self) -> int:
+        """Get the image tile size factor."""
+        min_bandset_resolution = min(
+            [band_set.resolution_factor for band_set in self.band_sets]
+        )
```
gabrieltseng, line 137: Possibly dumb question:

We don't actually support different resolutions within a `ModalitySpec` (since a modality spec has to be a datacube). What is the purpose of having different band set resolution factors within a modality spec? 

It seems like we resample to the min (just like this function assumes) anyway?

helios/data/dataset.py
```diff
@@ -745,6 +774,17 @@ def __getitem__(self, args: GetItemArgs) -> tuple[int, HeliosSample]:
...
+        for key in sample_dict.keys():
+            if key.endswith("_10"):
+                keys_to_replace.append(key)
+        for key in keys_to_replace:
+            val = sample_dict[key]
+            # add the key, without the "_10"
+            sample_dict[key[:-3]] = val
+            # remove the "_10"
+            del sample_dict[key]
+
```
gabrieltseng, line 787: This is definitely something we should remove before merging this PR. I have it here to get things running for now.

Otherwise, we get runs like [this](https://beaker.allen.ai/orgs/ai2/workspaces/earth-systems/work/01JTQSEDCY0DSYEMG0G3NBAFTH/logs?jobId=01JTQSEDH783SS31N322RXCQRC) which fail with 

```
TypeError: HeliosSample.__new__() got an unexpected keyword argument 'naip_10'
```

helios/data/constants.py
favyen2: The band set resolution configures what resolution those bands are stored at (at least before H5 conversion).

It may make sense to have a separate option to override the resolution that the band set or modality should be input to the model.

helios/dataset/sample.py
```diff
@@ -112,15 +112,22 @@ def image_tiles_to_samples(
 
     # Enumerate all the (grid_tile, time_span) tuples present in the dataset.
     # Each of these identifies a training example.
-    # We ignore static time span here, unless it is at the base resolution, in which
+    # We ignore static time span here, unless it is at the base resolution or if it is NAIP_10, in which
```
favyen2: The exception for 0.625 m/pixel tiled NAIP was because we don't have any multi-temporal modalities at 0.625 m/pixel, so we need to use the static modalities to identify which tiles we want to train at for 0.625 m/pixel.

At 10 m/pixel, we don't have this issue, so we shouldn't add this extra code.

helios/dataset/sample.py
```diff
@@ -168,10 +180,14 @@ def image_tiles_to_samples(
                 resolution_factor=modality.tile_resolution_factor,
                 col=grid_tile.col // downscale_factor,
                 row=grid_tile.row // downscale_factor,
+                image_tile_size_factor=modality.image_tile_size_factor,
```
favyen2: It seems strange to add a modality-specific field to the GridTile, which just identifies a projection, resolution, and tile. Should be added to ModalityTile instead, or just read from the ModalitySpec/BandSet directly without needing an extra field anywhere here.

I don't think this PR really needs to change parse.py at all, and sample.py except `image_tiles_to_samples`.

helios/dataset/sample.py
```diff
@@ -222,6 +239,7 @@ def load_image_for_sample(
                     raise ValueError(
                         f"expected tile to be square but width={raster.width} != height={raster.height}"
                     )
+                # Assuming all tiles cover the same area as the resolution factor 16 tile
```
favyen2: I don't understand this comment. The code below ensures that if the data is stored at a courser resolution (tiles covering a bigger area), it will read just the portion corresponding to the requested tile (`image_tile`).

helios/dataset/sample.py
```diff
@@ -235,10 +253,17 @@ def load_image_for_sample(
                 )
                 logger.debug(f"reading window={rasterio_window} from {fname}")
                 image: npt.NDArray = raster.read(window=rasterio_window)
+                logger.debug(f"image.shape={image.shape}")
 
                 # And then for now resample it to the grid resolution.
                 # The difference in resolution should always be a power of 2.
-                desired_subtile_size = IMAGE_TILE_SIZE // factor
+                # If the factor is less than 1 we want the desired size to be multiplied by the thing
```
favyen2: AFAICT this is the only change that is actually needed in `helios.dataset`.

helios/data/constants.py
```diff
@@ -98,7 +98,7 @@ class ModalitySpec:
 
     Args:
         name: the name of the modality.
-        tile_resolution_factor: the factor of how much more ground area is covered by the tile compared with a tile
+        tile_resolution_factor: the factor of how many more pixels are in the tile for the same ground area compared with a tile
```
favyen2, line 101: The previous comment seems to be more correct. Suggest to change this back perhaps with some rewording to avoid confusion. e.g. "the factor of how much more ground area is covered by the tile compared with a tile of IMAGE_TILE_SIZE x IMAGE_TILE_SIZE pixels at the base resolution. For example, tiles at a tile_resolution_factor=4 contain 4x4 base tiles."

helios/data/dataset.py
favyen2: It would make sense to add a naip_10 that is separate from naip to the sample.

An alternative is to remove the current `naip` and rename `naip_10 -> naip`, but would be best NOT to do this since that affects how the dataset is created and would cause problems.

helios/data/constants.py
favyen2: I think the image_tile_size_factor should be a property that we set (and defaults to 1, where 1 means the image should be IMAGE_TILE_SIZE; might need some changes to how it's used to achieve that) rather than computed from the band set resolution. Otherwise this may have unintended consequences like switching openstreetmap_raster to be 1024x1024 input instead of the current 256x256.

Hgherzog: Having it as a seperate property that defaulted to 1 was the original implementation, I didn't realize it would improperly affect open streetmap. This change is still open in a PR that hasnt been merged so I can change it there

PR #232, opened 2025-05-08
Naip

pjreddie:
New branch for NAIP support after git snafu

### DIFF
--- a/scripts/base_debug_scripts/galileo.py
+++ b/scripts/base_debug_scripts/galileo.py
@@ -132,7 +132,7 @@ def build_train_module_config(
             f"All modalities must be in token_exit_cfg_a: {common.training_modalities}"
         )
     token_exit_cfg_b = {modality: 0 for modality in common.training_modalities}
-    WARMUP_EPOCHS = 10
+    WARMUP_EPOCHS = 1
     dp_config = DataParallelConfig(name=DataParallelType.ddp)
 
     # TODO: would need a scheduler config and registry to be able to change this with overrides


helios/data/dataset.py
```diff
@@ -745,6 +775,19 @@ def __getitem__(self, args: GetItemArgs) -> tuple[int, HeliosSample]:
...
+            val = sample_dict[key]
+            # add the key, without the "_10"
+            sample_dict[key[:-3]] = val
+            # remove the "_10"
+            del sample_dict[key]
+        """
+
         # Fill any training modalities that are not present in the h5 file with missing values
         sample, missing_modalities = self.fill_sample_with_missing_values(sample_dict)
         subset_sample = self.apply_subset(sample, args)
```
gabrieltseng, line 838: So I think this is not helping performance - we are adding a ~ 4k x 4k image for every instance, even though NAIP is present only in a small subset of instances. Doing something more clever here is probably desirable.

This may be quite a major change and may be preferable as a separate PR though.

gabrieltseng: or maybe not. Attempted in https://github.com/allenai/helios/pull/232/commits/c089d9b3072506e754fe8db8d124d658a7a67dad

gabrieltseng: hmm this doesn't seem to be speeding things up much (green is with the change, yellow is before and orange is without NAIP). Its possible what we also need to do is cluster the NAIP instances so that they are in a batch - right now, if each batch has ~1 NAIP instance then this doesn't help at all.

Screenshot 2025-05-09 at 9 16 03 PM

Hgherzog: Is the slow down coming from data loading  or from the model forward?

pjreddie: it's a little of both, data loading is definitely a lot of it, model forward is also taking about twice as long maybe as non-naip model

helios/nn/flexihelios.py
```diff
@@ -1514,7 +1530,7 @@ def forward(
         Returns:
             TokensAndMasks containing the predicted tokens and their masks
         """
-        decoder_emedded_dict = x._asdict()
```
gabrieltseng, line 1517: Why are we using `_asdict` here? This returns a dict with keys for the None values in the tuples, which means the intersection in `get_modalities_to_process` doesn't help (since its just filtering against the training modalities)

gabrieltseng: I ask to figure out: (1) if its safe to replace `_asdict` with `as_dict` (which can filter for Nones) and (2) if we should replace this usage elsewhere in flexihelios (`_asdict` gets called relatively often)

PR #239, opened 2025-05-14
Token merging after encoder

pjreddie:


helios/nn/flexihelios.py
```diff
@@ -1238,6 +1275,8 @@ def forward(
                 token_exit_cfg=token_exit_cfg,
             )
         output = TokensAndMasks(**patchified_tokens_and_masks)
+        if self.pool_tokens:
+            output = output.pool_tokens_across_modalities()
```
gabrieltseng, line 1300: (This assumed you wanted to reconstruct all the tokens, not the modality-averaged tokens. The comment may still be helpful if you do want to reconstruct all the masked tokens)

I think some changes will also be required in the `predictor` here. 

Currently, the predictor looks at `output` to determine which mask tokens should be added (and what should be reconstructed). Given the `output.pool_tokens_across_modalities`, it will think that there is only some S1 tokens to reconstruct.

I think what you'd want here is a new predictor which uses these pooled modalities as the values, and uses the (original) masks to determine the keys and queries. This should be okay since the predictor already uses cross attention.

PR #240, opened 2025-05-15
Henryh/fix eval issues

Hgherzog:
Add more visibility into the evaluation times. Make the speed monitor callback go first post step to more accurately capture time

### DIFF
--- a/helios/train/callbacks/evaluator_callback.py
+++ b/helios/train/callbacks/evaluator_callback.py
@@ -167,7 +167,7 @@ class DownstreamEvaluatorCallback(Callback):
                 evaluator.eval_interval
             )
             if self.step <= 1 or self.step % eval_interval_steps != 0:
-                return
+                continue
             logger.info(f"Running {evaluator.evaluation_name} evaluations...")
             start_time = time.monotonic()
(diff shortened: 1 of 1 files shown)


helios/train/callbacks/evaluator_callback.py
```diff
@@ -167,13 +167,17 @@ def post_step(self) -> None:
                 evaluator.eval_interval
             )
             if self.step <= 1 or self.step % eval_interval_steps != 0:
-                continue
+                return
             logger.info(f"Running {evaluator.evaluation_name} evaluations...")
```
yawenzzzz: We should not `return` because each evaluator may have different evaluation steps/epochs, and this will stop iterating over the remaining evaluators.

helios/train/callbacks/evaluator_callback.py
```diff
@@ -167,13 +167,17 @@ def post_step(self) -> None:
                 evaluator.eval_interval
             )
             if self.step <= 1 or self.step % eval_interval_steps != 0:
-                continue
```
Hgherzog: fix

PR #245, opened 2025-05-20
cross masking speedups

Hgherzog:


scripts/galileo_style_masking/galileo_base.py
```diff
@@ -195,7 +195,6 @@ def build_dataset_config(common: CommonComponents) -> HeliosDatasetConfig:
     return HeliosDatasetConfig(
         h5py_dir=h5py_dir,
         training_modalities=common.training_modalities,
-        use_samples_with_missing_supported_modalities=True,
```
pjreddie, line 198: why remove this? is it the default or does this not work with missing modalities? or were you just experimenting? just wondering

Hgherzog: It is no longer an arg for the dataset config

PR #246, opened 2025-05-20
Windowed attention

favyen2:
Add support for windowed attention.

### DIFF
--- a/tests/unit/nn/test_st_model.py
+++ b/tests/unit/nn/test_st_model.py
@@ -61,8 +61,8 @@ class TestSTBase:
         B, H, W, T, B_S, D = 2, 3, 3, 4, 2, 4
         sentinel2_l2a_tokens = torch.randn(B, H, W, T, B_S, D)
         sentinel2_l2a_mask = torch.randint(0, 2, (B, H, W, T, B_S)).float()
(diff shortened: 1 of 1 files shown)


helios/nn/st_model.py
```diff
@@ -551,6 +663,97 @@ def split_and_expand_per_modality_spatial(
...
+                    hs=window_size,
+                    ws=window_size,
+                    bs=b_s,
+                )
+                # Remove beginning padding.
+                modality_tokens = modality_tokens[:, offset_padding:, offset_padding:]
+                # Remove end padding.
+                x_modality = modality_tokens[:, 0:h, 0:w]
+
+            elif len(dims) == 6:
```
pjreddie: I _think_ we can assume it's always 6 now? or is this not true. I thought we were going to have every modality have a temporal dimension and it's just 1 if it's non-temporal

favyen2: Thanks, I got rid of all the code to handle 5 dimensions, in future PR I want to clean it up further by having a separate AttentionMode class for each mode (so WindowedAttentionMode would have combine/split function for windowed attention, SpatialAttentionMode for spatial-only, etc.).

PR #249, opened 2025-05-30
Debug contrastive nan

yawenzzzz:
Previously, I encountered Galileo + contrastive loss nan issue (https://beaker.allen.ai/orgs/ai2/workspaces/earth-systems/work/01JW9J3J0QP7ZG5V3FEQZ9AGZN?taskId=01JW9J3J1F1HBT8AF34902MMXM&jobId=01JW9M4RBZR9BZMB45K1ZC33V8), and it popped up mainly when I set `use_modalities_with_missing_timesteps = False`. 

What happened is: right now, when generating h5 samples, we require at least one multitemporal modality (S1, S2, or Landsat) but they can be less than 12 months. When `use_modalities_with_missing_timesteps = True` all the missing timesteps will be imputed, and if the longest time sequence is less than 12 months, we will pad at the end with `MISSING_VALUE`. However, when `use_modalities_with_missing_timesteps = False`, we will treat all modalities the same, by imputing entirely as `MISSING_VALUE`. This is the cause of hitting the contrastive nan issue, if the original multitemporal modality is less than 12 months, the whole sample will just be entirely `MISSING_VALUE` and there's nothing to encode and decode from that sample. Once we hit nothing to decode, then it will trigger the contrastive loss nan. 

I made three fixes here: 
(1) Remove `use_modalities_with_missing_timesteps` argument as we don't want to add entirely missing sample into the training, which provides no benefit in learning (we still want to keep samples that with some modalities being missing), and since the missing percentage is quite small, it's safe to just impute them.
(2) When subsetting, we always search from the valid temporal range instead of a random `start_t` this will increase the likelihood of getting the non-missing timesteps. 
(3) In the masking, instead of checking `batch.time` we now check `batch.valid_time` which finds the minimum number of the non-missing timesteps in a batch, so if there isn't sufficient non-missing timesteps, we will switch to space masking instead.

### DIFF
--- a/helios/data/dataset.py
+++ b/helios/data/dataset.py
@@ -231,10 +231,10 @@ class HeliosSample(NamedTuple):
         if self.timestamps is None:
             raise ValueError("Timestamps are not present in the sample")
         for i in range(self.timestamps.shape[0]):
-            unique_timesteps = torch.unique(self.timestamps[i])
+            unique_timesteps = torch.unique(self.timestamps[i], dim=0)
             min_valid_time = min(min_valid_time, unique_timesteps.shape[0])
         if min_valid_time < self.time:
-            logger.warning(
+            logger.debug(
                 f"valid_time is smaller than time: {min_valid_time} < {self.time}"
             )
         return min_valid_time
--- a/helios/train/masking.py
+++ b/helios/train/masking.py
@@ -646,6 +646,8 @@ class SpaceTimeMaskingStrategy(MaskingStrategy):
     ) -> MaskedHeliosSample:
         """Apply space or time masking to the input data."""
         has_enough_timesteps = batch.valid_time >= 3
+        if not has_enough_timesteps:
+            logger.debug(f"Valid time: {batch.valid_time}, Time: {batch.time}")
         if (self.generator.random() < 0.5) or (not has_enough_timesteps):
             logger.info("Applying space masking")
             return self.space_strategy.apply_mask(batch, patch_size, **kwargs)


helios/internal/common.py
```diff
@@ -167,11 +167,11 @@ def build_common_components(
...
-        Modality.LATLON.name,
-        Modality.SRTM.name,
-        # Modality.NAIP.name,
-        Modality.LANDSAT.name,
-        Modality.OPENSTREETMAP_RASTER.name,
+        # Modality.LATLON.name,
+        # Modality.SRTM.name,
+        # # Modality.NAIP.name,
+        # Modality.LANDSAT.name,
+        # Modality.OPENSTREETMAP_RASTER.name,
```
yawenzzzz, line 174: Change this back.

helios/data/dataset.py
```diff
@@ -223,6 +223,22 @@ def time(self) -> int:
             raise ValueError("Timestamps are not present in the sample")
         return self.timestamps.shape[-2]
 
+    @property
+    def valid_time(self) -> int:
```
Hgherzog, line 227: Suggestion: Add a comment that this assumes the timestamps are edge padded and add a test that combines pad timestamp and this method.

helios/data/dataset.py
```diff
@@ -304,8 +325,16 @@ def subset(
         sampled_hw = sampled_hw_p * patch_size
         start_h = np.random.choice(self.height - sampled_hw + 1)
         start_w = np.random.choice(self.width - sampled_hw + 1)
-        start_t = np.random.choice(self.time - max_t + 1)
+
+        # If current_length > max_t, we can start from a random valid timestep
```
Hgherzog, line 329: nit: add a quick explanation of why this is happening ie we want sample that have less than 12 timestamps that are padded with missing data not to have missing data selected. Currently, the comment just explains code that is already very readable

PR #250, opened 2025-06-02
Update cross space-time masking

yawenzzzz:
This mainly addresses issues with cross-space, time, or space-time masking. The root cause was that some samples had no tokens to encode, especially when all selected encoding bandsets were missing. The main change is to select encoding and decoding bandsets per sample instead of per batch, ensuring that missing bandsets are excluded.

Test run: https://wandb.ai/eai-ai2/helios-debug/runs/758926w3

### DIFF
--- a/tests/unit/train/test_masking.py
+++ b/tests/unit/train/test_masking.py
@@ -890,17 +890,41 @@ def test_space_cross_modality_masking_with_missing_data(set_random_seeds: None) 
         worldcover=torch.full((b, h, w, 1, worldcover_num_bands), MISSING_VALUE),
     )
 
-    strategy = ModalityCrossSpaceMaskingStrategy(
+    strategy_allow_false = ModalityCrossSpaceMaskingStrategy(
         encode_ratio=0.1,
         decode_ratio=0.75,
         allow_encoding_decoding_same_bandset=False,
     )
-    masked_sample = strategy.apply_mask(batch, patch_size=patch_size)
-    logger.info(f"masked_sample: {masked_sample}")
+    strategy_allow_true = ModalityCrossSpaceMaskingStrategy(
+        encode_ratio=0.1,
+        decode_ratio=0.75,
+        allow_encoding_decoding_same_bandset=True,
+    )
+    masked_sample_allow_false = strategy_allow_false.apply_mask(
+        batch, patch_size=patch_size
+    )
+    masked_sample_allow_true = strategy_allow_true.apply_mask(
+        batch, patch_size=patch_size
+    )
+    logger.info(f"masked_sample_allow_false: {masked_sample_allow_false}")
+    logger.info(f"masked_sample_allow_true: {masked_sample_allow_true}")
     # Check that the worldcover mask has the expected values
     # Check that latlon mask has the expected values
     expected_latlon_mask = torch.tensor([[0], [2], [2], [0]])
 
     # Assert that the masks match the expected values
-    assert (masked_sample.worldcover_mask == MaskValue.MISSING.value).all()  # type: ignore
-    assert torch.equal(masked_sample.latlon_mask, expected_latlon_mask)
+    assert (masked_sample_allow_false.worldcover_mask == MaskValue.MISSING.value).all()  # type: ignore
+    assert torch.equal(masked_sample_allow_false.latlon_mask, expected_latlon_mask)
+
+    # Compare the masks between two strategies
+    assert not torch.equal(
+        masked_sample_allow_false.latlon_mask, masked_sample_allow_true.latlon_mask
+    )
+    assert not torch.equal(
+        masked_sample_allow_false.sentinel2_l2a_mask,
+        masked_sample_allow_true.sentinel2_l2a_mask,
+    )
+    assert not torch.equal(
+        masked_sample_allow_false.sentinel1_mask,
+        masked_sample_allow_true.sentinel1_mask,
+    )


helios/data/dataset.py
```diff
@@ -225,11 +215,15 @@ def time(self) -> int:
 
     @property
     def valid_time(self) -> int:
-        """Get the minimum number of valid time steps in the data."""
-        # Go through each sample, check the number of valid timesteps and get the min
+        """Get the minimum number of valid time steps in a batch.
+
+        Note that the timestamps are edge padded by repeating the last timestamp.
+        """
         min_valid_time = 12
```
Hgherzog: min_valid_time should be a constant as it is tied to the max length of the sequence which could change in the future

helios/train/loss.py
```diff
@@ -572,6 +572,16 @@ def compute(
         # online_encodings_b = predictions.pool_unmasked_tokens(
         #     PoolingType.MEAN, spatial_pooling=False
         # )
+
+        valid_mask = (
+            torch.all(torch.isfinite(predictions), dim=-1)
```
Hgherzog: nit: if all elements are finite then automatically all of the elements are not nan so the .isnan logic is redundant

helios/train/masking.py
```diff
@@ -739,56 +739,80 @@ def __init__(
         self._encode_ratio = encode_ratio
         self._decode_ratio = decode_ratio
         self.strategy = strategy
-        self.max_unmasking_bandsets = max_unmasking_bandsets
         self.min_encoding_bandsets = min_encoding_bandsets
         self.max_encoding_bandsets = max_encoding_bandsets
+        self.generator = np.random.default_rng(0)
```
Hgherzog: Suggestion: So far we have been using global random state for random consistency in masking instead of generator. This allows us to restore masking state with the rng component of the state dict the trainer reloads on restarts.

helios/train/masking.py
```diff
@@ -739,56 +739,80 @@ def __init__(
...
-            raise ValueError("No valid decoding bandset combinations found")
-        return decoding_bandset_combinations
+    def get_sample_present_modalities(
+        self, batch: MaskedHeliosSample
+    ) -> list[dict[str, dict[int, int]]]:
+        """Get the modalities that are present for each sample."""
+        masked_sample_dict = batch.as_dict(return_none=False)
+        present_modalities = []
+
+        for sample_idx in range(batch.timestamps.shape[0]):
```
Hgherzog: How long is this operation taking in practice? Looping over modalities and then doing something like the below where you vectorize in the sample dimension might be helpful. If the time is negligible, then it is okay to leave it. 
`missing_values_mask = (modality_mask == MaskValue.MISSING.value) # Shape (B, *dims, NumBands)

    # For each sample, check if ALL values for that modality (across all its dimensions other than batch) are missing.
    # Flatten all dimensions except the batch dimension for the torch.all check.
    # missing_values_mask.view(batch_size, -1) will have shape (B, total_elements_per_sample_for_modality)
    is_modality_completely_missing_for_samples = torch.all(
        missing_values_mask.view(batch_size, -1), 
        dim=1
    )
    # is_modality_completely_missing_for_samples has shape (B,)
    
    is_modality_present_for_samples = ~is_modality_completely_missing_for_samples # Shape (B,)

    num_bandsets = modality_mask.shape[-1]
    # This bandset_info is the same for all samples where this modality is present
    bandset_info = {i: 1 for i in range(num_bandsets)}

    # Get indices of samples where the current modality is present
    present_sample_indices = torch.where(is_modality_present_for_samples)[0]

    for sample_idx_tensor in present_sample_indices:
        sample_idx = sample_idx_tensor.item() # Convert 0-dim tensor to Python int
        all_present_modalities[sample_idx][modality] = bandset_info`

helios/train/masking.py
```diff
@@ -739,56 +739,80 @@ def __init__(
...
+                )
+            # Split the bandsets into encoded and decoded
+            if len(all_modalities_bandsets) <= 2:
+                encoded_bandset_idxs = all_modalities_bandsets
+                decoded_bandset_idxs = []
+            else:
+                num_encoded_bandsets = self.generator.integers(
+                    2, len(all_modalities_bandsets)
+                )
+                encoded_idxs = self.generator.choice(
```
Hgherzog: Suggestion: see comment above about using generation vs np.random.intergers

helios/train/masking.py
```diff
@@ -739,56 +739,80 @@ def __init__(
...
+                    # Add all bandsets indices, by default encode all bandsets (1: encode, 0: decode)
+                    bandset_idx = {i: 1 for i in range(modality_mask.shape[-1])}
+                    present_modalities_for_sample[modality] = bandset_idx
+
+            present_modalities.append(present_modalities_for_sample)
+        return present_modalities
+
+    def select_encoded_decoded_bandsets(
+        self, present_modalities: list[dict[str, dict[int, int]]]
+    ) -> list[dict[str, dict[int, int]]]:
```
Hgherzog: see above suggestion on simplifying the creation of the all_modalities_bandsets

helios/train/masking.py
```diff
@@ -739,56 +739,80 @@ def __init__(
...
+                ]
+                decoded_bandset_idxs = [
+                    band
+                    for i, band in enumerate(all_modalities_bandsets)
+                    if i not in encoded_idxs
+                ]
+            # Set the encoded and decoded bandsets (1: encode, 0: decode)
+            for encoded_modality, encoded_bandset in encoded_bandset_idxs:
+                present_modalities_for_sample[encoded_modality][encoded_bandset] = 1
+            for decoded_modality, decoded_bandset in decoded_bandset_idxs:
```
Hgherzog: I think it is cleaner and simpler to just return two objects one for encoded and another for decoded instead of storing a 1 or 0 for every bandset individually? But maybe the access pattern suggests this is better

helios/train/masking.py
```diff
@@ -808,53 +831,66 @@ def apply_bandset_mask_rules(
...
-                        == MaskValue.ONLINE_ENCODER.value
-                    )
-                    modality_mask[..., bandset_idx] = torch.where(
-                        online_encoder_mask,
-                        MaskValue.TARGET_ENCODER_ONLY.value,
-                        modality_mask[..., bandset_idx],
-                    )
+                encoded_decoded_bandset_idxs = present_modalities_for_sample[modality]
+                for bandset_idx in encoded_decoded_bandset_idxs:
+                    is_encoded = encoded_decoded_bandset_idxs[bandset_idx] == 1
```
Hgherzog: If we do it this way better to just store true and false than check against 1 or 0

helios/train/masking.py
```diff
@@ -808,53 +831,66 @@ def apply_bandset_mask_rules(
...
-                    )
-                    modality_mask[..., bandset_idx] = torch.where(
-                        not_missing_mask,
-                        forced_mask_value,
-                        modality_mask[..., bandset_idx],
+
+            for sample_idx in range(masked_batch.timestamps.shape[0]):
+                present_modalities_for_sample = encoded_decoded_bandsets[sample_idx]
+
+                if modality not in present_modalities_for_sample:
```
Hgherzog: I think having a helper function here would help quite a bit with readability

helios/train/masking.py
```diff
@@ -739,56 +739,80 @@ def __init__(
...
+                modality_mask = masked_sample_dict[modality_mask_name]
+                missing_mask = modality_mask == MaskValue.MISSING.value
+                # Modality is present only if it is not completely missing
+                if not torch.all(missing_mask[sample_idx]):
+                    # Add all bandsets indices, by default encode all bandsets (1: encode, 0: decode)
+                    bandset_idx = {i: 1 for i in range(modality_mask.shape[-1])}
+                    present_modalities_for_sample[modality] = bandset_idx
+
+            present_modalities.append(present_modalities_for_sample)
+        return present_modalities
```
Hgherzog: Suggestion: I think creating and returning the all_modalities_bandsets here would help simplify the for loops in select_encoded_decoded_bandsets significantly. e.g return a list per sample of lists of the available (modality, bandset_idx).

helios/train/masking.py
```diff
@@ -739,56 +739,80 @@ def __init__(
...
+            for modality in present_modalities_for_sample:
+                all_modalities_bandsets.extend(
+                    [
+                        (modality, bandset_idx)
+                        for bandset_idx in present_modalities_for_sample[modality]
+                    ]
+                )
+            # Split the bandsets into encoded and decoded
+            if len(all_modalities_bandsets) <= 2:
+                encoded_bandset_idxs = all_modalities_bandsets
```
Hgherzog: If there is only 1 present bandset in the sample we should remove this sample? Does this ever happen in practice?

helios/train/masking.py
```diff
@@ -739,56 +739,80 @@ def __init__(
...
+                    if i not in encoded_idxs
+                ]
+            # Set the encoded and decoded bandsets (1: encode, 0: decode)
+            for encoded_modality, encoded_bandset in encoded_bandset_idxs:
+                present_modalities_for_sample[encoded_modality][encoded_bandset] = 1
+            for decoded_modality, decoded_bandset in decoded_bandset_idxs:
+                present_modalities_for_sample[decoded_modality][decoded_bandset] = 0
+
+            encoded_decoded_bandsets.append(present_modalities_for_sample)
+        return encoded_decoded_bandsets
```
Hgherzog, line 858: Suggestion: I think it is much simpler to just return the encoded and decoded bandset indices directly. nit:  It may also make sense to have them be sets rather than lists.

scripts/base_debug_scripts/galileo.py
Hgherzog, line 1: I think this should be a seperate base_debug script

helios/data/dataset.py
yawenzzzz: fixed.

helios/train/loss.py
yawenzzzz: I think we can safely remove this valid_mask since we're checking the present modalities per sample now. I will double check.

helios/train/masking.py
yawenzzzz: fixed.

helios/train/masking.py
yawenzzzz: fixed.

scripts/base_debug_scripts/galileo.py
yawenzzzz: this is just for testing purpose, i will get it back later.

helios/train/masking.py
yawenzzzz: Actually, it's possible to happen (but we haven't ran into this issue yet), e.g., when there's only S1 available, no other modalities. It's mainly impacting the contrastive loss, as this sample will have nothing to decode. I think dropping those samples at the very beginning, like in the dataset filter, will be a better idea.

helios/train/masking.py
yawenzzzz: Fixed.

helios/train/masking.py
yawenzzzz: i think this is a faster implementation, which saves time on revisiting the modality mask. I will switch to this then.

helios/train/masking.py
yawenzzzz: agree, i removed the present_modalities and instead simply return encoding and decoding bandsets, i think that's a better idea.

helios/train/masking.py
yawenzzzz: fixed.

helios/train/masking.py
yawenzzzz: agree, fixed.

helios/train/masking.py
yawenzzzz: fixed.

PR #252, opened 2025-06-03
Henryh/investigate s1 landsat performance

Hgherzog:
Made evaluation faster
- split out embedding batch size and probe batch size
- enabled flash attention during eval 
- enabled picking max miou for the eval

Other small fixes around using properties of modalities instead of str, Fix bug to allow for sickle without s2 but with s1 and landsat

### DIFF
--- a/helios/nn/flexihelios.py
+++ b/helios/nn/flexihelios.py
@@ -1195,7 +1195,9 @@ class Encoder(FlexiHeliosBase):
             # of True indicates the value *should* take part in
             # attention
             # WARNING: THIS MAY CHANGE DEPENDING ON THE ATTENTION IMPLEMENTATION
-            tokens = blk(x=tokens, y=None, attn_mask=new_mask)
+            tokens = blk(
+                x=tokens, y=None, attn_mask=new_mask if self.training else None
+            )
 
         if exit_ids_seq is not None:
             # this should only ever be called by the target encoder,


helios/train/masking.py
```diff
@@ -128,7 +132,11 @@ def height(self) -> int:
     @property
     def width(self) -> int:
         """Get the width of the data."""
-        height_width_time_modalities = ["sentinel2_l2a", "sentinel1", "worldcover"]
+        height_width_time_modalities = [
+            "sentinel2_l2a",
+            Modality.SENTINEL1.name,
+            "worldcover",
+        ]
```
gabrieltseng: Do we want to replace all the modality names in this list?

Hgherzog: I think the correct solution is to actually loop through modalities with spatial and temporal dimensions based on Modality spec properties.

helios/evals/linear_probe.py
```diff
@@ -56,53 +57,73 @@ def train_and_eval_probe(
...
+            device=device,
+        )
+        print(f"Epoch {end_epoch}, MIoU: {eval_miou}")
+        eval_mious.append(eval_miou)
+    for i in range(len(eval_mious)):
+        print(f"Epoch {(i + 1) * eval_interval}, MIoU: {eval_mious[i]}")
+    max_miou = max(eval_mious)
+    max_epoch = (eval_mious.index(max_miou) + 1) * eval_interval
+    print(f"Max MIoU: {max_miou} at epoch {max_epoch}")
+    return max(eval_mious)
```
gabrieltseng: IMO this is leakage - basically using the evaluation data to pick the best epoch. 

When reporting test results, we will use the same code (I think?), so we should be careful about this. 

I think the right solution here is to explicitly separate the validation and test sets - then, if we pick the best performing epoch using the val set, we are not peaking at test set results. FWIW, I think sweeping probe learning rates is similar to picking the best epoch, so for consistency with the Galileo results it may be best to only sweep the learning rates using a fixed number of epochs (50).

Hgherzog: I agree there may be leakage if you do this on the test set. I can update this to just warning log instead if the max eval miou is significantly higher at an early epoch. For all the evals except sickle this created a negligible difference in performance and the last epoch was usually the best anyway. Because sickle has so few labels, it overfits way before 50 epochs once the embeddings are pretty good so the goal of this was just to prevent overfitting which seems to be the main culprit of the really high variance looking sickle evals. I would honestly rather not use sickle if we can find a better landsat eval.

gabrieltseng: okay. It seems like sickle doesn't have a test set (cc @yawenzzzz ?) in which case for sickle our val set == test set, so it most problematic to do this for sickle.

Given this I think we should either split sickle's train set into a train and validation set, or not use sickle and find another landsat dataset as you suggested. Maybe we can do both, since splitting sickle's training data should be easy.

gabrieltseng: https://github.com/allenai/helios/pull/255

Here's an example of how we could have a fully untouched "test" set.

Hgherzog: Thanks I incorporated that into this.

yawenzzzz: yes, the authors didn't share test labels, I have just requested and still wait for response, we can add test set later. For now, splitting train into train & val looks good.

helios/evals/linear_probe.py
```diff
@@ -56,53 +60,79 @@ def train_and_eval_probe(
...
-        ),
-        probe=probe,
-        num_classes=config.num_classes,
-        patch_size=output_patch_size,
-        device=device,
-    )
-    return eval_miou
+    logits_per_patch = int(config.num_classes * output_patch_size * output_patch_size)
+    probe = nn.Sequential(nn.Linear(in_features, logits_per_patch)).to(device)
+    num_eval_groups = math.ceil(epochs / eval_interval)
```
yawenzzzz: Suggestion: add a comment to explain the usage of `num_eval_groups` as it haven't been used before.

PR #253, opened 2025-06-03
Pjreddie/naming fixes

pjreddie:
Encoder and Predictor had different names for what was semantically the same variable, this unifies the naming and also removes the config options from a lot of tests when they were set to the default anyway.

Some day I might advocate for changing Predictor to be Decoder but not today...

### DIFF
--- a/helios/nn/st_model.py
+++ b/helios/nn/st_model.py
@@ -46,13 +46,13 @@ class STBase(nn.Module):
         self,
         embedding_size: int,
         max_sequence_length: int,
-        use_channel_embs: bool,
         num_heads: int,
         mlp_ratio: float,
         depth: int,
         drop_path: float,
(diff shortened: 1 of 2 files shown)


helios/nn/flexihelios.py
```diff
@@ -593,8 +593,8 @@ def __init__(
         embedding_size: int,
         supported_modalities: list[ModalitySpec],
         max_sequence_length: int,
-        use_channel_embs: bool = True,
-        random_channel_embs: bool = False,
+        learnable_channel_embeddings: bool = True,
+        random_channel_embeddings: bool = False,
```
Hgherzog, line 597: Suggestion: name this random_init_channel_embeds so its clear it is the initialization not whether it is learned

PR #254, opened 2025-06-03
Henryh/128 datasets

Hgherzog:
Enables creation of datasets where each sample has a tile size split up from the base image tile size and also lets you do zstd via command line

### DIFF
--- a/helios/dataset/convert_to_h5py.py
+++ b/helios/dataset/convert_to_h5py.py
@@ -292,7 +292,7 @@ class ConvertToH5py:
                 image = convert_to_db(image)
 
             if modality.is_spatial:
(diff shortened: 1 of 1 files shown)


helios/dataset/convert_to_h5py.py
```diff
@@ -257,6 +290,17 @@ def _create_h5_file(
             # Convert Sentinel1 data to dB
             if modality == Modality.SENTINEL1:
                 image = convert_to_db(image)
+
+            if modality.is_spatial:
+                # Calculate row and column indices for 2x2 grid
```
favyen2: Fix comment to not be specific about 2x2 grid

PR #257, opened 2025-06-05
Linear probe for classification evals

gabrieltseng:


helios/train/callbacks/evaluator_callback.py
```diff
@@ -71,6 +75,34 @@ def __init__(
...
+        if eval_mode == "linear_probe":
+            if self.probe_lr is None:
+                raise ValueError("probe_lr cannot be none for segmentation tasks.")
+            if self.config.task_type == TaskType.SEGMENTATION:
+                if self.config.height_width is None:
+                    raise ValueError(
+                        "config.height_width cannot be none for segmentation tasks."
+                    )
+                if self.config.height_width % self.patch_size != 0:
+                    raise ValueError("Image height / width indivisable by patch size.")
```
gabrieltseng, line 100: Moved a lot of these checks to the init so that they will trigger earlier in the script if they are wrong.

PR #258, opened 2025-06-05
add token budget script

Hgherzog:
Screenshot 2025-06-05 at 11 08 02 AM

### DIFF
--- a/scripts/check_token_budget.py
+++ b/scripts/check_token_budget.py
@@ -1,5 +1,7 @@
 """Script to check the token budget for a given sample."""
 
+import argparse
+
 import pandas as pd
 import torch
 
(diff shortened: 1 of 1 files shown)


scripts/check_token_budget.py
```diff
@@ -0,0 +1,89 @@
...
+    df = pd.DataFrame(results, index=[f"{b}" for b in token_budgets])
+    df.index.name = "Token Budget"
+
+    print("This table shows the maximum number of time steps (t) possible for a given")
+    print("h_w_p (patch tokens in height/width) and a total token budget.\n")
+    print(f"Sample created with modalities: {modalities_to_test}")
+    print(f"Base dimensions: H={H}, W={W}, T={T}\n")
+    print(df)
+    # write to csv
+    df.to_csv("token_budget_results.csv")
```
favyen2, line 107: If you like you could make the modalities and max H / W / T customizable from command line along with optional output filename but it's just a script so I think it's also fine as is.

Hgherzog: sure

PR #259, opened 2025-06-06
Henryh/var seq len flash atten

Hgherzog:


helios/nn/utils.py
```diff
@@ -38,5 +57,5 @@ def apply_ddp(
             self,
             device_mesh=dp_mesh,
             bucket_cap_mb=100,
-            find_unused_parameters=find_unused_parameters,
+            find_unused_parameters=True,  # find_unused_parameters,
```
pjreddie: should this be hardcoded? doesn't this mean 2x forward passes or something?

PR #264, opened 2025-06-10
Using worldcover and OSM as supervision only

gabrieltseng:
Example accuracies:
Screenshot 2025-06-17 at 6 21 56 PM

Todos:
- ✅ Weighting for the loss (e.g. for the contrastive loss a weight of 0.1 worked well)
- ✅ Make the MLP vs. Linear layer easily configurable 
- Try combined pastis - is it better at combining s1 and s2?
- :x: Does this prevent mode collapse? Can we try full-depth encodings (or EMA) with this? (no performance is still bad with EMA)


helios/train/train_module/supervised_latent_mim.py
```diff
@@ -0,0 +1,414 @@
...
+        max_grad_norm: float | None = None,
+        scheduler: Scheduler | None = None,
+        device: torch.device | None = None,
+        state_dict_save_opts: dist_cp_sd.StateDictOptions | None = None,
+        state_dict_load_opts: dist_cp_sd.StateDictOptions | None = None,
+        ema_decay: tuple[float, float] = (0.996, 1.0),
+        warmup_duration: Duration = Duration.epochs(2),
+        regularizer_config: LossConfig | None = None,
+        supervisory_modalities: list[str] = [Modality.WORLDCOVER.name],
+        compute_accuracies: bool = False,
```
gabrieltseng, line 148: Computing the accuracies provides a very nice sanity check but it *really* slows the model down. Brown is without computing the accuracies, blue is with computing the accuracies.

Screenshot 2025-06-17 at 6 09 03 PM

helios/nn/flexihelios.py
```diff
@@ -1070,6 +1132,29 @@ def __init__(
...
+                else:
+                    multiplier = 1
+                for idx, band_set in enumerate(modality.band_sets):
+                    probes[f"{modality_name}_{idx}"] = nn.Sequential(
+                        nn.Linear(self.embedding_size, probe_dim),
+                        nn.GELU(),
+                        nn.Linear(
+                            probe_dim,
+                            output_hw * len(band_set.bands) * multiplier,
+                        ),
```
gabrieltseng: Should this be an MLP? or a linear layer? 🤷 probably configurable

helios/nn/flexihelios.py
```diff
@@ -1070,6 +1133,40 @@ def __init__(
...
+            # TODO - right now this assumes a resolution factor of 1. Needs to be fixed to handle (e.g.) NAIP
+            output_hw = self.max_patch_size**2
+            for modality_name in probe_modalities:
+                modality = Modality.get(modality_name)
+                if modality_name == Modality.WORLDCOVER.name:
+                    multiplier = NUM_WORLDCOVER_CLASSES
+                else:
+                    multiplier = 1
+                for idx, band_set in enumerate(modality.band_sets):
+                    if len(probe_dims) == 0:
```
favyen2: I think this special case is not needed -- as far as I can tell, the code below handles empty probe_dims correctly.

helios/train/train_module/supervised_latent_mim.py
```diff
@@ -0,0 +1,425 @@
...
+        argmax_pred = pred.argmax(dim=-1)
+        matches = argmax_pred == targets
+        return sum(matches) / len(matches)
+
+    @classmethod
+    def supervisory_losses(
+        cls,
+        supervisory_modalities: dict[str, torch.Tensor],
+        probe_outputs: dict[str, torch.Tensor],
+        loss: torch.Tensor,
```
favyen2: It feels a bit odd to me to pass in the other loss here and then add them within this function. At least for me it feels more natural to return just the supervisory loss here and add them up outside this function. I guess the same for total_batch_sup and total_batch_acc, the latter would still need a compute_acc bool to be passed.

helios/train/train_module/supervised_latent_mim.py
```diff
@@ -0,0 +1,425 @@
...
+                            * supervisory_weight
+                        )
+                        if total_batch_acc is not None:
+                            batch_acc = get_local_tensor(
+                                cls.accuracy_score(
+                                    filtered_targets,
+                                    filtered_modality_bandset,
+                                )
+                            )
+                            total_batch_acc[f"{modality}_{idx}"] += batch_acc
```
favyen2: Does something need to be detached here? You have `modality_loss.detach()` but I'm not sure if the accuracy is detached. Maybe that could cause slowdown that you mentioned too. I think it would be clearer to do the summation (along with get_local_tensor / detach) in the loop function instead of here.

helios/train/train_module/supervised_latent_mim.py
```diff
@@ -0,0 +1,425 @@
...
+                modality_bandset = modality_tensor[:, :, :, 0, bands]
+                if len(bands) > 1:
+                    # then we need to turn it into indices
+                    modality_bandset = torch.argmax(
+                        modality_bandset, dim=-1, keepdim=True
+                    )
+                probe_output = probe_outputs[
+                    f"{modality}_{idx}"
+                ]  # B, H, W, T, Bandsets or 11 if its worldcover
+                if probe_output.shape[-3] != modality_bandset.shape[-2]:
```
favyen2, line 259: Is this in case patch_size < max patch size? It may be helpful to add a comment here.

helios/train/train_module/supervised_latent_mim.py
```diff
@@ -0,0 +1,425 @@
...
+                modality_bandset = modality_bandset.long()
+                flat_modality_bandset = modality_bandset.flatten().to(
+                    probe_output.device
+                )
+                target_mask = torch.logical_and(
+                    spatial_mask.flatten(), (flat_modality_bandset != MISSING_VALUE)
+                )
+                filtered_modality_bandset = flat_modality_bandset[target_mask]
+                filtered_targets = probe_output.flatten(end_dim=-2)[target_mask, :]
+                if len(filtered_modality_bandset) > 0:
```
favyen2: It may be more clear to rewrite with this pattern:
```
for idx, bands in enumerate(modality_spec.bandsets_as_indices()):
    ...
    if len(filtered_modality_bandset) == 0:
        logger.info(f"All values missing for {modality}")
        continue

    modality_loss = loss_fn(filtered_targets, filtered_modality_bandset)
    ...
```

helios/train/train_module/supervised_latent_mim.py
```diff
@@ -0,0 +1,425 @@
...
+                    modality_loss = loss_fn(
+                        filtered_targets,
+                        filtered_modality_bandset,
+                    )
+                    if torch.isnan(modality_loss).any():
```
favyen2: Same as comment above, it is typically more clear to have code like this:
```
if torch.isnan(modality_loss).any():
    logger.warning(f"NaN in unsupervised loss for {modality}")
    continue

loss += supervisory_weight * modality_loss
...
```
This way the if check is treated as a special case or easy case rather than a branching condition which leads to more indentation for the common/hard case.

favyen2 (approved):
Left a few minor comments but the overall structure looks good to me.

Eventually we may want to support different loss functions for the probes other than cross entropy, but it makes sense to start with just adding cross entropy. Future modifications may move the loss and conversion-into-indices to a configurable class similar to the loss classes.

helios/nn/flexihelios.py
gabrieltseng: updated!

helios/train/train_module/supervised_latent_mim.py
gabrieltseng: Added - this may indeed be the slowdown

helios/train/train_module/supervised_latent_mim.py
gabrieltseng: Added

helios/train/train_module/supervised_latent_mim.py
gabrieltseng: updated - thanks

helios/train/train_module/supervised_latent_mim.py
gabrieltseng: updated - thanks

tests/integration/train/train_module/test_latent_mim_mae.py
```diff
@@ -212,6 +212,6 @@ def test_train_batch_with_missing_modalities(
         logger.info(mock_trainer._metrics)
         assert torch.allclose(
             mock_trainer._metrics["train/PatchDisc+MAE"],
-            torch.tensor(2.9),
+            torch.tensor(3.3),
```
gabrieltseng, line 215: I am going to assume this is due to changes in the model initialization; I am having trouble seeing where else it may be coming from.

helios/train/train_module/supervised_latent_mim.py
gabrieltseng: updated

PR #268, opened 2025-06-13
Henryh/cross masking experiments

Hgherzog:


tests/unit/train/test_masking.py
```diff
@@ -906,11 +906,9 @@ def test_space_cross_modality_masking_with_missing_data(set_random_seeds: None)
     masked_sample_allow_true = strategy_allow_true.apply_mask(
         batch, patch_size=patch_size
     )
-    logger.info(f"masked_sample_allow_false: {masked_sample_allow_false}")
-    logger.info(f"masked_sample_allow_true: {masked_sample_allow_true}")
     # Check that the worldcover mask has the expected values
     # Check that latlon mask has the expected values
-    expected_latlon_mask = torch.tensor([[0], [2], [2], [0]])
+    expected_latlon_mask = torch.tensor([[2], [2], [0], [0]])
```
Hgherzog: This value changed because there are addditional random calls in the function now so even if the initial state is fixed this changes

helios/train/masking.py
```diff
@@ -815,22 +847,49 @@ def select_encoded_decoded_bandsets(
...
+                ]
+                num_encodable_modalities = len(encodable_modalities)
+                max_encoded_bandsets = (
+                    min(self.max_encoded_bandsets, num_encodable_modalities)
+                    if self.max_encoded_bandsets is not None
+                    else num_encodable_modalities
+                )
+
+                min_encoded_bandsets = (
+                    self.min_encoded_bandsets or num_encodable_modalities
```
favyen2: Suggest to rewrite as `self.min_encoded_bandsets if self.min_encoded_bandsets is not None else num_encodable_modalities` for clarity.

helios/train/masking.py
```diff
@@ -815,22 +847,49 @@ def select_encoded_decoded_bandsets(
...
+                )
+
+                # if min and max are none we will encode all encodable bandsets
```
favyen2: I think this comment makes more sense above the definition for `max_encoded_bandsets`.

helios/train/masking.py
favyen2: Should computing `min_encoded_bandsets` use `min(self.min_encoded_bandsets, num_encodable_modalities)` (with the min)? You do use that for computing `min_decoded_bandsets` but not here.

helios/train/masking.py
```diff
@@ -844,14 +903,29 @@ def select_encoded_decoded_bandsets(
...
+                    num_decoded_bandsets = len(decoded_bandset_idxs)
+                    min_decoded_bandsets = min(
+                        min_decoded_bandsets, num_decoded_bandsets
+                    )
+                    max_decoded_bandsets = min(
+                        max_decoded_bandsets, num_decoded_bandsets
+                    )
+                    # select the decoded bandsets
+                    decoded_idxs = np.random.choice(
+                        len(present_modalities_bandsets_for_sample),
```
Hgherzog: change to decoded_bandset_idxs

helios/train/masking.py
```diff
@@ -868,6 +942,7 @@ def apply_bandset_mask_rules(
         encoded_decoded_bandsets: list[
             tuple[set[tuple[str, int]], set[tuple[str, int]]]
         ],
+        present_modalities_bandsets: list[list[tuple[str, int]]],
     ) -> MaskedHeliosSample:
         """Allow encoding of encoded bandsets and decoding of decoded bandsets."""
```
favyen2: For the docstring, I think something like "Compute masks for each band set based on the encode and decode selections" would be more clear. Currently it is not clear that this function is computing the masks. After the first line of docstring I also suggest saying that this is intended to be called after `select_encoded_decoded_bandsets`, or if you list the args you can say that the encoded_decoded_bandsets generally comes from that other function.

helios/train/masking.py
```diff
@@ -1126,6 +1233,42 @@ def apply_mask(
...
+            decode_ratio=decode_ratio,
+            allow_encoding_decoding_same_bandset=allow_encoding_decoding_same_bandset,
+            min_encoded_bandsets=min_encoded_bandsets,
+            max_encoded_bandsets=max_encoded_bandsets,
+            min_decoded_bandsets=min_decoded_bandsets,
+            max_decoded_bandsets=max_decoded_bandsets,
+            only_decode_modalities=only_decode_modalities,
+        )
+
+    def overide_random_mask_condition(self, modality_spec: ModalitySpec) -> bool:
```
favyen2: If the intention is for this function to always be overridden, it may be better for the default implementation to `raise NotImplementedError`. Otherwise, it is unnecessary to override here.

Also, I'm having trouble wrapping my head around `overide_random_mask_condition`. It could benefit from a more descriptive docstring plus comments when you check this in `apply_bandset_mask_rules` about exactly where it differs from the regular logic. It seems like, without this enabled, it is still setting the mask to target encoder if it was decoder but it shouldn't be decoded, or if it was online_encoder but it shouldn't be encoded.

Is the difference that, if the condition is enabled for a modality, then if a patch was marked decoder but the band set shouldn't be decoded and should be encoded, then it will mark the patch online_encoder instead of target_encoder (as would happen if the condition was disabled); and vice versa?

PR #271, opened 2025-06-17
Token fusing via attention

favyen2:
This supports fusing tokens into one on the last layer of the encoder. Currently decoder sees the fused token repeated across the unmasked input tokens (instead of just one token).

I also added an option to directly configure the attention mode of each layer, since I was testing different alternatives there. There are lots of options now for STEncoder/STPredictor and it may make sense to consolidate them in the future, but perhaps we keep them for now so the previous launch scripts still work.

### DIFF
--- a/helios/nn/st_model.py
+++ b/helios/nn/st_model.py
@@ -182,7 +182,7 @@ class STBase(nn.Module):
         w: int | None = None
         for modality in modalities_to_process:
             x_modality = x[modality]
-            if len(x_modality.shape) not in [5, 6]:
+            if len(x_modality.shape) != 6:
                 continue
             cur_h = x_modality.shape[1]
             cur_w = x_modality.shape[2]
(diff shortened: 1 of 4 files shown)


helios/nn/st_model.py
```diff
@@ -940,22 +994,71 @@ def apply_attn(
...
+                # This is to keep consistent with the expected output format.
+                tokens = tokens.expand(-1, attention_seq_len, -1)
+            else:
+                tokens = blk(x=tokens, y=None, attn_mask=new_mask)
 
             # Apply normalization on last block.
             if i_blk == len(self.blocks) - 1:
                 tokens = self.norm(tokens)
 
+            if do_token_fusing and not self.fuse_using_cross_attn:
```
pjreddie, line 1054: does this like work well at all? is it just to have an else case here or is it an experiment you wanted to try?

favyen2: I implemented it with self attention first and then realized it wasn't too complicated to implement with cross attention so added that option.

They seem to perform about the same -- self-attention was higher on eval but I think it could be random variability:
https://wandb.ai/eai-ai2/v0.2_sweep?nw=7dn5ug4avbb

Maybe can keep for now until there are more experiments/debugging to validate what's going on, maybe the cross attention isn't really implemented properly or maybe it would work well if the decoder just saw one token instead of same one repeated.

PR #275, opened 2025-06-20
Experiments with NAIP at 2.5 m/pixel

favyen2:
This has #232 merged in. The naip_10 modality is changed to 2.5 m/pixel rather than creating a new modality, that doesn't seem good, in general I think we should make it easier to add new "modalities" especially for ones not used in eval.

Besides NAIP stuff, I added FixedModalityMaskingStrategy which allows using one masking strategy while also marking a list of modalities as decode-only.

### DIFF
--- a/helios/train/masking.py
+++ b/helios/train/masking.py
@@ -419,6 +419,22 @@ class SpaceMaskingStrategy(MaskingStrategy):
         patch_size_at_16: int,
         device: torch.device | None = None,
     ) -> torch.Tensor:
+        """Create a h_p x w_p spatial mask.
+
+        Here, h_p and w_p are the number of patches along height and width dimension
+        respectively.
+
+        The mask computed here is modality-agnostic, but we still expect a specific
+        modality to be passed since it will be used to compute h_p/w_p. The mask will
+        then need to be resized using _resize_spatial_mask_for_modality to the
+        modality's patch size.
+
+        Args:
+            modality: the modality we are using to compute h_p/w_p.
+            shape: the shape of the image for that modality.
+            patch_size_at_16: the patch size measured in 10 m/pixel pixels.
+            device: the device to use.
+        """
         if not modality.is_spatial:
             raise ValueError("Non-spatial modality {modality}")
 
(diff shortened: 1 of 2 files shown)


helios/data/constants.py
```diff
@@ -198,7 +198,7 @@ class Modality:
         band_sets=[BandSet(["R", "G", "B", "IR"], 1)],
         is_multitemporal=False,
         ignore_when_parsing=False,
-        image_tile_size_factor=16,
+        image_tile_size_factor=4,
```
yawenzzzz, line 204: Suggestion: add a comment about NAIP_10 spatial resolution.

helios/data/dataloader.py
```diff
@@ -292,6 +293,9 @@ def get_mock_batch(self) -> HeliosSample:
         if Modality.SENTINEL2_L2A.name in self.dataset.training_modalities:
             mock_sentinel2_l2a = rng.random((256, 256, 12, 12), dtype=np.float32)
             output_dict["sentinel2_l2a"] = mock_sentinel2_l2a
+        if Modality.NAIP_10.name in self.dataset.training_modalities:
+            mock_naip_10 = rng.random((4096, 4096, 1, 4), dtype=np.float32)
```
yawenzzzz: Suggestion: using (1024, 1024, 1, 4) is more consistent with the training samples.

helios/data/dataset.py
```diff
@@ -336,43 +341,94 @@ def subset(
...
+        else:
+            value = np.full(
+                self.get_expected_shape(modality),
+                fill_value=MISSING_VALUE,
+                dtype=self.dtype,
+            )
+        if modality == "latlon":
+            return value
+        else:
+            return value[..., :min_timestamps, :]
```
yawenzzzz: The `fill_sample_with_missing_values` should already handle this case, if any modality is missing in that sample, the that modality will be imputed with `MISSING_VALUE`.

helios/train/masking.py
```diff
@@ -491,7 +450,30 @@ def _create_spatial_mask(
...
+        patch_size = patch_size_at_16 * modality.image_tile_size_factor
+        assert (h % patch_size == 0) and (w % patch_size == 0)
+        h_p = h // patch_size
+        w_p = w // patch_size
+
+        if (patch_mask.shape[1] != h_p) or (patch_mask.shape[2] != w_p):
+            raise ValueError(
+                f"Mismached shapes for {modality.name}: "
+                f"got patch_mask {patch_mask.shape} for h_p {h_p}, w_p {w_p}"
+            )
```
yawenzzzz: Suggestion: It seems better to have the original `_create_spatial_mask` which includes the conversion from patch-level to pixel-level mask. The `_create_spatial_mask` still needs to recompute the `patch_size` and also `h_p`, `w_p`, which are already available in `_create_patch_spatial_mask`.

scripts/v0.2_sweep/latent_mim_128_naip_moredata.py
```diff
@@ -0,0 +1,284 @@
...
+        optim_config=AdamWConfig(lr=0.0001, weight_decay=0.02),
+        warmup_duration=Duration.steps(8000),
+        rank_microbatch_size=64,  # Can be 256 on titan, needs to be <= 64 (i think) on jupiter
+        masking_config=MaskingConfig(
+            strategy_config={
+                "type": "space_time",
+                "encode_ratio": 0.1,
+                "decode_ratio": 0.75,
+            }
+        ),
```
yawenzzzz: Could you also update this to the best config? We can use the script with best configs to run all evals.

helios/data/dataset.py
favyen2: @pjreddie do you think I can revert the changes in collate_helios now that we handle missing values elsewhere? I'm not too familiar with this portion of the code.

pjreddie: yes it seems like _fill_missing_modality deals with the issue we were running in to and we can revert this

favyen2: I think there may have been an attempt to optimize things for batches that contain zero samples with NAIP_10 so that it doesn't need to deal with that modality. But I think the conclusion from before was that, with batch size 512, it'll be very rare to have a batch with no NAIP_10 whatsoever. So probably fill_sample_with_missing_values is sufficient. I'll update it.

helios/train/masking.py
pjreddie: we used to use the same spatial mask for each modality but now patch_size depends on image_tile_size_factor on a per modality basis so it's not as simple to reuse. It's possible we could rearchitect this some but I think the basic logic of the 2 functions is correct

favyen2: I refactored the code a bit to make the purpose of the functions more clear (added docstring and renamed _create_spatial_mask -> _resize_spatial_mask_for_modality), I also moved the check for matching shapes out to the calling function since it didn't make too much sense to pass shape into _resize_spatial_mask_for_modality just to check that the resulting mask would have matching shape. @yawenzzzz can you take another look?

yawenzzzz: Great, it looks better now!

PR #278, opened 2025-06-24
Add ERA5

yawenzzzz:
presto: `/weka/dfive-default/helios/dataset/presto/160_era5_monthly/`
osm_sampling: `/weka/dfive-default/helios/dataset/osm_sampling/160_era5_monthly/`


data/rslearn_dataset_configs/config_era5.json
```diff
@@ -4,13 +4,11 @@
...
-            "snow-depth",
-            "soil-temperature-level-1",
+            "2m-temperature",
+            "2m-dewpoint-temperature",
+            "surface-pressure",
             "10m-u-component-of-wind",
             "10m-v-component-of-wind",
-            "2m-temperature",
-            "skin-temperature",
             "total-precipitation"
```
gabrieltseng, line 12: How were these decided ? LFMC + ag requirements?

yawenzzzz: Yes, primarily based on LFMC and agricultural applications. For crop yield forecasting, we previously used temperature, precipitation, wind, pressure, and humidity (which can be derived from temperature and dew point temperature).

PR #280, opened 2025-06-25
Ryanp/update debug scripts

ryspark:
Small updates to scripts in `base_debug_scripts`
* Add beaker callback to display progress bar in UI
* Add support for 128x128 datasets
* Fix bugs in MAE debug script

### DIFF
--- a/helios/internal/common.py
+++ b/helios/internal/common.py
@@ -22,7 +22,7 @@ from helios.internal.experiment import (
 )
 
 logger = logging.getLogger(__name__)
-BUDGET = "ai2/d5"
+BUDGET = "ai2/es-platform"
 WORKSPACE = "ai2/earth-systems"
 
 DEFAULT_HELIOS_WEKA_BUCKET = BeakerWekaBucket("dfive-default", "/weka/dfive-default")
--- a/helios/nn/flexihelios.py
+++ b/helios/nn/flexihelios.py
@@ -819,6 +819,7 @@ class FlexiHeliosBase(nn.Module):
         learnable_channel_embeddings: bool = True,
         random_channel_embeddings: bool = False,
         use_flash_attn: bool = False,
+        qk_norm: bool = False,
     ) -> None:
         """Initialize the FlexiHeliosBase class."""
         super().__init__()
@@ -840,6 +841,7 @@ class FlexiHeliosBase(nn.Module):
                     num_heads,
                     mlp_ratio,
                     qkv_bias=True,
+                    qk_norm=qk_norm,
                     norm_layer=nn.LayerNorm,  # TODO: This should be configurable
                     cross_attn=self.cross_attn,
                     drop_path=drop_path,
@@ -1037,6 +1039,7 @@ class Encoder(FlexiHeliosBase):
         aggregate_then_project: bool = True,
         use_flash_attn: bool = False,
         frozen_patch_embeddings: bool = False,
+        qk_norm: bool = False,
     ):
         """Initialize the encoder.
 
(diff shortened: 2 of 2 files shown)


scripts/base_debug_scripts/contrastive_latentmim.py
```diff
@@ -42,6 +43,7 @@
 
 MAX_PATCH_SIZE = 8
 MIN_PATCH_SIZE = 1
+USE_SMALL_DATASET = False
```
Hgherzog: nit: Use 4_x_128 is clearer because small is ambiguous

scripts/base_debug_scripts/contrastive_latentmim.py
```diff
@@ -150,8 +152,12 @@ def build_dataloader_config(common: CommonComponents) -> HeliosDataLoaderConfig:
 
 def build_dataset_config(common: CommonComponents) -> HeliosDatasetConfig:
     """Build the dataset config for an experiment."""
+    if USE_SMALL_DATASET:
+        dataset_url = "/weka/dfive-default/helios/dataset/presto/h5py_data_w_missing_timesteps_128_x_4_zstd_3/landsat_openstreetmap_raster_sentinel1_sentinel2_l2a_srtm_worldcover/469892"
```
Hgherzog: nit: this is a path in a distributed filesystem weka not not a url

scripts/base_debug_scripts/contrastive_latentmim.py
```diff
@@ -150,8 +152,12 @@ def build_dataloader_config(common: CommonComponents) -> HeliosDataLoaderConfig:
 
 def build_dataset_config(common: CommonComponents) -> HeliosDatasetConfig:
     """Build the dataset config for an experiment."""
+    if USE_SMALL_DATASET:
```
Hgherzog: Suggestion: Because the 128 dataset has 4x as many samples it might be good to also configure the number of epochs to reduce by 4 if that is the case. Similarly, it may make more sense to set the warmup in steps rather than epochs

scripts/base_debug_scripts/mae.py
```diff
@@ -44,6 +45,7 @@
 logger = logging.getLogger(__name__)
 MAX_PATCH_SIZE = 16  # NOTE: actual patch_size <= max_patch_size
```
Hgherzog: @pjreddie Is there a reason mae should be using a different patch size range? Or should we just align with others

scripts/base_debug_scripts/readme.MD
```diff
@@ -6,3 +6,5 @@ These settings should do about 20 minute per epoch on 1 GPU with batch size 128
 For faster Training use more gpus bigger batch size potentially more workers until you are saturated
 
 IF You want to share the runs with others please set the wandb project use override `--trainer.wandb.project=helios-train`
+
+If debugging different architectures or datasets, please remove checkpoints in `./local_output/checkpoints/<user>/test_run` (for sessions) or `/weka/dfive-default/helios/checkpoints/<user>/test_run` (for batch jobs).
```
Hgherzog, line 10: Thanks for adding this!

scripts/base_debug_scripts/mae.py
pjreddie: no particular reason I just thought it was weird to do MAE with such a small patch size, we never run it standalone anyway we could probably delete it or just update it to be in line with other scripts

ryspark: Thanks, will update patch size

PR #285, opened 2025-07-01
Henryh/eval label efficiency

Hgherzog:


scripts/2025_06_23_naip/eval.py
```diff
@@ -59,31 +59,31 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
     # Safe to collect everys tep for now
     garbage_collector_callback = GarbageCollectorCallback(gc_interval=1)
     EVAL_TASKS = {
-        "m-eurosat": DownstreamTaskConfig(
+        "m_eurosat": DownstreamTaskConfig(
```
Hgherzog, line 62: - gets converted to _ by omegaconfig so using _ is better for overiding keys

helios/evals/datasets/__init__.py
```diff
@@ -15,12 +16,24 @@
 logger = logging.getLogger(__name__)
 
 
+class EvalDatasetPartition(StrEnum):
+    """Enum for different dataset partitions."""
+
+    DEFAULT = "default"  # 1.0x_train
```
gabrieltseng: nit - maybe `TRAIN_1X` is better? I can imagine `DEFAULT` being ambiguous for someone reading the code.

scripts/2025_06_23_naip/eval.py
gabrieltseng: we can just get rid of the `m-`. I can't imagine any of us are confused that these are the geobench versions.

helios/train/callbacks/evaluator_callback.py
```diff
@@ -313,6 +317,7 @@ def build(self, trainer: Trainer) -> Callback | None:
                     eval_interval=task.eval_interval,
                     epochs=task.epochs,
                     eval_mode=task.eval_mode,
+                    partition=task.partition,
```
gabrieltseng, line 320: I think we want this partition to be at the downstream task config level. not at the evaluator level right?

Since i'd like (in a single run) to be able to run eurosat and eurosat 0.01x.

scripts/2025_06_23_naip/eval.py
```diff
@@ -111,46 +111,47 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
...
-            num_workers=2,
-            pooling_type=PoolingType.MEAN,
-            norm_stats_from_pretrained=True,
-            probe_lr=0.002,
-            eval_interval=Duration.epochs(10),
-            input_modalities=[Modality.SENTINEL1.name, Modality.LANDSAT.name],
-            epochs=50,
-        ),
+        # THESE EVALS DO NOT yet supportt partition
+        # "sickle_sentinel1": DownstreamTaskConfig(
```
gabrieltseng: As a quick change so that these still run, can we just add a kwarg for the partition that errors out if its not default?

(this is also partly solved by having the partition at the downstreamtaskconfig level, not the evaluatorconfig level).

scripts/get_max_eval_metrics_from_wandb.py
```diff
@@ -68,15 +138,43 @@ def get_max_metrics(project_name: str, run_prefix: str) -> dict[str, float]:
...
+
+        print("\nFinal Results:")
+        for metric in METRICS:
             try:
-                metric = metric.replace("-", "_")
                 k = f"eval/{metric}"
                 print(f"{metric} {metrics[k]}")
             except KeyError:
-                print(f"Metric {metric} not found")
+                try:
```
gabrieltseng, line 173: hah nice i had the same change locally.

helios/train/callbacks/evaluator_callback.py
Hgherzog: Currently, we create a new Downstream Evaluator for every task so if you configure a separate Task called like m_eurosat0.01x we will create a new evaluator that evaluates for that task. The partition is settable in the task config and is just passed to the evaluator.

scripts/2025_06_23_naip/eval.py
Hgherzog: This shouldn't error out. I just commented them out because I wanted the sweeps to finish faster. I can uncomment them.

helios/train/callbacks/evaluator_callback.py
Hgherzog: @gabrieltseng Feel free to merge this while I am gone if you want

scripts/2025_06_23_naip/eval.py
```diff
@@ -59,31 +59,31 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
     # Safe to collect everys tep for now
     garbage_collector_callback = GarbageCollectorCallback(gc_interval=1)
     EVAL_TASKS = {
-        "m-eurosat": DownstreamTaskConfig(
+        "m_eurosat": DownstreamTaskConfig(
             dataset="m-eurosat",
             embedding_batch_size=128,
-            num_workers=8,
+            num_workers=0,
```
gabrieltseng: is this a temporary change for local testing?

helios/train/callbacks/evaluator_callback.py
gabrieltseng: ah i see - this makes sense.

scripts/2025_06_23_naip/eval.py
gabrieltseng: fine for this script.

scripts/2025_06_23_naip/eval.py
gabrieltseng: Assuming yes and reverting

PR #286, opened 2025-07-04
Togo in loop

gabrieltseng:
Adding the Togo cropharvest dataset.

https://wandb.ai/eai-ai2/2025_06_30_dataset_percentage__eval/runs/1sp0450q/overview

This actually also gives us (for free) evals in China (cropland), Kenya (maize) and Brazil (coffee), example China run:
https://wandb.ai/eai-ai2/2025_06_30_dataset_percentage__eval/runs/nti2m1pl

and example China run with fewer timesteps:
https://beaker.allen.ai/orgs/ai2/workspaces/earth-systems/work/01JZX8XKJP7ZBSM020NMP13FP3/logs?jobId=01JZX8XKQB2M2NDHHDVV79JEA8


helios/evals/datasets/cropharvest.py
```diff
@@ -0,0 +1,263 @@
...
+        days = torch.ones_like(months)
+        # years are not used. Currently (20250710) flexihelios only uses the months when
+        # computing the composite encodings
+        years = torch.ones_like(months) * 2017
+        timestamp = torch.stack([days, months, years], dim=-1)  # t, c=3
+
+        if self.norm_stats_from_pretrained:
+            s2 = self.normalizer_computed.normalize(Modality.SENTINEL2_L2A, s2)
+            s1 = self.normalizer_computed.normalize(Modality.SENTINEL1, s1)
+            srtm = self.normalizer_computed.normalize(Modality.SRTM, srtm)
```
gabrieltseng, line 315: Our current training setup makes this a little tricky. SRTM *is* a seen modality, but its always masked from the perspective of the online encoder! so a todo is to handle that.

.pre-commit-config.yaml
```diff
@@ -20,6 +20,7 @@ repos:
       - id: check-case-conflict
       - id: check-docstring-first
       - id: check-added-large-files
+        args: ["--maxkb=2000"]
```
gabrieltseng, line 23: This is for the cropharvest shapefiles, the max file size is 1500Kb (the default limit is 500Kb)

helios/evals/datasets/breizhcrops.py
```diff
@@ -148,8 +148,8 @@ def __getitem__(self, idx: int) -> tuple[MaskedHeliosSample, torch.Tensor]:
             :,
             :,
             :,
-            EVAL_TO_HELIOS_S2_BANDS,
-        ]
+            INPUT_TO_OUTPUT_BAND_MAPPING,
+        ][:, :, :, EVAL_TO_HELIOS_S2_BANDS]
```
gabrieltseng, line 152: I think (?) this was a breizhcrops bug?

helios/evals/datasets/cropharvest_package/__init__.py
```diff
@@ -0,0 +1 @@
+# copied from github.com/nasaharvest/cropharvest
```
gabrieltseng, line 1: I am sorry about this :( the cropharvest package has some heavy dependencies so (IMO) this is the cleanest thing to do, but I appreciate it adds a lot of cropharvest specific code. Hopefully its suitably hidden away here.

helios/evals/datasets/breizhcrops.py
gabrieltseng: or maybe not? it seems like it does marginally worse with this change (0.72 vs 0.74). confused

helios/evals/datasets/cropharvest.py
```diff
@@ -0,0 +1,315 @@
...
+        834.5202295842477,
+        1099.279298427952,
+        947.4737894434511,
+        9.808570091118522,
+        0.0033269562766611444,
+        655.0334629731693,
+        6.18644770735272,
+        0.17878359339732122,
+    ]
+)
```
yawenzzzz: Suggest adding keys to these values, so that it's clearer which bands these values correspond to. Also, the names `X_MEAN` and `X_STD` is not intuitive, consider using more informative names, like the exact modalities.

yawenzzzz: or adding some comments to this, just help to read.

helios/evals/datasets/breizhcrops.py
gabrieltseng: okay it doesn't matter since the breizhcrops bands are the same as the helios eval bands, so the first indexing doesn't do anything. Keeping it for robustness.

helios/evals/datasets/cropharvest.py
gabrieltseng: added ! its for all the modalities (since they're treated as a single array in the cropharvest package).

PR #287, opened 2025-07-09
Henryh/add more probing heads

Hgherzog:
Enables attentive probing for segmentation tasks across modalities. I left enabling attentive pooling across space and time to another issue https://github.com/allenai/helios/issues/288.  

- Refactors Downstream Evaluatior to accept a task directly so it is easier to add new configurations to a task config

### DIFF
--- a/helios/evals/linear_probe.py
+++ b/helios/evals/linear_probe.py
@@ -57,7 +57,7 @@ class AttnPoolLinearProbe(nn.Module):
         nn.init.trunc_normal_(self.linear.weight, std=0.02)
         nn.init.zeros_(self.linear.bias)
 
-    def forward(self, feat_tokens: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
+    def forward(self, feat_tokens: torch.Tensor) -> dict:
         """Forward pass for attention pooling linear probe.
 
         Args:
@@ -90,6 +90,7 @@ class AttnPoolLinearProbe(nn.Module):
         x = x.reshape(B, H, W, D)
         return {"logits": self.linear(x), "attn_weights": attn_weights}
 
+
 class LinearProbe(nn.Module):
     """Linear Probe for classification tasks."""
 
@@ -102,7 +103,7 @@ class LinearProbe(nn.Module):
         else:
             self.batchnorm = nn.Identity()
 
-    def forward(self, x: torch.Tensor) -> torch.Tensor:
+    def forward(self, x: torch.Tensor) -> dict:
         """Forward pass for linear probe."""
         return {"logits": self.linear(self.batchnorm(x))}
 


helios/evals/linear_probe.py
```diff
@@ -17,6 +18,79 @@
...
+                - Attention weights, shape (B*H*W, num_heads, 1, N).
+        """
+        B, H, W, N, D = feat_tokens.shape
+        feat_tokens = rearrange(feat_tokens, "b h w n d -> (b h w) n d")
+        collapsed_dim = B * H * W
+        q = self.query_token.expand(collapsed_dim, 1, -1)
+        q = q.reshape(
+            collapsed_dim, 1, self.num_heads, D // self.num_heads
+        )  # [B, 1, head, D_head]
+        q = q.permute(0, 2, 1, 3)  # [B, head, 1, D_head]
```
favyen2: Consider using rearrange here, then the comment may not be needed.

helios/evals/linear_probe.py
```diff
@@ -121,9 +211,13 @@ def train_probe(
             batch_emb = batch_emb.to(device)
 
             with torch.amp.autocast(device_type=device.type, dtype=torch.bfloat16):
+                if isinstance(probe, nn.Sequential):
```
favyen2: Consider either saving probe_type and using that instead here, or setting a separate flag attribute indicating whether average pooling should be applied at each spatial patch.

helios/evals/linear_probe.py
```diff
@@ -121,9 +211,13 @@ def train_probe(
...
+                    batch_emb = torch.mean(batch_emb, dim=(-2))
                 logits = probe(
                     batch_emb
                 )  # (bsz, num_patches, logits_per_patch) or (bsz, n_cls)
+                if isinstance(logits, tuple):
+                    logits, attn_weights = logits
```
favyen2: I don't think attn_weights is used here.

Also, consider returning dict from the different Probe implementations instead of either logits or (logits, attn_weights) so that it can be consistent for accessing d["logits"] and then can check if it returned `"attn_weights" in d`.

helios/evals/linear_probe.py
```diff
@@ -202,6 +300,15 @@ def evaluate_probe(
             preds = torch.argmax(logits, dim=1).cpu()
             all_preds.append(preds)
             all_labels.append(batch_labels)
+            if not isinstance(probe, nn.Sequential):
+                all_attn_weights.append(attn_weights)
+
+    if not isinstance(probe, nn.Sequential):
+        all_attn_weights_tensor = torch.cat(all_attn_weights)
+        per_head = all_attn_weights_tensor.mean(dim=(0, 2))  # → [heads, 3]
```
favyen2: Suggest to replace 3 in comment with "modalities".

helios/train/callbacks/evaluator_callback.py
```diff
@@ -137,6 +143,7 @@ def _get_embeddings(
             model=self.trainer.train_module.model.encoder,
             patch_size=self.patch_size,
             pooling_type=self.pooling_type,
+            concat_features=(self.probe_type == ProbeType.ATTNPOOL),
```
favyen2, line 146: Hm, I thought you added logic in `linear_probe.py` to apply `batch_emb = torch.mean(batch_emb, dim=(-2))` if it is not attentive pooling (which would imply you need to always set concat_features=True here so it is the expected shape when it ends up in linear_probe.py).

helios/nn/flexihelios.py
```diff
@@ -176,7 +179,29 @@ def pool_unmasked_tokens(
             spatial_pooling: Whether to keep the spatial dimensions when pooling. If true,
                 this expects the masks within a spatial modality to be consistent (e.g. all
                 s2 tokens would have the same mask.)
+            concat_features: Whether to concatenate the features instead of averaging them, only enabled for spatial pooling as of now
```
favyen2: Could be useful to clarify that concat_features will only work when all the tokens are unmasked.

helios/evals/linear_probe.py
Hgherzog: I am requiring all the averaging to be done in get_embeddings based on the probe type. The pooling here was leftover from when I was debugging the attn pool method. Thanks for catching this

PR #289, opened 2025-07-09
Add WorldCereal

gabrieltseng:
https://wandb.ai/eai-ai2/v0.2_sweep/runs/fr9268co

### DIFF
--- a/helios/dataset/convert_to_h5py.py
+++ b/helios/dataset/convert_to_h5py.py
@@ -16,14 +16,7 @@ from olmo_core.config import Config
 from tqdm import tqdm
 from upath import UPath
 
-from helios.data.constants import (
-    IMAGE_TILE_SIZE,
-    SENTINEL1_NODATA,
-    YEAR_NUM_TIMESTEPS,
-    Modality,
-    ModalitySpec,
-    TimeSpan,
-)
+from helios.data.constants import IMAGE_TILE_SIZE, Modality, ModalitySpec, TimeSpan
 from helios.data.utils import convert_to_db
 from helios.dataset.parse import parse_helios_dataset
 from helios.dataset.sample import (
@@ -35,8 +28,6 @@ from helios.dataset.sample import (
 from helios.dataset.utils import get_modality_specs_from_names
 
 logger = logging.getLogger(__name__)
-
-mp.set_start_method("spawn", force=True)
 
 
 @dataclass
@@ -93,7 +84,6 @@ class ConvertToH5py:
         self,
         tile_path: UPath,
         supported_modalities: list[ModalitySpec],
-        multiprocessed_sample_processing: bool = True,
         multiprocessed_h5_creation: bool = True,
         compression: str | None = None,
         compression_opts: int | None = None,
@@ -107,7 +97,6 @@ class ConvertToH5py:
         Args:
             tile_path: The path to the tile directory, containing csvs for each modality and tiles
             supported_modalities: The list of modalities to convert to h5py that will be available in this dataset
-            multiprocessed_sample_processing: Whether to process samples in parallel
             multiprocessed_h5_creation: Whether to create the h5py files in parallel
             compression: Compression algorithm to use (None, "gzip", "lzf", "szip")
             compression_opts: Compression level (0-9 for gzip), only used with gzip compression
@@ -124,7 +113,6 @@ class ConvertToH5py:
         self.tile_path = tile_path
         self.supported_modalities = supported_modalities
         logger.info(f"Supported modalities: {self.supported_modalities}")
-        self.multiprocessed_sample_processing = multiprocessed_sample_processing
         self.multiprocessed_h5_creation = multiprocessed_h5_creation
         self.compression = compression
         self.compression_opts = compression_opts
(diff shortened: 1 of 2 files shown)


data/rslearn_dataset_configs/config_worldcereal.json
```diff
@@ -0,0 +1,26 @@
...
+            "tc-maize-main-maize-confidence",
+            "tc-maize-second-irrigation-confidence",
+            "tc-maize-second-maize-confidence",
+            "tc-springcereals-springcereals-confidence",
+            "tc-wintercereals-irrigation-confidence"
+          ],
+          "dtype": "float32"
+        }
+      ],
+      "data_source": {
```
favyen2, line 12: Is the two-week time period fine here because the items in WorldCereal data source have time range of one calendar year? Or maybe it is just time range of the agricultural season but we anyway don't want to match that to windows outside the agricultural season?

favyen2: nevermind I think the WorldCereal is one timestamp so it doesn't really have a time range in the items

helios/data/dataloader.py
```diff
@@ -30,8 +30,6 @@
 
 logger = logging.getLogger(__name__)
 
-BASE_TOKEN_BUDGET = 1500
```
gabrieltseng, line 33: This was not being used anywhere

helios/data/constants.py
```diff
@@ -23,6 +23,12 @@
 # Default maximum sequence length.
 MAX_SEQUENCE_LENGTH = 12
 
+# Default nodata value for Sentinel-1 data.
+SENTINEL1_NODATA = -32768
+
+# Number of timesteps for YEAR data.
+YEAR_NUM_TIMESTEPS = 12
```
gabrieltseng, line 30: changes from #319  for the run_h5_conversion updates

scripts/2025_08_14_worldcereal/train.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+def build_dataloader_config(common: CommonComponents) -> HeliosDataLoaderConfig:
+    """Build the dataloader config for an experiment."""
+    # things should be set during building
+
+    return HeliosDataLoaderConfig(
+        num_workers=16,
+        global_batch_size=512,
+        token_budget=2500,
```
gabrieltseng, line 166: this requires more than 1500 tokens

helios/dataset/convert_to_h5py.py
```diff
@@ -164,16 +176,19 @@ def process_sample_into_h5(
         """Process a sample into an h5 file."""
         i, (sublock_index, sample) = index_sample_tuple
         h5_file_path = self._get_h5_file_path(i)
-        if h5_file_path.exists():
-            return
+        # Comment out the exist check because if somehow the h5 generation is interrupted
```
favyen2: Instead of commenting it out like this, I think it would be better to remove the exists check and comment the _create_h5_file call like:
```
# This call will fail if the file already exists, but this is desired because there could be some
# broken h5 files from a previously interrupted h5 generation.
```

helios/dataset/convert_to_h5py.py
```diff
@@ -164,16 +176,19 @@ def process_sample_into_h5(
...
         self._create_h5_file(sample, h5_file_path, sublock_index)
 
     def create_h5_dataset(self, samples: list[tuple[int, SampleInformation]]) -> None:
         """Create a dataset of the samples in h5 format in a shared weka directory under the given fingerprint."""
         total_sample_indices = len(samples)
 
         if self.multiprocessed_h5_creation:
-            num_processes = max(1, mp.cpu_count() - 4)
+            # Processes may go to sleep state if we use too many processes
+            num_processes = max(1, mp.cpu_count() - 10)
```
favyen2: Yawen has this change too but I think it would be better to make the num_processes a configurable option instead.

helios/dataset/convert_to_h5py.py
```diff
@@ -276,19 +291,88 @@ def _create_missing_timesteps_masks(
...
+            image = self.load_sample(sample_modality, sample)
+            # Remove modalities that contains any nan
+            if np.any(np.isnan(image)):
+                logger.warning(
+                    f"Image for modality {modality.name} contains NaN values, removing this modality"
+                )
+                modalities_to_remove.add(modality)
+            # Remove ERA5 and OSM Raster and worldcereal if they are all zeros
+            if (
+                # modality == Modality.ERA5_10 or
```
favyen2: If there is code needed from the ERA5 PR it is better to have this PR depend on that one instead of having the same changes in both PRs. As is, I would suggest merging the ERA5 PR first, so that commenting out these lines like this is not needed.

gabrieltseng: Yes this is a good point. Since the h5 script is isolated from the world cereal changes, I'll undo the changes to convert_to_h5py.py and let those get merged in in Yawen's PR

helios/dataset/convert_to_h5py.py
gabrieltseng: Undone these changes so they are contained in Yawen's PR: https://github.com/allenai/helios/pull/289#discussion_r2277569133

helios/dataset/convert_to_h5py.py
gabrieltseng: Undone these changes so they are contained in Yawen's PR: https://github.com/allenai/helios/pull/289#discussion_r2277569133

PR #294, opened 2025-07-18
Henryh/add nccl debug overide

Hgherzog:
Add an override to automatically turn on a bunch of NCCL DEBUG Env vars to help track down distributed bugs

### DIFF
--- a/helios/internal/common.py
+++ b/helios/internal/common.py
@@ -190,7 +190,8 @@ def build_common_components(
     # Extract nccl_debug from overrides if present
     nccl_debug = False
     for override in overrides:
-        if override.startswith("common.nccl_debug="):
+        if override.startswith("--common.nccl_debug="):
+            logger.info(f"Setting nccl_debug to {override}")
             nccl_debug = override.split("=")[1].lower() in ("true", "1", "yes")
             break
 


helios/internal/common.py
```diff
@@ -123,7 +123,12 @@ def build_launch_config(
         allow_dirty=False,
         priority=BeakerPriority.high,
         env_vars=[
-            BeakerEnvVar(name="NCCL_DEBUG", value="INFO" if nccl_debug else "WARN"),
+            BeakerEnvVar(name="NCCL_DEBUG", value="DETAIL" if nccl_debug else "WARN"),
+            BeakerEnvVar(
+                name="TORCH_NCCL_TRACE_BUFFER_SIZE",
+                value="1000000000" if nccl_debug else "0",
```
pjreddie, line 129: from what I remember the default is like 2000? This value seems quite high but idk if it even matters, just tagging this in case

helios/internal/common.py
```diff
@@ -182,14 +187,26 @@ def build_common_components(
     if cmd == SubCmd.launch_prep:
         cmd_to_launch = SubCmd.prep
 
+    # Extract nccl_debug from overrides if present
```
pjreddie, line 190: why would it not just be `--launch.nccl_debug=true/false` why do you need all this?

Hgherzog: Because the launch config doesn't have a nccl_debug field it just contains many env vars. The build launch function need to process this and use this key to decide whether or not to set the env vars in the config

PR #295, opened 2025-07-18
Henryh/repeat data loading

Hgherzog:
When training with small datasets for many steps, dataloader worker respawning becomes a bottleneck for training. This allows us to fit many epochs of training within a single olmo_core epoch. It is also usable for large datasets but at that point the spawn times don't add up as much.

### DIFF
--- a/helios/data/dataloader.py
+++ b/helios/data/dataloader.py
@@ -296,10 +296,7 @@ class HeliosDataLoader(DataLoaderBase):
                 parts.append(f"{key}{value}")
         return "_".join(parts)
 
-    def get_mock_batch(self) -> HeliosSample:
-        """Get a mock batch, for dry-run of forward and backward pass."""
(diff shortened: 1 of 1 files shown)


helios/data/dataloader.py
```diff
@@ -320,26 +336,19 @@ def _get_mock_sample(self, rng: np.random.Generator) -> HeliosSample:
...
                     patch_size,
-                    self._get_mock_sample(rng).subset(
+                    HeliosSample(**output_dict).subset(
                         patch_size,
                         max_tokens_per_instance=1500,
                         sampled_hw_p=6,
                         current_length=12,
                     ),
                 )
-                for num in range(batch_size)
```
favyen2: What was the reason to make it one sample instead of batch for the mock?

Hgherzog: I think this is just a vestige of before the other PR was merged. It should be a batch.

PR #297, opened 2025-07-23
add geobench support for landsat

gabrieltseng:
We get `0.33` using KNN which seems to be in the right ball park (but very much on the low end - Panopticon reports accuracies between 33 - 53) 

Using a linear probe gets us 0.44 (https://beaker.allen.ai/orgs/ai2/workspaces/earth-systems/work/01K18J34MH8VM87XPGWXQCKQMP).

### DIFF
--- a/helios/data/dataloader.py
+++ b/helios/data/dataloader.py
@@ -61,6 +61,7 @@ class HeliosDataLoader(DataLoaderBase):
         drop_last: bool = True,
         persistent_workers: bool = True,
         multiprocessing_context: str = "spawn",
+        num_dataset_repeats_per_epoch: int = 1,
     ):
         """Initialize the HeliosDataLoader."""
         super().__init__(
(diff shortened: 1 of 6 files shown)


helios/evals/datasets/configs.py
```diff
@@ -66,6 +66,19 @@ class EvalDatasetConfig(NamedTuple):
...
     ),
+    "m-forestnet": EvalDatasetConfig(
+        task_type=TaskType.CLASSIFICATION,
+        imputes=[
+            # src (we have), tgt (we want), using the geobench L8 names
+            # we don't need to impute B8 since our
+            # _landsathelios2geobench_name implicitly does it for us,
+            ("02 - Blue", "01 - Coastal aerosol"),
+            ("07 - SWIR2", "09 - Cirrus"),
+            ("07 - SWIR2", "10 - Tirs1"),
```
gabrieltseng, line 77: maybe bands 10 and 11 should be a separate band set?

helios/evals/datasets/geobench_dataset.py
```diff
@@ -29,6 +34,32 @@
...
+        "B7": "07 - SWIR2",
+        # B8 is the panchromatic band. Geobech does not include it.
+        # its wavelength most overlaps with B3, so this is what we
+        # will treat as B8 from the GeoBench dataset.
+        "B8": "03 - Green",
+        "B9": "09 - Cirrus",
+        "B10": "10 - Tirs1",
+        "B11": "10 - Tirs1",
+    }
+    return transform[band_name]
```
yawenzzzz, line 57: yeah, I think the imputation might cause the performance drop, they only provides 6 bands while we require 11 bands, especially when `norm_stats_from_pretrained = True` as the pretrained norm stats actually corresponds to the original bands, in case of imputation, maybe we should replace the norm stats with the impute one (what we have).

gabrieltseng: done in https://github.com/allenai/helios/pull/297/commits/735994e8d8059219da00e1faca33a6bb490abc6a

PR #298, opened 2025-07-24
Ignore targets which are too similar

gabrieltseng:
Following @Hgherzog 's analysis

https://wandb.ai/eai-ai2/v0.2_sweep/runs/xuyql1j5?nw=nwusergabrieltseng

### DIFF
--- a/helios/train/loss.py
+++ b/helios/train/loss.py
@@ -273,7 +273,7 @@ class PatchDiscriminationLoss(Loss):
         scores = torch.einsum("npd,nqd->npq", pred, target) / self.tau
         count = (all_masks == MaskValue.DECODER.value).sum(dim=-1)
         if self.mask_other_samples:
-            logit_mask = torch.full_like(scores, -torch.finfo(scores.dtype).max)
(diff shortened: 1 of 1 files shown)


helios/train/loss.py
```diff
@@ -177,6 +186,25 @@ def compute(
...
+                )
+                # unmask the diagonal, and re add the bz dimension
+                target_similarities_thresholded.fill_diagonal_(0)
+                target_similarities_thresholded = (
+                    target_similarities_thresholded.unsqueeze(0)
+                )
+                # this will essentially set the loss of the similar patches to 0
+                score_sample[target_similarities_thresholded] = -torch.finfo(
+                    score_sample.dtype
+                ).max
```
favyen2, line 206: May be better to use `torch.finfo(...).min` here instead of `-torch.finfo(...).max`.

gabrieltseng: thanks! updated

gabrieltseng:
closing - superseded by @yawenzzzz 's PR

PR #299, opened 2025-07-29
Henryh/dino v2 eval

Hgherzog:
Running Evals with Panopticon and DinoV2

- Added EvalWrapper class to have a common interface for evals depending on the model being used

Also found some bugs along the way
- Fix bugs for dataset based normalization for pastis
- Fix dataset normalization bug in so2sat with 10k multiply

### DIFF
--- a/helios/evals/models/dinov2/dinov2.py
+++ b/helios/evals/models/dinov2/dinov2.py
@@ -138,7 +138,9 @@ class DINOv2(nn.Module):
                     "DINOv2 does not yet support multiple modalities via multiple forward passes"
                 )
             if modality not in self.supported_modalities:
-                logger.warning(f"Skipping modality {modality} as it is not in the supported modalities list {self.supported_modalities}")
+                logger.warning(
+                    f"Skipping modality {modality} as it is not in the supported modalities list {self.supported_modalities}"
+                )
                 continue  # Skip non-rgb modalities
 
             data = getattr(masked_helios_sample, modality)
--- a/helios/evals/models/panopticon/panopticon.py
+++ b/helios/evals/models/panopticon/panopticon.py
@@ -100,7 +100,9 @@ class Panopticon(nn.Module):
             modality_yaml_name = "landsat8"
         else:
             modality_yaml_name = modality
-        with resources.open_text("helios.evals.models.panopticon.sensors", f"{modality_yaml_name}.yaml") as f:
+        with resources.open_text(
+            "helios.evals.models.panopticon.sensors", f"{modality_yaml_name}.yaml"
+        ) as f:
             sensor_config = yaml.safe_load(f)
         modality_spec = Modality.get(modality)
         # Data is prepared in helios band order so we need to tell panopticon whcich band it is
--- a/scripts/dino_v2_evals/dino_v2_eval.py
+++ b/scripts/dino_v2_evals/dino_v2_eval.py
@@ -109,7 +109,7 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
             probe_lr=0.1,
             eval_interval=Duration.epochs(10),
         ),
-              "pastis_sentinel2": DownstreamTaskConfig(
+        "pastis_sentinel2": DownstreamTaskConfig(
             dataset="pastis",
             embedding_batch_size=32,
             probe_batch_size=8,
--- a/scripts/dino_v2_evals/eval_sweep.py
+++ b/scripts/dino_v2_evals/eval_sweep.py
@@ -7,7 +7,12 @@ import subprocess  # nosec
 # Normalization
 # helios pretrained, dataset norms and imagenet normalization
 
-dino_v2_torchub_id = ["dinov2_vitb14", "dinov2_vitl14", "dinov2_vitg14", "dinov2_vitb14_reg"]
+dino_v2_torchub_ids = [
+    "dinov2_vitb14",
+    "dinov2_vitl14",
+    "dinov2_vitg14",
+    "dinov2_vitb14_reg",
+]
 LP_LRs = [1e-4, 5e-4, 1e-3, 5e-3, 1e-2, 5e-2, 1e-1, 5e-1]
 
 Normalization_MODES = ["imagenet", "dataset", "helios"]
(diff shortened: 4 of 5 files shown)


scripts/2025_06_26_dataset_percentage_experiments/eval.py
Hgherzog, line 1: I will revert the changes in this folder it was for debugging

helios/evals/dinov2/dinov2.py
```diff
@@ -0,0 +1,231 @@
+"""DINOv2 model wrapper."""
```
gabrieltseng: can dinov2 and panopticon be moved into `helios/evals/models`?

This will prevent `helios/evals` from getting too cluttered, especially as we add more models.

helios/evals/dinov2/dinov2.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+def make_normalize_transform(
+    mean = IMAGENET_DEFAULT_MEAN,
+    std = IMAGENET_DEFAULT_STD,
+) -> transforms.Normalize:
+    return transforms.Normalize(mean=mean, std=std)
+
+HELIOS_SENTINEL2_RGB_BANDS = [3, 2, 1]
+HELIOS_LANDSAT_RGB_BANDS = [4, 3, 2]
```
gabrieltseng: Could this be linked to `Modality` instead, so that it doesn't break if we change things?

I think it would look like

```python
HELIOS_SENTINEL2_RGB_BANDS = [Modality.SENTINEL2_L2A.band_order.index(b) for b in ["B04", "B03", "B02"]]
```
I also think `HELIOS_SENTINEL2_RGB_BANDS` should be `[2, 1, 0]`, not `[3, 2, 1]` since B1 is later in the band order:

```python
>>> from helios.data.constants import Modality
>>> Modality.SENTINEL2_L2A.band_order
['B02', 'B03', 'B04', 'B08', 'B05', 'B06', 'B07', 'B8A', 'B11', 'B12', 'B01', 'B09']
>>> Modality.SENTINEL2.band_order
['B02', 'B03', 'B04', 'B08', 'B05', 'B06', 'B07', 'B8A', 'B11', 'B12', 'B01', 'B09', 'B10']
```

helios/evals/dinov2/dinov2.py
```diff
@@ -0,0 +1,231 @@
...
+        per_timestep_inputs = []
+        for i, input_data_i in input_data_timesteps.items():
+            # Concatenate all modality data along channel dimension
+            concatenated_imgs = torch.cat(input_data_i, dim=1)
+
+            per_timestep_inputs.append(concatenated_imgs)
+        return per_timestep_inputs
+
+    # pooling type is on the timesteps only right now
+    def forward(self, masked_helios_sample: MaskedHeliosSample, pooling: PoolingType = PoolingType.MEAN, apply_imagenet_normalization: bool = False) -> torch.Tensor:
```
gabrieltseng: nit - `apply_imagenet_normalization` may be better as an argument to `__init__` so that we can control it from the config.

helios/evals/embeddings.py
```diff
@@ -1,31 +1,29 @@
 """Embeddings from models."""
 
 import logging
-
+import math
```
gabrieltseng: This import isn't used. I feel like `ruff` should have caught this?

gabrieltseng: Same with `import torch.nn.functional as F` below.

gabrieltseng: And `einops`, and `Modality`.

helios/evals/eval_wrapper.py
```diff
@@ -0,0 +1,114 @@
...
+
+class EvalWrapper:
+    def __init__(self, model: nn.Module, task_type: TaskType, patch_size: int, pooling_type: PoolingType, concat_features: bool = False, apply_imagenet_normalization: bool = False):
+        super().__init__()
+        self.model = model
+        self.task_type = task_type
+        self.patch_size = patch_size
+        self.pooling_type = pooling_type
+        self.concat_features = concat_features
+        self.apply_imagenet_normalization = apply_imagenet_normalization
```
gabrieltseng: this seems like a very specific kwarg to have in this general wrapper. I think only Dino and Panopticon will make use of them ; maybe it makes sense as a model specific kwarg (as suggested [here](https://github.com/allenai/helios/pull/299/files#r2242354491)).

helios/evals/panopticon/panopticon.py
```diff
@@ -0,0 +1,241 @@
...
+        patch_size: int = 14,
+    ):
+        """Initialize the Panopticon wrapper.
+
+        Args:
+            torchhub_id: The torch hub model ID for panopticon
+            patch_size: Patch size for the vision transformer (default 14)
+        """
+        super().__init__()
+        self.patch_size = patch_size
```
gabrieltseng: I don't think Panopticon supports variable patch sizes; should we throw a `ValueError` if patch size is not 14?

helios/evals/dinov2/dinov2.py
```diff
@@ -0,0 +1,231 @@
...
+                    align_corners=False
+                )
+            else:
+                # Resize the image to the dinov2 pre-training input size
+                image_size = 224
+                data_i = F.interpolate(
+                    data_i,
+                    size=(image_size, image_size),
+                    mode="bilinear",
+                    align_corners=False
```
gabrieltseng: nit - here and in `panopticon.py`, this code block could be simplified:

```python
new_height = self.patch_size if original_height == 1 else 224

data_i = F.interpolate(
        data_i,
        size=(new_height, new_height),
        mode="bilinear",
        align_corners=False
    )
```

helios/evals/embeddings.py
Hgherzog: I need to run pre-commit will take care of all of this.

helios/evals/dinov2/dinov2.py
Hgherzog: Good catch.

helios/evals/eval_wrapper.py
Hgherzog: it only even applies to dinov2

helios/evals/panopticon/panopticon.py
Hgherzog: I just made patch size a class variable because it is a constant for these models

PR #301, opened 2025-07-30
Add Mixup

gabrieltseng:
Screenshot 2025-07-30 at 1 46 10 PM

### DIFF
--- a/helios/internal/override_examples.md
+++ b/helios/internal/override_examples.md
@@ -21,3 +21,8 @@ Targeting specific Beaker host example:
 ```
 python scripts/X.py launch [name] ai2/titan-cirrascale --launch.hostnames=["titan-cs-aus-465.reviz.ai2.in","titan-cs-aus-466.reviz.ai2.in"]
 ```
+
+To add mixup, add the following CLI arguments:
+```
+--train_module.transform_config.transform_type=mixup --train_module.transform_config.transform_kwargs={"alpha": 1.3}
+```


helios/data/dataset.py
```diff
@@ -414,6 +414,46 @@ def subset(
                 new_data_dict[attribute] = modality
         return HeliosSample(**new_data_dict)
 
+    def mul_by_float(self, m: float) -> "HeliosSample":
```
pjreddie: nit: I'd call this `scale`

helios/data/dataset.py
```diff
@@ -414,6 +414,46 @@ def subset(
...
+        summed_dict: dict[str, ArrayTensor] = {}
+        for key, val in self.as_dict(ignore_nones=True).items():
+            assert val is not None  # keep mypy happy. True because ignore_nones=True
+            other_val = getattr(other, key)
+            if other_val is None:
+                raise ValueError(
+                    f"Add requires both HeliosSamples to have the same modalities, other is missing {key}"
+                )
+            summed_dict[key] = val + other_val
+        summed_dict["timestamps"] = timestamps_to_keep
```
pjreddie, line 438: hadn't thought about timestamps. this might get weird if we loop around for one batch, it's *possible* we don't even want it circular, just have one less timestamp but it seems like it works the way it is. maybe something to think about in the future

helios/train/train_module/latent_mim.py
```diff
@@ -204,12 +204,12 @@ def train_batch(
         # Split into micro-batches.
         microbatches = split_batch(batch_data, self.rank_microbatch_size)
         num_microbatches = len(microbatches)
-        for microbatch_idx, microbatch in enumerate(microbatches):
```
pjreddie: do we ever need the original? keeping it around just in case?

gabrieltseng: yea, it was due to how i implemented it originally. I'll revert this change

helios/data/dataset.py
gabrieltseng: updated

PR #308, opened 2025-08-04
Split eval dependencies out from the core set to simplify run-time dependencies.

joshhvulcan:
- Also splits dev-specific requirements into requirements-dev.txt
- Renames ai2-requirements to requirements-beaker.txt for clarity/consistency.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-ai2-olmo-core @ git+https://github.com/allenai/OLMo-core.git@ebe0b1981b81576ce7a87362501dc84c6d9adb3f # Pin here until >2.1.0 is released.
+ai2-olmo-core @ git+https://github.com/allenai/OLMo-core.git@abc12e50ba756c21e575452cfc6f150dafa9509e # Pin here until >2.1.0 is released.
 albumentations
 beaker-py==1.34.1
 breizhcrops==0.0.4.1 # Force the latest available


pyproject.toml
```diff
@@ -24,7 +24,12 @@ include = ["helios*"]
 
 [tool.setuptools.dynamic]
 dependencies = {file = ["requirements.txt"]}
-optional-dependencies = {extra = { file = ["extra_requirements.txt"] }}
+
+[tool.setuptools.dynamic.optional-dependencies]
+dataset_creation = { file = ["requirements_dataset_creation.txt"] }
+dev = { file = ["requirements-dev.txt"] }
+evals = { file = ["requirements-evals.txt"] }
+beaker = { file = ["requirements-beaker.txt"] }
```
Hgherzog, line 32: Suggestion: add an all group

requirements-beaker.txt
```diff
@@ -0,0 +1 @@
+beaker-py>=2.0
```
Hgherzog: I am a little scared about bumping this dependency. I think we need to launch training runs to confirm that this doesn't break anything. What does moving this dependency to >=2.0 help with?

cmwilhelm: 2.x had a bunch of breaking changes because it cuts over to the new beaker gRPC API (as opposed to the old json API). they opted not to make it backwards compatible because the interface wasn't easily mappable.

cmwilhelm: To wit, beaker-py should only really need `requests`, `pydantic`; the newer version needs a grpc client. Sam's team has also said they're supporting the v1 backend indefinitely.

joshhvulcan: This is already at >=2 in the existing `ai2_requirements.txt`: https://github.com/allenai/rslearn_projects/commit/5f497d6357673c2a2a9fce4a0df22a64ad38fbaa

pyproject.toml
joshhvulcan: @Hgherzog Would you want the dataset creation deps included in that? I haven't touched that one at all and there appears to be some overlap.  I'd rather not go deep on that one at this point so I am going to advocate to leave it out for now.

requirements-beaker.txt
Hgherzog: Helios and rslearn launch experiments differently so they may have different dependencies on beaker-py. Helios depends on olmo-core <2 https://github.com/allenai/OLMo-core/blob/main/pyproject.toml#L57 so this update would not be compatible with olmo_core. Is there any benefit to bumping it on your side? If not I would prefer to keep the previous pin.

pyproject.toml
Hgherzog: @joshhvulcan ^ I am agreeing with your advocating to leave it out.

pyproject.toml
```diff
@@ -11,7 +11,7 @@ authors = [
 ]
 readme = "README.md"
 license = {file = "LICENSE"}
-requires-python = ">=3.10"
+requires-python = ">=3.12"
```
Hgherzog: I need to double check and potentially update the base training image that we use so it uses 3.12 (I think it currently uses 3.11) I think I may be able to find a version with 3.12 and this will just work but better to double check.

Hgherzog: On this note, having a workflow that pulls our training iamge and installs the dependencies seems necessary to remove this manual checking step. Not sure how straightforward it is to pull a docker image from beaker and use it in github actions

Hgherzog: I _think_ the olmo-core iamge on beaker is always going to be the same as the one they publish to
 ghcr

requirements-beaker.txt
joshhvulcan: @Hgherzog @cmwilhelm See my comment.  I added an `all` group in response and have left the beaker requirements file out of that group.  Please advise on next steps.

Hgherzog: I suggest we revert to the  pinned version of beaker-py  `beaker-py==1.34.1`.

pyproject.toml
Hgherzog: @joshhvulcan would 3.11 be sufficient for now? If 3.12 is needed I would have to go make a new olmo-core image on beaker with 3.12 and confirm our entire workflow still works.

.pre-commit-config.yaml
```diff
@@ -72,7 +72,7 @@ repos:
             --fail-under=80,
           ]
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.4.4
+    rev: v0.12.8
```
Hgherzog, line 75: is this bump to be consistent across all of earth system or for another reason?

pyproject.toml
```diff
@@ -24,7 +24,13 @@ include = ["helios*"]
...
 [tool.setuptools.dynamic]
 dependencies = {file = ["requirements.txt"]}
-optional-dependencies = {extra = { file = ["extra_requirements.txt"] }}
+
+[tool.setuptools.dynamic.optional-dependencies]
+dataset_creation = { file = ["requirements_dataset_creation.txt"] }
+dev = { file = ["requirements-dev.txt"] }
+eval = { file = ["requirements-eval.txt"] }
+beaker = { file = ["requirements-beaker.txt"] }
+all = { file = ["requirements.txt", "requirements-dev.txt", "requirements-eval.txt"] }
```
Hgherzog, line 33: Can we add beaker reqs to all? We need that to launch training experiments and the code for that currently runs ` pip install -e '.[all]'`  on top of our docker image after cloning the commit of the repository

requirements.txt
```diff
@@ -1,25 +1,16 @@
-ai2-olmo-core @ git+https://github.com/allenai/OLMo-core.git@abc12e50ba756c21e575452cfc6f150dafa9509e # Pin here until >2.1.0 is released.
+ai2-olmo-core @ git+https://github.com/allenai/OLMo-core.git@ebe0b1981b81576ce7a87362501dc84c6d9adb3f # Pin here until >2.1.0 is released.
```
Hgherzog: We have updated the olmo-core commit to pull a bug-fix upstream. Use hash from main. 
```suggestion
ai2-olmo-core @ git+https://github.com/allenai/OLMo-core.git@abc12e50ba756c21e575452cfc6f150dafa9509e# Pin here until >2.1.0 is released.
```

pyproject.toml
joshhvulcan: Does beaker not install its own dependencies when installing it?

.pre-commit-config.yaml
joshhvulcan: Ruff wasn't in the requirements anywhere, so when I pip installed locally it grabbed the latest version and thats what I ended up pinning it to since there were some discrepancies between the two versions when it hit CI.  I can roll it back if you think its going to be a problem.

pyproject.toml
Hgherzog: not within the image. Different teams use different versions of beaker-py so we still need to install on top of the image to my knowledge.

.pre-commit-config.yaml
Hgherzog: I think it is okay to leave the bump in!

PR #312, opened 2025-08-05
Henryh/cross modality cheating

Hgherzog:
Major Changes
1. Adds pool Encoder and predictor so that we can pool in different dimensions at different layers of depth and then decode from the pooled tokens
2. Updated the encoder output to be a dict that can be unpacked to pass decoder kwargs specific to the paired decoder. The goal of this is to make it easier to add and design new encoder and deocoder architectures and also to not have to go update code in different train modules if we choose to return something else from the encoder. 

TODO: Tests of these new components still needed.

### DIFF
--- a/helios/nn/flexihelios.py
+++ b/helios/nn/flexihelios.py
@@ -1051,7 +1051,7 @@ class FlexiHeliosBase(nn.Module):
 
     @staticmethod
     def split_and_expand_per_modality(
-        x: Tensor, modalities_to_dims_dict: dict[str, tuple]
+        x: Tensor, modalities_to_dims_dict: dict
     ) -> dict[str, Tensor]:
         """Split and expand the tokens per modality.
 
--- a/helios/nn/pooled_modality_predictor.py
+++ b/helios/nn/pooled_modality_predictor.py
@@ -358,13 +358,12 @@ class EncodeEarlyAttnPool(Encoder):
     def apply_unpooled_attn(
         self,
         tokens_and_masks_dict: dict[str, Tensor],
-        modalities_to_dims_dict: dict[str, int],
+        modalities_to_dims_dict: dict,
         exit_ids_seq: Tensor | None = None,
         exited_tokens: Tensor | None = None,
         always_pass_none_mask_to_transformer: bool = False,
     ) -> dict[str, Tensor]:
         """Apply the attention to the tokens and masks."""
-
         tokens, mask = self.collapse_and_combine_hwtc(tokens_and_masks_dict)
 
         bool_mask = mask == MaskValue.ONLINE_ENCODER.value
@@ -466,7 +465,7 @@ class EncodeEarlyAttnPool(Encoder):
         )
         tokens_dict.update(original_masks_dict)
 
-        #TODO: token exit config isn't really meant to be used here but is no-op so leaving it in
+        # TODO: token exit config isn't really meant to be used here but is no-op so leaving it in
         tokens_dict = self.apply_unpooled_attn(
             tokens_dict,
             pre_pooled_modality_to_dims_dict,


helios/data/dataset.py
```diff
@@ -373,6 +373,7 @@ def subset(
         max_t = self._get_max_t_within_token_budget(
             sampled_hw_p, max_tokens_per_instance
         )
+        # Trying to reproduce the bug
```
gabrieltseng: nit - this can be removed?

helios/nn/flexihelios.py
```diff
@@ -1402,7 +1457,12 @@ def forward(
                 always_pass_none_mask_to_transformer=always_pass_none_mask_to_transformer,
             )
         output = TokensAndMasks(**patchified_tokens_and_masks)
-        return output, self.project_and_aggregate(output)
+        # TODO: we should probably switch this to a dict
```
gabrieltseng, line 1477: I have this as a named tuple in #264:  https://github.com/allenai/helios/blob/947c84d32f13e5262786e2e080087f8b5bda2e94/helios/nn/latent_mim.py#L25 

But yea in general we are tossing around quite a few model outputs and I agree organizing them better would be more desirable.

helios/nn/flexihelios.py
```diff
@@ -1751,6 +1822,7 @@ def forward(
         for modality in modalities_to_process:
             x_modality = getattr(x, modality)
             # Are these normalizations masked correctly?
+            # Does not account for missing tokens
```
gabrieltseng: We observed this mistake in Presto - fixing it did not change results significantly (but of course presto is a different model)

helios/nn/pooled_modality_predictor.py
```diff
@@ -0,0 +1,855 @@
...
+        """Initialize weights for the probe."""
+        nn.init.trunc_normal_(self.query_tokens, std=0.02)
+        nn.init.trunc_normal_(self.kv.weight, std=0.02)
+        nn.init.zeros_(self.kv.bias)
+
+        nn.init.zeros_(self.kv.bias)
+        if self.gate is not None:
+            nn.init.zeros_(self.gate.weight)  # start near uniform mix
+
+    def masked_mean(
```
gabrieltseng: is this used anywhere? can it be removed?

helios/evals/eval_wrapper.py
```diff
@@ -45,6 +48,7 @@ def __init__(
         self.pooling_type = pooling_type
         self.concat_features = concat_features
         self.spatial_pool = task_type == TaskType.SEGMENTATION
+        self.use_pooled_tokens = use_pooled_tokens
```
gabrieltseng, line 52: nit - maybe make sure the model is an `EncodeEarlyAttnPool` instance and that it can support pooled tokens

helios/nn/flexihelios.py
gabrieltseng: I see you use dicts a fair bit below. A nit, but I much prefer named tuples since it doesn't require guessing what the keys are.

helios/nn/pooled_modality_predictor.py
```diff
@@ -0,0 +1,855 @@
...
+        kwargs = self.as_dict(exclude_none=True, recurse=False)
+        # supported_modality_names is replaced by supported_modalities
+        kwargs.pop("supported_modality_names")
+        kwargs["supported_modalities"] = self.supported_modalities
+        logger.info(f"Predictor kwargs: {kwargs}")
+        return PooledModalityPredictor(**kwargs)
+
+
+# TODO: make sure 12 doing all the layers first still works
+class EncodeEarlyAttnPool(Encoder):
```
gabrieltseng, line 159: tiny nit: can the encoder be before the predictor just for consistency with other files

helios/nn/pooled_modality_predictor.py
```diff
@@ -0,0 +1,855 @@
...
+        x = self.apply_unpooled_attn(
+            x,
+            timestamps,
+            patch_size,
+            input_res,
+            token_exit_cfg,
+            always_pass_none_mask_to_transformer,
+        )
+        tokens_only_dict, original_masks_dict, pre_pooled_modality_to_dims_dict = (
+            self.split_tokens_masks_and_dims(x)
```
gabrieltseng, line 452: this is called twice, since `apply_unpooled_attn` also calls it. Can we just call it once and send the outputs to `apply_unpooled_attn`?

gabrieltseng: In general there's a bunch of duplicated code in here. Do we want to apply the composite encodings twice (once again before pooling?, lines 704 - 709)

helios/nn/flexihelios.py
Hgherzog: I agree named tuples are clearer though dicts can make the code more hackable.  Would you prefer for me to make into a Named Tuple here or let you take care of that change in your pr?

helios/nn/flexihelios.py
Hgherzog: All the operations are in the token dimension and not cross token so it ends up not mattering at all as the forward passes from the masks tokens are never backpropagated.

helios/nn/pooled_modality_predictor.py
Hgherzog: Ah yeah, at one point I was going to try to initialize the cross pool to just be an average of tokens but decided not to. Will remove it, feel free to come back to it if you think it is a good idea.

helios/nn/pooled_modality_predictor.py
Hgherzog: Good catch. I doubt it makes much difference either way so I will try to remove the redundancy and confirm it doesn't change the results.

Hgherzog: Screenshot 2025-08-23 at 5 11 31 PM
Seems like this even improved performance slightly!

helios/nn/flexihelios.py
gabrieltseng: Yes that sounds good - we can leave it as a dict here and I'll change it to a named tuple if I still feel that's better

PR #315, opened 2025-08-06
Run all evals wrapper

Hgherzog:
Provide a single script to launch any of our models given the script used to train it and the checkpoint to evaluate. You can optionally add a project to submit the evals to and pick the cluster

### DIFF
--- a/scripts/run_all_evals/full_eval_sweep.py
+++ b/scripts/run_all_evals/full_eval_sweep.py
@@ -62,6 +62,18 @@ helios_args = " ".join(
 )
 
 
+def loop_through_params():
+    """Yield a dict of the hps we are sweeping over."""
+    for lr in LP_LRs:
+        for norm_mode in Normalization_MODES:
+            for pooling_type in pooling_types:
+                yield {
+                    "lr": lr,
+                    "norm_mode": norm_mode,
+                    "pooling_type": pooling_type,
+                }
+
+
 def main():
     """Run the full evaluation sweep or just the defaults."""
     parser = argparse.ArgumentParser()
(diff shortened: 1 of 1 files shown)


scripts/run_all_evals/all_evals.py
```diff
@@ -0,0 +1,316 @@
...
+        pooling_type=PoolingType.MEAN,
+        norm_stats_from_pretrained=False,
+        eval_interval=Duration.epochs(5),
+    ),
+    "m_eurosat": DownstreamTaskConfig(
+        dataset="m-eurosat",
+        embedding_batch_size=128,
+        num_workers=0,
+        pooling_type=PoolingType.MEAN,
+        norm_stats_from_pretrained=False,  # True, #False,
```
yawenzzzz: Suggest changing this back to what we use before (and find better), so that sometimes we can just run the default norm type instead of sweeping them, though the norm sweep is good.

scripts/run_all_evals/full_eval_sweep.py
```diff
@@ -0,0 +1,110 @@
...
+
+                if project_name is None:
+                    project_name = "helios_in_loop_evals"
+                cmd = (
+                    f"TRAIN_SCRIPT_PATH={module_path} python3 scripts/run_all_evals/all_evals.py "
+                    f"launch {run_name} {cluster} --launch.priority=high {cmd_args} "
+                    f"--launch.task_name=eval --trainer.load_path={checkpoint_path} --trainer.callbacks.wandb.project={project_name}"
+                )
+                print(cmd)
+                subprocess.run(cmd, shell=True)  # nosec
```
favyen2: I think without `check=True` then if the process fails it may keep looping. If that is the case, it may be better to add `check=True` here.

scripts/run_all_evals/full_eval_sweep.py
```diff
@@ -0,0 +1,110 @@
...
+    parser.add_argument(
+        "--checkpoint_path", type=str, required=True, help="Checkpoint path"
+    )
+    parser.add_argument(
+        "--module_path", type=str, required=True, help="Path to module .py"
+    )
+    parser.add_argument(
+        "--project_name", type=str, required=False, help="Wandb project name"
+    )
+    args = parser.parse_args()
```
favyen2: Would be useful to have a way to provide custom arguments to include in the eval run. Since we often provide command-line overrides when launching training runs.

Hgherzog: Adding that right now.

PR #318, opened 2025-08-11
Add Cropland DataLayer data

favyen2:


helios/data/constants.py
```diff
@@ -402,6 +402,14 @@ class Modality:
         ignore_when_parsing=False,
     )
 
+    CDL = ModalitySpec(
```
Hgherzog, line 405: Suggestion: Use a more verbose name then CDL for the main constant variable.

favyen2: I agree it would be better like CROPLAND_DATA_LAYER for both constant name and the name everywhere but it is difficult to change it now (I could change just the constant name but all the other modalities are consistent between the name and the constant name and I think we make some assumptions about that in places) since the data has been generated into helios dataset, and I think we will quickly get used to this name. I will make sure to use more verbose name in the future.

PR #319, opened 2025-08-12
Intra-modality PatchDisc loss & CutMix

yawenzzzz:
1. Add ERA5, some changes in h5 generation mainly to check the validity of modalities in a sample
2. Add intra-modality PatchDisc loss & AllPatchDisc loss, runs: https://wandb.ai/eai-ai2/2025_07_21_era5_experiments?nw=yaf79gexv7
3. Add CutMix as an option in subsetting, runs: https://wandb.ai/eai-ai2/2025_07_21_era5_experiments?nw=bbjp0m2q60s


scripts/2025_07_21_add_era5/eval.py
yawenzzzz, line 1: This can be removed now since evals can be launched via #315

helios/data/dataset.py
```diff
@@ -363,55 +367,98 @@ def subset(
...
-                        start_w + sampled_hw
-                    )
-                    * modality_spec.image_tile_size_factor,
-                ]
-            elif modality_spec.is_time_only_varying:
-                new_data_dict[attribute] = modality[start_t : start_t + max_t]
-            elif modality_spec.is_static_in_space_and_time:
-                new_data_dict[attribute] = modality
+
+        if not apply_cutmix:
```
favyen2: Since these code paths are largely separate, it seems like it may be better to have separate `subset_cutmix` and `subset_default` functions that this `subset` functions calls instead of having it all in one function.

helios/data/dataset.py
```diff
@@ -482,6 +529,7 @@ class GetItemArgs(NamedTuple):
     patch_size: int
     sampled_hw_p: int
     token_budget: int | None = None
+    apply_cutmix: bool = False
```
favyen2: Is there a reason to have cutmix be a property of the data loader instead of the dataset? If not, then I think it is better to have it configured on the dataset, otherwise there are more changes needed to configure the data loader and then have this GetItemArgs include the option just to pass it off to the dataset.

helios/dataset/convert_to_h5py.py
```diff
@@ -164,16 +176,19 @@ def process_sample_into_h5(
...
         self._create_h5_file(sample, h5_file_path, sublock_index)
 
     def create_h5_dataset(self, samples: list[tuple[int, SampleInformation]]) -> None:
         """Create a dataset of the samples in h5 format in a shared weka directory under the given fingerprint."""
         total_sample_indices = len(samples)
 
         if self.multiprocessed_h5_creation:
-            num_processes = max(1, mp.cpu_count() - 4)
+            # Processes may go to sleep state if we use too many processes
+            num_processes = max(1, mp.cpu_count() - 10)
```
favyen2: Suggest to make this configurable instead of subtracting more. I also ran into issue where the processes were using too much memory so had to decrease this, so I think a configurable option will be most useful.

helios/dataset/sample.py
```diff
@@ -225,60 +225,70 @@ def load_image_for_sample(
...
-                    # In this case we need to downscale.
-                    # This should not be common, since usually bands would be stored at
-                    # the image tile resolution or lower. But it could happen for
-                    # OpenStreetMap. We just subsample the numpy array since averaging
-                    # the pixels would not be correct for OpenStreetMap.
-                    downscale_factor = subtile_size // desired_subtile_size
-                    image = image[:, ::downscale_factor, ::downscale_factor]
-                elif desired_subtile_size > subtile_size:
+                # If the modality does not vary in space (e.g., ERA5), we read the entire tile.
+                if not image_tile.modality.is_spatial:
```
favyen2, line 229: This might be cleaner as a shortcut case.

```
if not image_tile.modality.is_spatial:
    ...
    image = image.reshape(-1, len(band_set.bands))
    band_set_images.append(image)
    continue

# Assuming all tiles cover the same area as the resolution factor 16 tile
subtile_size = raster.width // factor
...
image = image.reshape(shape)
logger.debug(f"shape after scaling image.shape={image.shape}")
band_set_images.append(image)
```

helios/dataset_creation/rslearn_to_helios/era5.py
```diff
@@ -71,23 +70,18 @@ def convert_era5(window_path: UPath, helios_path: UPath) -> None:
...
-            raster_dir, adjusted_projection, adjusted_bounds
+            raster_dir, window.projection, window.bounds
         )
 
         year_images.append(image)
         year_time_ranges.append(time_range)
 
         # Should we use this image for the frequent data for this window?
         if (
-            window.time_range[0] < time_range[1]
```
favyen2, line 89: If this is because time_range is `tuple[datetime, datetime] | None` then I think it'd be better to add an assert (`e.g. assert window.time_range is not None`) rather than ignoring the type for this line.

helios/evals/linear_probe.py
```diff
@@ -18,7 +18,7 @@
 logger = getLogger(__name__)
 
 
-class ProbeType(StrEnum):
+class ProbeType(Enum):
```
favyen2: Should this be `ProbeType(str, Enum)`? Also what is the reason not to use StrEnum, do we support Python 3.10? With 3.11+ there should be StrEnum available.

helios/train/loss.py
```diff
@@ -189,6 +267,210 @@ def compute(
...
+
+            if len(losses) == 0:
+                # If no losses were computed, skip this modality
+                # logger.warning("No decoded values for this modality")
+                continue
+            loss = torch.stack(losses).mean()
+            total_loss += loss
+            total_decoded_modalities += 1
+
+        return self.weight * total_loss  # / total_decoded_modalities
```
favyen2: If total_decoded_modalities is not being used, then it can be removed.

yawenzzzz: Fixed.

helios/evals/linear_probe.py
yawenzzzz: Updated, yeah, we should just use 3.12.

helios/dataset_creation/rslearn_to_helios/era5.py
yawenzzzz: Thanks, fixed.

helios/dataset/sample.py
yawenzzzz: Thanks, fixed.

helios/dataset/convert_to_h5py.py
yawenzzzz: make sense, fixed!

helios/data/dataset.py
yawenzzzz: Yeah, I think separating them is better, thanks for the suggestion!

helios/data/dataset.py
yawenzzzz: Fixed, i moved it to the Dataset class.

yawenzzzz: This is also helpful when we apply CutMix to certain dataset, and not apply it on other dataset.

PR #323, opened 2025-08-14
separate generator for data loader params per rank/worker

pjreddie:


helios/data/dataloader.py
```diff
@@ -466,6 +436,49 @@ class _IterableDatasetWrapper(torch.utils.data.IterableDataset[HeliosSample]):
     def __init__(self, data_loader: HeliosDataLoader):
         """Initialize the IterableDatasetWrapper."""
         self.data_loader = data_loader
+        self.rng = get_rng(data_loader.seed + data_loader.epoch + data_loader.dp_rank)
```
Hgherzog: Nit: Instead of having a self.rng and self.rngs always have self.rngs and just have a list of length 1 with 0 workers.  I think this is clearer and easier to track

pjreddie: updated to always use list, lmk if you think it looks better

PR #326, opened 2025-08-19
Add Galileo to evals

gabrieltseng:
https://beaker.allen.ai/orgs/ai2/workspaces/earth-systems/work/01K31DS6SYKNY8NPXPF47KAD3Q?taskId=01K31DS6TMMJ85Y83Y5YFJFRAP&jobId=01K31DS6YGSQF0VFJ4PD3GBGGN

### DIFF
--- a/scripts/dino_v2_evals/galileo.py
+++ b/scripts/dino_v2_evals/galileo.py
@@ -94,30 +94,6 @@ def build_dataset_config(common: CommonComponents) -> HeliosDatasetConfig:
             training_modalities=common.training_modalities,
             dataset_percentage=common.dataset_percentage,
         ),
-        # # osm_sampling
-        # HeliosDatasetConfig(
-        #     h5py_dir="/weka/dfive-default/helios/dataset/osm_sampling/h5py_data_w_missing_timesteps_128_x_4_zstd_3/landsat_openstreetmap_raster_sentinel1_sentinel2_l2a_srtm_worldcover/1141152",
-        #     training_modalities=common.training_modalities,
-        #     dataset_percentage=common.dataset_percentage,
-        # ),
-        # # osmbig
-        # HeliosDatasetConfig(
-        #     h5py_dir="/weka/dfive-default/helios/dataset/osmbig/h5py_data_w_missing_timesteps_zstd_3_128_x_4/landsat_openstreetmap_raster_sentinel1_sentinel2_l2a_srtm_worldcover/1297928",
-        #     training_modalities=common.training_modalities,
-        #     dataset_percentage=common.dataset_percentage,
-        # ),
-        # # presto_neighbor
-        # HeliosDatasetConfig(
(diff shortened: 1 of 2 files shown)


helios/evals/models/__init__.py
```diff
@@ -1,6 +1,14 @@
...
+from helios.evals.models.galileo import GalileoConfig, GalileoWrapper
 from helios.evals.models.panopticon.panopticon import Panopticon, PanopticonConfig
 
-__all__ = ["DINOv2", "DINOv2Config", "Panopticon", "PanopticonConfig"]
+__all__ = [
+    "DINOv2",
+    "DINOv2Config",
+    "Panopticon",
+    "PanopticonConfig",
+    "GalileoWrapper",
```
Hgherzog, line 12: nit: Just calling this Galileo would be more consistent with how we are naming other models

scripts/dino_v2_evals/galileo.py
```diff
@@ -0,0 +1,131 @@
...
+def build_dataset_config(common: CommonComponents) -> HeliosDatasetConfig:
+    """Build the dataset config for an experiment."""
+    dataset_configs = [
+        # presto
+        HeliosDatasetConfig(
+            h5py_dir="/weka/dfive-default/helios/dataset/osm_sampling/h5py_data_w_missing_timesteps_zstd_3_128_x_4/cdl_gse_landsat_openstreetmap_raster_sentinel1_sentinel2_l2a_srtm_worldcover_worldpop/1141152",
+            training_modalities=common.training_modalities,
+            dataset_percentage=common.dataset_percentage,
+        ),
+        # # osm_sampling
```
Hgherzog: nit: remove commented code

scripts/dino_v2_evals/galileo_evals.py
```diff
@@ -0,0 +1,272 @@
+"""Launch script for evaluation.
```
Hgherzog, line 1: FYI we have a script to run and launch all evals and to sweep over the different params here https://github.com/allenai/helios/tree/main/scripts/run_all_evals . In my dinoV3 PR, i am making it compatible with non-helios models so you don't need to worry about that for now. We should move towards using those scripts instead of making tons of new eval scripts.

helios/evals/models/__init__.py
gabrieltseng: yes - I agree, but I used the new name since we already have "Galileo" littered around our codebase and I wanted to make this explicitly different from https://github.com/allenai/helios/blob/main/helios/nn/galileo.py

scripts/dino_v2_evals/galileo.py
gabrieltseng: updated

scripts/dino_v2_evals/galileo_evals.py
gabrieltseng: great. when the dinov3 PR is merged in (or perhaps as part of the dinov3 PR) we can delete the individual scripts.

PR #331, opened 2025-08-26
Henryh/fix dataset percentage bug

Hgherzog:
- Fix bug where the dataset percentage changes when job is pre-empted or restarted
- In the original results for dataset percentage pre-emption almost never happened so those results still hold. 

https://beaker.allen.ai/orgs/ai2/workspaces/earth-systems/work/01K3M5GZTFS597VVC28Z22AA9J?taskId=01K3M5GZV1S0KCAVM97T4QEDPC&jobId=01K3M5GZZ2R1MR059ZYCB76RWN

### DIFF
--- a/helios/data/dataloader.py
+++ b/helios/data/dataloader.py
@@ -134,15 +134,13 @@ class HeliosDataLoader(DataLoaderBase):
         assert len(self.dataset) < np.iinfo(np.uint32).max
 
         rng: np.random.Generator | None = None
-        data_percentage_rng: np.random.Generator | None = None
-        if self.dataset_percentage < 1.0:
-            # Dataset subset selected should be deterministic across random seeds and restarts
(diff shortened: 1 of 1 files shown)


helios/data/dataloader.py
```diff
@@ -132,12 +134,25 @@ def _build_global_indices(self) -> np.ndarray:
         assert len(self.dataset) < np.iinfo(np.uint32).max
 
         rng: np.random.Generator | None = None
+        data_percentage_rng: np.random.Generator | None = None
```
favyen2: This looks fine but I think it may be more clear if you just do the `if self.dataset_percentage < 1.0` check once below and there you can both initialize the `data_percentage_rng = get_rng(self.seed)` and the rest of the `indices` sub-sampling code. Instead of needing to check `self.dataset_percentage < 1.0` first and then `data_percentage_rng is not None` later.

PR #340, opened 2025-09-10
Script for phase 1 sweep

favyen2:


scripts/2025_09_10_phase1/launch.sh
```diff
@@ -0,0 +1,10 @@
+python scripts/2025_09_10_phase1/script.py launch base_default ai2/ceres-cirrascale
+python scripts/2025_09_10_phase1/script.py launch base_contrastive0.05 ai2/ceres-cirrascale --train_module.contrastive_config.loss_config.weight=0.05
+python scripts/2025_09_10_phase1/script.py launch base_lr2e-4 ai2/ceres-cirrascale --train_module.optim_config.lr=0.0002
+python scripts/2025_09_10_phase1/script.py launch base_encode0.25 ai2/ceres-cirrascale --train_module.masking_config.strategy_config.encode_ratio=0.25 --train_module.masking_config.strategy_config.decode_ratio=0.75
+# Encode ratio 0.75 needs smaller microbatch size.
```
Hgherzog, line 5: I think we should add a default run with just the batch size changed for fair comparison as this materially changes the contrastive loss.

Hgherzog: We can try to reduce memory consumption afterwards if there is a significant difference

favyen2: sgtm, added

scripts/2025_09_10_phase1/script.py
```diff
@@ -0,0 +1,297 @@
...
+        ),
+        contrastive_config=LossConfig(
+            loss_config={
+                "type": "InfoNCE",
+                "weight": 0.1,
+            }
+        ),
+        token_exit_cfg={modality: 0 for modality in common.training_modalities},
+        max_grad_norm=1.0,
+        scheduler=CosWithWarmup(warmup_steps=8000),
```
Hgherzog: I thought we agreed to default to constant with Warmup so we could experiment with decays at different points. See phase 1 of this doc https://docs.google.com/document/d/1O3zI0Z1VkWD0mkD2Jx81GagDRWbnZ9mJpuORzs6gwo8/edit?tab=t.5eqyhf2dma3u

scripts/2025_09_10_phase1/script.py
```diff
@@ -0,0 +1,297 @@
...
+            probe_batch_size=8,
+            num_workers=8,
+            pooling_type=PoolingType.MEAN,
+            norm_stats_from_pretrained=True,
+            probe_lr=0.1,
+            eval_interval=Duration.steps(20000),
+            input_modalities=[Modality.SENTINEL2_L2A.name],
+            epochs=50,
+        ),
+    }
```
Hgherzog, line 287: Maybe add so2sat every 10k and cashew plant every 50k steps as well. Also maybe add sickle sentinel1 and landsat every 20k steps

scripts/2025_09_10_phase1/script.py
favyen2: Good catch, fixed.

scripts/2025_09_10_phase1/script.py
```diff
@@ -0,0 +1,339 @@
...
+        seed=3622,
+    )
+
+
+def build_dataset_config(common: CommonComponents) -> HeliosDatasetConfig:
+    """Build the dataset config for an experiment."""
+    dataset_configs = [
+        HeliosDatasetConfig(
+            h5py_dir="/weka/dfive-default/helios/dataset/osm_sampling/h5py_data_w_missing_timesteps_zstd_3_128_x_4/cdl_gse_landsat_openstreetmap_raster_sentinel1_sentinel2_l2a_srtm_worldcereal_worldcover_worldpop_wri_canopy_height_map/1138828",
+            training_modalities=common.training_modalities,
```
gabrieltseng, line 177: this h5py folder is missing era5

favyen2: I think we decided not to use ERA5 for this sweep. Yawen/Henry are separately trying out one run that includes ERA5.

PR #342, opened 2025-09-11
Repro 2

Hgherzog:


helios/evals/datasets/normalize.py
Hgherzog, line 1: maybe this changes evla if there is a bug

helios/evals/linear_probe.py
```diff
@@ -127,9 +126,19 @@ def train_and_eval_probe(
     if train_embeddings.shape[-1] != test_embeddings.shape[-1]:
         raise ValueError("Embedding dims don't match.")
     in_features = train_embeddings.shape[-1]
-
+    output_pixels_per_side_of_patch = None
     if config.task_type == TaskType.SEGMENTATION:
-        logits_per_patch = int(config.num_classes * patch_size * patch_size)
+        assert config.height_width is not None, (
```
Hgherzog, line 131: This touches the pastis eval but I think it is right

PR #347, opened 2025-09-15
Henryh/inference throughput updates

Hgherzog:
- Updates #203 to fit within our existing launch cli 
- create an inference benchmark runner prototype that re-uses existing functionality
- Removes all gradient syncs when running inference
- only benchmarks model configs instead of actually loading

Grad syncs
1.  Fast path to not sort and recombine data based on masks
2. Adds non persistent buffer to index into tensors for bandsets. 

Example launch: https://beaker.allen.ai/orgs/ai2/workspaces/earth-systems/work/01K57QNXDZZ6CEDTHMEQYJMABV?taskId=01K57QNXEJRPE3R9A6KGPRYKC5&jobId=01K57QNXJYQMH59RYRWRDS56C6

### DIFF
--- a/helios/nn/pooled_modality_predictor.py
+++ b/helios/nn/pooled_modality_predictor.py
@@ -497,7 +497,7 @@ class EncodeEarlyAttnPool(Encoder):
             )
         cu_seqlens = get_cumulative_sequence_lengths(seq_lengths)
 
-        attn_mask = self.get_attn_or_none_mask(new_mask, fast_pass)
+        attn_mask = self._maybe_get_attn_mask(new_mask, fast_pass)
         # Apply attn with varying encoder depths
         for i_blk, blk in enumerate(self.blocks):
             if i_blk < self.num_pre_modality_pooling_layers:


helios/inference_benchmarking/constants.py
```diff
@@ -0,0 +1,95 @@
...
+    "batch_size": [1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192],
+}
+sweep_image_sizes = {
+    "image_size": [1, 2, 4, 8, 16, 32, 64, 128],
+}
+sweep_patch_sizes = {"patch_size": [1, 2, 4, 8]}
+sweep_num_timesteps = {"num_timesteps": [1, 2, 4, 6, 8, 12]}
+sweep_use_s1 = {"use_s1": [True, False]}
+sweep_use_s2 = {"use_s2": [True, False]}
+sweep_use_landsat = {"use_landsat": [True, False]}
```
cmwilhelm, line 71: I ran into some interesting stuff way back when I started working on this where I think landsat data was dropped immediately, so the tokens calculations were incorrect. I assume that's no longer the case?

helios/inference_benchmarking/run_throughput_benchmark.py
```diff
@@ -0,0 +1,497 @@
...
+                config=run_params.as_dict(),
+            )
+            wandb_callback.trainer = MinimalTrainer(device, self.work_dir)
+            callbacks.append(wandb_callback)
+
+        # Initialize profiler
+        for callback in callbacks:
+            callback.pre_train()
+
+            if run_params.use_s1:
```
cmwilhelm: wrong indentation here?

helios/inference_benchmarking/run_throughput_benchmark.py
```diff
@@ -0,0 +1,497 @@
...
+
+            num_s1_tokens = calculate_num_token_embeddings(results.sentinel1)
+            num_s2_tokens = calculate_num_token_embeddings(results.sentinel2_l2a)
+            num_landsat_tokens = calculate_num_token_embeddings(results.landsat)
+            tokens_processed_per_batch.append(
+                num_s1_tokens + num_s2_tokens + num_landsat_tokens
+            )
+        if device.type == "cuda":
+            torch.cuda.synchronize()
+        overall_time_taken = time.monotonic() - overall_start_time
```
cmwilhelm, line 442: what is the effect of the `torch.cuda.synchronize()` operations before and after the `while` loop here relative to the direct timing calculations from before?

helios/internal/experiment.py
```diff
@@ -157,6 +168,35 @@ def build_config(
...
+    return config
+
+
+def benchmark(config: BenchmarkExperimentConfig) -> None:
+    """Benchmark an experiment."""
+    runner = config.benchmark.build()
+    runner.run()
+
+
+def launch_benchmark(config: BenchmarkExperimentConfig) -> None:
```
cmwilhelm, line 195: thanks for this consolidation into common entrypoint

helios/inference_benchmarking/run_throughput_benchmark.py
Hgherzog: Previously, the gpu work could still be ongoing while the metrics were being logged, so we were just timing how long it took to launch all the kernels not for the work to finish. The difference became much more noticeable when I removed the existing gradient syncs.

cmwilhelm: would this be functionally equivalent to consuming the `results` tensor in some way?

Hgherzog: yes. For instance calling .item() on a tensor would result in a host device sync.

helios/inference_benchmarking/data_models.py
```diff
@@ -0,0 +1,165 @@
...
+
+from helios.inference_benchmarking import constants
+
+
+@dataclass
+class RunParams(Config):
+    """Defines the parameters for a throughput run."""
+
+    # TODO: Add a named constant for the default model size
+    model_size: str = "base"
```
pjreddie, line 17: idk if it really matters just for defaults but these days our "base" model is probably really "base_shallow_decoder"

helios/inference_benchmarking/run_throughput_benchmark.py
```diff
@@ -0,0 +1,497 @@
...
+        else:
+            dtype = torch.float32
+        callbacks = []
+        # insantiate callbacks
+        if run_params.profiler_enabled:
+            profiler = ProfilerCallback(
+                skip_first=0,  # Don't skip any steps
+                wait=0,  # Start profiling immediately
+                warmup=5,  # Warmup for 5 steps (matches your current warmup)
+                active=3,  # Profile for 10 steps
```
pjreddie: comment seems wrong

helios/inference_benchmarking/run_throughput_benchmark.py
```diff
@@ -0,0 +1,497 @@
...
+            # Set up a mock trainer object for the profiler
+
+            profiler.trainer = MinimalTrainer(device, self.work_dir)
+
+            callbacks.append(profiler)
+
+        if run_params.wandb_enabled:
+            project = os.getenv(constants.PARAM_KEYS["project"], constants.PROJECT_NAME)
+            owner = os.getenv(constants.PARAM_KEYS["owner"], constants.ENTITY_NAME)
+            # add a uuid to the name
```
pjreddie: same also can remove 278

helios/nn/flexihelios.py
```diff
@@ -1432,7 +1483,12 @@ def apply_attn(
         tokens = self.norm(tokens)
         # we don't care about the mask returned by add_removed_tokens, since we will
         # just use the original, unclipped mask here
-        tokens, _ = self.add_removed_tokens(tokens, indices, new_mask)
+        # TODO: Need this to be even more of a no-op
+        if not always_pass_none_mask_to_transformer:
+            tokens, _ = self.add_removed_tokens(tokens, indices, new_mask)
+        else:
```
pjreddie: why's this else case here?

helios/nn/flexihelios.py
```diff
@@ -1455,13 +1511,18 @@ def forward(
...
             token_exit_cfg: Configuration for token exit
-            always_pass_none_mask_to_transformer: Whether to always pass None as the mask to the transformer, this enables torch based flash attention
+            always_pass_none_mask_to_transformer: Whether to always pass None as the mask to the transformer, this enables torch based flash attention, and skips mask construciton and sorting
 
         Returns:
             TokensAndMasks containing the encoded representations and their masks
         """
-        # TODO: Add step to validate the exit config is valid
-        patchified_tokens_and_masks = self.patch_embeddings.forward(x, patch_size)
+        if always_pass_none_mask_to_transformer and token_exit_cfg is not None:
```
pjreddie: I don't love having the fast pass being interleaved in here I think it makes stuff kinda confusing but I guess it's not the end of the world. Ideally eventually we have the single-file inference helios I think it would be better to do benchmarking on that and remove this but. we can punt that

tests/integration/nn/test_flexihelios.py
```diff
@@ -567,6 +579,94 @@ def test_entire_modality_masked(
...
+            latlon_num_band_sets,
+            expected_embedding_size,
+        ), f"Expected output latlon shape {latlon.shape}, got {output.latlon.shape}"
+        assert output.latlon_mask.shape == (
+            B,
+            latlon_num_band_sets,
+        ), (
+            f"Expected output latlon_mask shape {latlon_mask.shape}, got {output.latlon_mask.shape}"
+        )
+
```
pjreddie, line 722: Could test the output is the same as the non-fast pass forward

helios/inference_benchmarking/data_models.py
Hgherzog: Because this only uses the encoder, I picked base as it is the same

helios/nn/flexihelios.py
Hgherzog: Would you prefer I hid it within the helper functions they affected as no-ops?

PR #348, opened 2025-09-17
Add Satlas to evals

gabrieltseng:
eurosat knn: 0.825 using pretrained norm
m_bigearthnet score: 0.4997446808510638 using pretrained norm
pastis S2 miou: 0.19 using pretrained norm and mean pooling with an lr of 0.0001 

This is all aligned with the Galileo results

### DIFF
--- a/helios/evals/models/satlas/satlas.py
+++ b/helios/evals/models/satlas/satlas.py
@@ -175,6 +175,10 @@ class Satlas(nn.Module):
         processed_inputs, modality = self.prepare_input(masked_helios_sample)
         outputs_list: list[torch.Tensor] = []
         for per_t_input in processed_inputs:
+            # we only take the last feature map from satlas for consistency with
+            # the other models. For segmentation tasks, multi scale feature maps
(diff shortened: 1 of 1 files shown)


helios/evals/models/satlas/satlas.py
```diff
@@ -0,0 +1,210 @@
...
+        """
+        super().__init__()
+        self.size = size
+        self.load_directory = UPath(load_directory)
+        # need to have some model at init so that the trainer can build correctly
+        supported_modalities = self.modality_size_to_weights[self.size].keys()
+        self.models = nn.ModuleDict(
+            {modality: self._load_model(modality) for modality in supported_modalities}
+        )
+        self.dim = 1024 if size == "base" else 768
```
yawenzzzz, line 85: We may want to add a note that only the last feature map is used, though for segmentation task usually multi-scale feature maps from swin is more preferred.

gabrieltseng: Added to line 178

PR #350, opened 2025-09-18
Add Prithvi v2.0 to evals

gabrieltseng:
eurosat knn using pre-trained norm: 0.762
m_bigearthnet score: 0.49755274261603377
m_so2sat score: 0.29918864097363085

Lower for eurosat but otherwise consistent with the Galileo benchmark results.

### DIFF
--- a/helios/evals/models/prithviv2/prithvi_mae.py
+++ b/helios/evals/models/prithviv2/prithvi_mae.py
@@ -440,7 +440,7 @@ class PrithviViT(nn.Module):
         # embed patches
         x = self.patch_embed(x)
 
-        pos_embed = self.interpolate_pos_encoding(sample_shape)
+        pos_embed = self.interpolate_pos_encoding(sample_shape).to(x.device)
         # add pos embed w/o cls token
         x = x + pos_embed[:, 1:, :]
 
@@ -483,7 +483,7 @@ class PrithviViT(nn.Module):
         # embed patches
         x = self.patch_embed(x)
 
-        pos_embed = self.interpolate_pos_encoding(sample_shape)
+        pos_embed = self.interpolate_pos_encoding(sample_shape).to(x.device)
         # add pos embed w/o cls token
         x = x + pos_embed[:, 1:, :]
 


helios/evals/models/prithvi/prithvi_mae.py
```diff
@@ -0,0 +1,759 @@
...
+            embed_dim: int = 768,
+            norm_layer: nn.Module | None = None,
+            flatten: bool = True,
+            bias: bool = True,
+    ):
+        super().__init__()
+        self.input_size = input_size
+        self.patch_size = patch_size
+        self.grid_size = [s // p for s, p in zip(self.input_size, self.patch_size)]
+        assert self.grid_size >= [1, 1, 1], "Patch size is bigger than input size."
```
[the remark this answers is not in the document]
gabrieltseng: Not fixing - this is from the original prithvi codebase

helios/evals/models/prithvi/prithvi.py
```diff
@@ -0,0 +1,190 @@
...
+        elif pooling == PoolingType.MAX:
+            output_features = torch.max(torch.cat(outputs_list, dim=0), dim=0)[0]
+        return output_features
+
+
+@dataclass
+class PrithviConfig(Config):
+    """olmo_core style config for Prithvi Wrapper."""
+
+    load_directory: str = "/weka/dfive-default/helios/models/prithvi"
```
Hgherzog: Is the model available on torchhub or huggingface? I think it is preferable to download from there when possible so it is easy for others to reproduce our eval suite.

helios/evals/models/prithvi/prithvi.py
```diff
@@ -0,0 +1,190 @@
...
+    2248.0,
+    2179.0,
+    2178.0,
+    1850.0,
+    1242.0,
+    1049.0,
+]
+
+
+class Prithvi(nn.Module):
```
Hgherzog: Suggestion: change name to prithvi v2 for clarity

helios/evals/models/prithvi/prithvi.py
Hgherzog: https://huggingface.co/ibm-nasa-geospatial/Prithvi-EO-2.0-300M/tree/main

helios/evals/models/prithvi/prithvi.py
```diff
@@ -0,0 +1,190 @@
...
+        self,
+        masked_helios_sample: MaskedHeliosSample,
+        pooling: PoolingType = PoolingType.MEAN,
+        spatial_pool: bool = False,
+    ) -> torch.Tensor:
+        """Forward pass through the satlas model."""
+        processed_inputs = self.prepare_input(masked_helios_sample)
+        outputs_list: list[torch.Tensor] = []
+        for per_t_input in processed_inputs:
+            output = self.model.forward_features(per_t_input)[-1]
```
Hgherzog: Doesn't Prithvi except multi-temporal input? So we should pass all as one input?

gabrieltseng: 🤦 thanks for catching this

helios/evals/models/prithvi/prithvi.py
gabrieltseng: added

helios/evals/models/prithvi/prithvi.py
gabrieltseng: updated, getting the same eurosat results (which means I didn't break things)

helios/evals/models/prithvi/prithvi.py
gabrieltseng: How much would you like it changed?
I can change the filenames too, but if you are happy with the classes being updated then that's done

Hgherzog: If everything can be consistent it would be prefered

PR #351, opened 2025-09-18
Add AnySat to evals

gabrieltseng:
KNN eurosat accuracy: 0.785 (the same norm stats should get 0.805 according to Galileo)
forestnet KNN: 0.3822042467138524
pastis s2 with the lowest lr: 0.4098317623138428

### DIFF
--- a/helios/evals/models/anysat/anysat.py
+++ b/helios/evals/models/anysat/anysat.py
@@ -66,7 +66,7 @@ class AnySat(nn.Module):
             "gastruc/anysat",
             "anysat",
             pretrained=True,
-            flash_attn=True,
+            flash_attn=False,
             force_reload=True,
         )
 
--- a/requirements-eval.txt
+++ b/requirements-eval.txt
@@ -1,6 +1,5 @@
 albumentations
 breizhcrops==0.0.4.1 # Force the latest available
-flash-attn
 geobench @ git+https://github.com/ServiceNow/geo-bench.git  # Unreleased relaxed dependencies
 geopandas
 rioxarray


helios/evals/embeddings.py
```diff
@@ -3,16 +3,18 @@
...
+from helios.evals.datasets.configs import TaskType
+from helios.evals.eval_wrapper import AnySatEvalWrapper, EvalWrapper
 from helios.train.masking import MaskedHeliosSample
 
 logger = logging.getLogger(__name__)
 
 
 def get_embeddings(
-    data_loader: DataLoader, model: EvalWrapper
+    data_loader: DataLoader, model: EvalWrapper, is_train: bool
```
Hgherzog: I strongly prefer that any model specific steps be pushed inside the model eval wrapper. I think doing this for the embedding creation is straightforward and for the labels you could add an extra method that is a no-op for everything else that you can override for anysat. For is_train consider, making it a state in the model_wrapper instead.

helios/evals/models/anysat/anysat.py
```diff
@@ -0,0 +1,239 @@
...
+    ]
+
+    def __init__(self) -> None:
+        """AnySat wrapper."""
+        super().__init__()
+        self.model = torch.hub.load(
+            "gastruc/anysat",
+            "anysat",
+            pretrained=True,
+            flash_attn=False,
```
Hgherzog: Would flash_attn True help simplify the memory issues or help these run faster

helios/evals/models/anysat/anysat.py
```diff
@@ -0,0 +1,239 @@
...
+                for v in self.ANYSAT_S1_BAND_ORDERING
+                if v in Modality.SENTINEL1.band_order
+            ],
+            Modality.LANDSAT.name: [
+                Modality.LANDSAT.band_order.index(v)
+                for v in self.ANYSAT_L8_BAND_ORDERING
+                if v in Modality.LANDSAT.band_order
+            ],
+        }
+        self.month = 5  # default month, if none is given (indexing from 0)
```
Hgherzog: Add a comment about how this default month was chosen or if it was copied from anysat github

helios/evals/models/anysat/anysat.py
```diff
@@ -0,0 +1,239 @@
...
+    @staticmethod
+    def _month_day_to_day_of_year(
+        months: torch.Tensor, days: torch.Tensor | None = None
+    ) -> torch.Tensor:
+        output_tensors = []
+        for i in range(months.shape[0]):
+            output_tensors.append(
+                torch.tensor(
+                    [datetime(2025, m + 1, 1).timetuple().tm_yday for m in months[i]]
+                ).to(device=months.device)
```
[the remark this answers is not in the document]
Hgherzog: +1

helios/evals/models/anysat/anysat.py
```diff
@@ -0,0 +1,239 @@
...
+                months = repeat(
+                    torch.tensor([self.month]).to(device=processed_data.device),
+                    "d -> b d",
+                    b=processed_data.shape[0],
+                )
+                doy = self._month_day_to_day_of_year(months=months)
+            else:
+                months = masked_helios_sample.timestamps[:, :, 1]
+                days = masked_helios_sample.timestamps[:, :, -1]
+                doy = self._month_day_to_day_of_year(months=months, days=days)
```
Hgherzog: Why not use the year here as well?

helios/evals/models/anysat/anysat.py
```diff
@@ -0,0 +1,239 @@
...
+from einops import rearrange, repeat
+from olmo_core.config import Config
+from torch import nn
+
+from helios.data.constants import Modality
+from helios.nn.flexihelios import PoolingType
+from helios.train.masking import MaskedHeliosSample
+
+logger = logging.getLogger(__name__)
+
```
Hgherzog, line 16: Does anysat not use any pre-trained normalization stats?

helios/evals/models/anysat/anysat.py
```diff
@@ -0,0 +1,239 @@
...
+        hs = []
+        for key, val in processed_inputs.items():
+            if not key.endswith("dates"):
+                hs.append(val.shape[-1])
+        if len(set(hs)) != 1:
+            raise RuntimeError("Expected all inputs to have the same dimension")
+        patch_size = self._calculate_patch_size(hs[0])
+
+        # from the README:
```
Hgherzog: Link to the readme

helios/evals/models/anysat/anysat.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def forward(
+        self,
+        masked_helios_sample: MaskedHeliosSample,
+        pooling: PoolingType = PoolingType.MEAN,
+        spatial_pool: bool = False,
+    ) -> torch.Tensor:
+        """Forward pass through the AnySat model."""
+        processed_inputs = self.prepare_input(masked_helios_sample)
+        if pooling == PoolingType.MAX:
```
Hgherzog, line 203: Anysat does some feature prep internally but we could still do max pooling vs mean pooling over there dense features for classification? Or is there another reason we can't do max?

helios/evals/embeddings.py
gabrieltseng: updated

helios/evals/models/anysat/anysat.py
gabrieltseng: No - https://github.com/gastruc/AnySat/issues/3#issuecomment-2571081394

they use the normalization stats from their (different) pretraining datasets, and don't provide a unified set of values.

helios/evals/models/anysat/anysat.py
gabrieltseng: This makes things finnicky (requires flash-attn which doesn't work nicely with cpu only) without significantly speeding things up (44 seconds on eurosat with, 46 seconds without).

It wouldn't handle the memory issues from the large tensors in memory (so the subsetting would still be necessary) so I think its okay to leave it as is.

helios/evals/models/anysat/anysat.py
gabrieltseng: anysat returns its own "tile" feature for classification, its not an aggregation of the dense features

helios/evals/models/anysat/anysat.py
gabrieltseng: added

helios/evals/models/anysat/anysat.py
gabrieltseng: there's no need for a default month actually since timestamps can't be None. Removed

helios/evals/models/anysat/anysat.py
gabrieltseng: updated to use the tessera function

PR #353, opened 2025-09-19
Add Copernicus FM to evals

gabrieltseng:
I am a bit unsure about the licensing of this code. 

Copernicus FM itself has an Apache license (https://github.com/zhu-xlab/Copernicus-FM/blob/main/LICENSE) but it uses a bunch of code which *doesn't* have an apache license, e.g. moco v3 code which has a noncommercial attribution license: https://github.com/zhu-xlab/Copernicus-FM/blob/5c162a83b00f41fe4964e736c54a27488ba2d555/Copernicus-FM/src/util/pos_embed.py#L18

Confirmed this works:
Eurosat knn score: 0.849
Pastis LP score: 0.2968265414237976

### DIFF
--- a/helios/evals/models/copernicusfm/copernicusfm.py
+++ b/helios/evals/models/copernicusfm/copernicusfm.py
@@ -91,52 +91,22 @@ class CopernicusFM(torch.nn.Module):
             state_dict = check_point
         self.model.load_state_dict(state_dict, strict=False)
 
-        self.modality_to_wavelength_bandwidths = {
-            Modality.SENTINEL2_L2A.name: {
-                "band_wavelengths": [
-                    MODALITY_TO_WAVELENGTH_BANDWIDTHS[Modality.SENTINEL2_L2A.name][
-                        "band_wavelengths"
-                    ][
-                        MODALITY_TO_WAVELENGTH_BANDWIDTHS[Modality.SENTINEL2_L2A.name][
-                            "band_names"
-                        ].index(b)
-                    ]
-                    for b in Modality.SENTINEL2_L2A.band_order
-                ],
-                "band_bandwidths": [
-                    MODALITY_TO_WAVELENGTH_BANDWIDTHS[Modality.SENTINEL2_L2A.name][
-                        "band_bandwidths"
-                    ][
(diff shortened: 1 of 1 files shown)


helios/evals/datasets/cropharvest.py
```diff
@@ -320,9 +320,9 @@ def __getitem__(self, idx: int) -> tuple[MaskedHeliosSample, torch.Tensor]:
         if Modality.SENTINEL2_L2A.name in self.input_modalities:
             input_dict[Modality.SENTINEL2_L2A.name] = torch.tensor(s2).float()
         if Modality.SENTINEL1.name in self.input_modalities:
-            input_dict[Modality.SENTINEL1.name] = torch.tensor(s2).float()
+            input_dict[Modality.SENTINEL1.name] = torch.tensor(s1).float()
         if Modality.SRTM.name in self.input_modalities:
-            input_dict[Modality.SRTM.name] = torch.tensor(s2).float()
+            input_dict[Modality.SRTM.name] = torch.tensor(srtm).float()
         if Modality.LATLON.name in self.input_modalities:
```
gabrieltseng, line 326: 😬

pjreddie: 🫢

helios/evals/models/copernicusfm/copernicusfm.py
```diff
@@ -0,0 +1,266 @@
...
+        check_point = torch.load(
+            UPath(load_directory) / "CopernicusFM_ViT_base_varlang_e100.pth"
+        )
+        if "model" in check_point:
+            state_dict = check_point["model"]
+        else:
+            state_dict = check_point
+        self.model.load_state_dict(state_dict, strict=False)
+
+        self.modality_to_wavelength_bandwidths = {
```
pjreddie: I have basically no idea what's happening here ha! sometimes I think the linter really kills readability

pjreddie: IMO maybe do this all above with the first dict you create so at least there's only one weird dict lying around? idk

gabrieltseng: Turns out it was totally avoidable. Hopefully its a bit clearer now? I also added a comment explaining why we do this

PR #356, opened 2025-09-22
Henryh/add tessera

Hgherzog:
Adds Tessera to the repo

Tessera only accepts pixel time series and tasks that have both s1 and s2. This means it only works for crop harvest and pastis_sentinel1_sentinel2.

### DIFF
--- a/helios/evals/eval_wrapper.py
+++ b/helios/evals/eval_wrapper.py
@@ -171,6 +171,7 @@ class PrithviV2EvalWrapper(EvalWrapper):
 
 class DINOv2EvalWrapper(EvalWrapper):
     """Wrapper for DINOv2 models."""
+
     def __call__(self, masked_helios_sample: MaskedHeliosSample) -> torch.Tensor:
         """Forward pass through the model produces the embedding specified by initialization."""
         # i need to do the apply imagenet normalizer thing in here


helios/evals/datasets/configs.py
```diff
@@ -32,6 +32,7 @@ class EvalDatasetConfig(NamedTuple):
     # this is only necessary for segmentation tasks,
     # and defines the input / output height width.
     height_width: int | None = None
+    timeseries: bool = False
```
gabrieltseng, line 35: Can we call this `pixel_timeseries`? Pastis is time series too but doesn't get `True`

helios/evals/models/tessera/tessera.py
```diff
@@ -0,0 +1,263 @@
...
+        """Load the Tessera model."""
+
+        # TODO: Implement actual Tessera model loading
+        model = build_inference_model()
+        state_dict = torch.load(checkpoint_path, map_location="cpu")['model_state_dict']
+        # remove the _orig_mod prefix from all the keys
+        state_dict = {k.replace("_orig_mod.", ""): v for k, v in state_dict.items()}
+        # remove all keys starting with "projector."
+        state_dict = {k: v for k, v in state_dict.items() if not k.startswith("projector.")}
+        # remove all "fusion_method." prefix from the keys
```
gabrieltseng: nit - this comment seems to not happen? (or I'd expect a line saying `state_dict = {k: v for k, v in state_dict.items() if not k.startswith("fusion_method.")}`

helios/evals/models/tessera/tessera.py
```diff
@@ -0,0 +1,263 @@
...
+        model.load_state_dict(state_dict)
+        logger.info(f"Loaded Tessera model from {checkpoint_path}")
+        self.model = model
+
+
+    def calculate_day_of_year(self, timestamp: torch.Tensor) -> torch.Tensor:
```
gabrieltseng, line 120: This function can be simplified using the `tm_yday` attribute: https://stackoverflow.com/a/623312

helios/evals/models/tessera/tessera.py
```diff
@@ -0,0 +1,263 @@
...
+        if (s2_x := masked_helios_sample.sentinel2_l2a) is not None:
+            # Steps filter to the bands and reshape to (b, h, w, t, c)
+            s2_x = s2_x[..., HELIOS_SENTINEL2_TESSERA_BANDS]
+            if self.use_pretrained_normalizer:
+                s2_x = self.standardize(s2_x, torch.tensor(S2_BAND_MEAN, device=s2_x.device), torch.tensor(S2_BAND_STD, device=s2_x.device))
+                # standardize the s2_x
+            # get day of year
+            doy = self.calculate_day_of_year(masked_helios_sample.timestamps)
+            # concatenate as an extra band at every HW as the last band use einops to repeat
+            doy = repeat(doy, "b t  -> b  h w t 1", h=s2_x.shape[1], w=s2_x.shape[2])
```
gabrieltseng: Since we are bound to have s1 and s2, the `doy` computations could be done once at the very beginning.

helios/evals/models/tessera/tessera.py
```diff
@@ -0,0 +1,263 @@
...
+            if self.use_pretrained_normalizer:
+                s1_x = self.standardize(s1_x, torch.tensor(S1_BAND_MEAN, device=s1_x.device), torch.tensor(S1_BAND_STD, device=s1_x.device))
+                # standardize the s1_x
+            # get day of year
+            doy = self.calculate_day_of_year(masked_helios_sample.timestamps)
+            # concatenate as an extra band at every HW as the last band use einops to repeat
+            doy = repeat(doy, "b t  -> b h w t 1", h=s1_x.shape[1], w=s1_x.shape[2])
+            s1_x = torch.cat([s1_x, doy], dim=-1)
+        if s2_x is None or s1_x is None:
+            raise ValueError("Tessera does not support single modality input")
```
gabrieltseng, line 176: in the spirit of failing fast, this should probably be the first thing this function checks for

helios/evals/models/tessera/tessera.py
```diff
@@ -0,0 +1,263 @@
...
+        output_features = torch.zeros(*output_shape, device=s2_x.device)
+
+        for i,j in product(range(h), range(w)):
+            if s2_x is not None:
+                s2_x_ij = s2_x[:, i, j, :, :]
+            else:
+                s2_x_ij = None
+            if s1_x is not None:
+                s1_x_ij = s1_x[:, i, j, :, :]
+            else:
```
gabrieltseng: I don't think this `if/else` logic is not necessary? Since in `prepare_input` we trigger a ValueError if `s2_x` or `s1_x` is None

helios/evals/models/tessera/tessera.py
```diff
@@ -0,0 +1,263 @@
...
+            output_features_ij = self.model(s2_x_ij, s1_x_ij)
+            output_features[:, i, j, :] = output_features_ij
+        # Forward pass through Tessera model
+        # Loop over H and W as well to get the output feature map
+        logger.info(f"Output features shape: {output_features.shape}")
+
+        output_features = reduce(output_features, "b ... d -> b d", pooling)
+        return output_features
+
+    def forward_features(
```
gabrieltseng: Is this function necessary? Where is it called?

gabrieltseng: (I'm guessing it should be removed based on the comments / implementation)

helios/internal/full_eval_sweep.py
```diff
@@ -234,6 +261,8 @@ def _get_normalization_args(args: argparse.Namespace, norm_mode: str) -> str:
             return get_galileo_args(pretrained_normalizer=False)
         elif norm_mode == "pre_trained":
             return get_galileo_args(pretrained_normalizer=True)
+    elif args.tessera:
+        return get_tessera_args()
```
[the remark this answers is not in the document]
gabrieltseng: +1

helios/evals/models/tessera/tessera.py
Hgherzog: Because of the batching and the fact there are 12 timestamps often for each batch, I think it actually clearer and more efficient to do it this way especially because we will use large batch sizes for the single pixel stuff.

helios/evals/models/tessera/tessera.py
Hgherzog: yea, i realize I had some vestiges of when i thought tessera could handle a modality being missing at inference time

helios/evals/datasets/configs.py
Hgherzog: I did give Pastis True. I think it is reasonable to evaluate pixel time series models on pastis because it is conceivable they could do well with just the temporal attention. Let me know if you disagree.

gabrieltseng: oh it was hidden by the collapsed lines. okay this is fine & makes sense to me

PR #365, opened 2025-09-30
Increase test tolerances

gabrieltseng:
Increase test tolerances to make them less brittle

### DIFF
--- a/tests/integration/train/train_module/test_contrastive_latentmim.py
+++ b/tests/integration/train/train_module/test_contrastive_latentmim.py
@@ -20,6 +20,8 @@ from helios.train.train_module.contrastive_latentmim import (
     ContrastiveLatentMIMTrainModuleConfig,
 )
 
+from .helper import check_loss_is_a_reasonable_value
+
 torch.set_default_device("cpu")
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


tests/integration/train/train_module/test_contrastive_latentmim.py
```diff
@@ -208,8 +208,8 @@ def test_train_batch_with_missing_modalities(
         logger.info(mock_trainer._metrics)
         assert torch.allclose(
             mock_trainer._metrics["train/PatchDisc"],
-            torch.tensor(2.2),
```
favyen2, line 211: Can you add some comment about where the 2.1 comes from? I looked at the samples_with_missing_modalities fixture but it's not clear to me how to compute 2.1 from those values. Or maybe Henry can review this PR instead since he may be more familiar with the code here.

Hgherzog: It was just the loss computed from the random inputs, which is dependent on all the  randomness being properly seeded. The purpose was to catch anything that creates non sensical outputs like nan or something really high for our different train modules, and make sure that all the apis connections are happy for the different train modules. It seems like this broke on something we don't care about. I would suggest just checking like < 4 and is not nan or inf.

gabrieltseng: Sounds good. Should I update the other tests while I'm at it?

gabrieltseng: Updated

tests/integration/train/train_module/test_contrastive_latentmim.py
```diff
@@ -174,16 +174,16 @@ def test_train_batch_without_missing_modalities(
...
-            torch.tensor(2.14),
-            atol=1e-1,
-        )
-        assert torch.allclose(
-            mock_trainer._metrics["train/InfoNCE"],
-            torch.tensor(0.196),
-            atol=1e-1,
-        )
+        loss = mock_trainer._metrics["train/PatchDisc"]
+        assert not torch.isinf(loss).any()
```
Hgherzog: nit: use a helper that checks these conditions in a single line assert

Hgherzog: re-use across tests

gabrieltseng: Added

PR #366, opened 2025-09-30
Log test results

gabrieltseng:
Running here: https://beaker.allen.ai/orgs/ai2/workspaces/earth-systems/work/01K6DSDD7876MSVB3G4EC2XFJD?taskId=01K6DSDD7PJTBHAR4KZ010SQF4&jobId=01K6DSDDB9Z577WYWG2B1PGJG9

I've checked that all eval datasets support a "test" split

### DIFF
--- a/helios/internal/all_evals.py
+++ b/helios/internal/all_evals.py
@@ -382,7 +382,7 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
                 tasks=EVAL_TASKS,
                 eval_on_startup=True,
                 cancel_after_first_eval=True,
-                run_on_test=True,
+                run_on_test=False,
             ),
         )
         .with_callback("garbage_collector", garbage_collector_callback)
--- a/helios/train/callbacks/evaluator_callback.py
+++ b/helios/train/callbacks/evaluator_callback.py
@@ -216,12 +216,14 @@ class DownstreamEvaluator:
             f"train embeddings shape for {self.dataset}: {train_embeddings.shape}"
         )
         logger.info(f"val embeddings shape for {self.dataset}: {val_embeddings.shape}")
-        logger.info(
-            f"test embeddings shape for {self.dataset}: {test_embeddings.shape}"
-        )
+        if test_embeddings is not None:
+            logger.info(
+                f"test embeddings shape for {self.dataset}: {test_embeddings.shape}"
+            )
         logger.info(f"train labels shape for {self.dataset}: {train_labels.shape}")
         logger.info(f"val labels shape for {self.dataset}: {val_labels.shape}")
-        logger.info(f"test labels shape for {self.dataset}: {test_labels.shape}")
+        if test_labels is not None:
+            logger.info(f"test labels shape for {self.dataset}: {test_labels.shape}")
 
         kwargs = {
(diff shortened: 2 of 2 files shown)


helios/train/callbacks/evaluator_callback.py
```diff
@@ -50,6 +50,7 @@ class DownstreamTaskConfig:
     patch_size: int = 4
     probe_batch_size: int = 32
     epochs: int = 50  # Number of training epochs for linear probing task
+    linear_probe_eval_interval: int = 50  # calculate val results every N epochs
     eval_interval: Duration = field(default_factory=lambda: Duration.epochs(1))
```
gabrieltseng, line 54: I think we were overloading what `eval_interval` meant before. It meant both (i) how often we ran each eval, and (ii) how often the validation mious were computed during the linear probe. I've separated them out, and only have us computing the val / test miou results once now.

helios/train/callbacks/evaluator_callback.py
```diff
@@ -303,11 +312,14 @@ def pre_train(self) -> None:
                         f"Skipping {evaluator.evaluation_name} because it doesn't match input requirements of the model"
                     )
                     continue
-                val_result, eval_time = self._perform_eval(evaluator)
+                val_result, test_result, eval_time = self._perform_eval(evaluator)
                 if wandb_callback.enabled:
                     wandb_callback.wandb.log(
                         {"eval/" + evaluator.evaluation_name: val_result}
                     )
+                    wandb_callback.wandb.log(
```
Hgherzog: Suggestion: Add a way to configure whether we should log the test result so as to prevent peaking during our model selection

helios/train/callbacks/evaluator_callback.py
```diff
@@ -328,18 +340,23 @@ def post_step(self) -> None:
...
 
-    def _perform_eval(self, evaluator: DownstreamEvaluator) -> tuple[float, float]:
+    def _perform_eval(
+        self, evaluator: DownstreamEvaluator
+    ) -> tuple[float, float, float]:
         """Run the evaluator."""
         logger.info(f"Running {evaluator.evaluation_name} evaluations...")
         start_time = time.monotonic()
-        val_result = evaluator.val()
+        val_result, test_result = evaluator.val()
```
Hgherzog, line 362: Suggestion: Make running the test set and logging it configurable for in loop

helios/train/callbacks/evaluator_callback.py
gabrieltseng: Added

helios/train/callbacks/evaluator_callback.py
gabrieltseng: Added

helios/internal/all_evals.py
```diff
@@ -382,6 +382,7 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
                 tasks=EVAL_TASKS,
                 eval_on_startup=True,
                 cancel_after_first_eval=True,
+                run_on_test=False,
```
gabrieltseng, line 385: cc @Hgherzog  by default it won't run on the test data

PR #368, opened 2025-09-30
Add script to compute min/max stats for non-geobench eval datasets

Hgherzog:
- Created compute_eval_dataset_minmax.py to compute min/max per band for all modalities
- Processes all non-geobench datasets: mados, sen1floods11, pastis, pastis128, breizhcrops, sickle, cropharvest
- Added --max-samples CLI flag for smoke testing
- Output saved to helios/evals/datasets/minmax_stats.json

Panopticon ran here 
https://beaker.allen.ai/orgs/ai2/workspaces/earth-systems/work/01K6G7Z2BNM7XR8GQX82NFE3C2?taskId=01K6G7Z2C1FY0CRWEK9V3NDG79&jobId=01K6G7Z2FHAVPWW7JF998S1V8W

### DIFF
--- a/helios/evals/datasets/config/minmax_stats.json
+++ b/helios/evals/datasets/config/minmax_stats.json
@@ -58,64 +58,64 @@
   "cropharvest": {
     "sentinel1": {
       "vh": {
-        "max": 14123.0,
-        "min": 35.0
+        "max": 7.333084128492677,
+        "min": -66.72575349656982
       },
       "vv": {
-        "max": 22525.0,
-        "min": 70.0
+        "max": 18.229049457558652,
+        "min": -55.960540925721425
       }
     },
     "sentinel2_l2a": {
       "02 - Blue": {
-        "max": 18.229049457558652,
-        "min": -55.960540925721425
-      },
-      "03 - Green": {
-        "max": 7.333084128492677,
-        "min": -66.72575349656982
-      },
-      "04 - Red": {
         "max": 17634.0,
         "min": 436.0
       },
-      "05 - Vegetation Red Edge": {
+      "03 - Green": {
         "max": 17622.0,
         "min": 272.0
       },
-      "06 - Vegetation Red Edge": {
+      "04 - Red": {
         "max": 20703.0,
         "min": 156.0
       },
-      "07 - Vegetation Red Edge": {
+      "05 - Vegetation Red Edge": {
         "max": 21430.0,
         "min": 1.0
       },
-      "08 - NIR": {
+      "06 - Vegetation Red Edge": {
         "max": 21864.0,
         "min": 1.0
       },
-      "08A - Vegetation Red Edge": {
+      "07 - Vegetation Red Edge": {
         "max": 22324.0,
         "min": 99.0
       },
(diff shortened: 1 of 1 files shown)


helios/evals/datasets/utils.py
```diff
@@ -17,3 +19,9 @@ def eval_collate_fn(
     collated_sample = default_collate([s.as_dict(return_none=False) for s in samples])
     collated_target = default_collate([t for t in targets])
     return MaskedHeliosSample(**collated_sample), collated_target
+
+
+def load_min_max_stats() -> dict:
```
gabrieltseng, line 26: nit: currently, this dict gets loaded for every eval dataset. Since its pretty small do we want to just load it once and keep it in memory?

scripts/compute_eval_dataset_minmax.py
```diff
@@ -0,0 +1,607 @@
...
+                        )
+
+            total_samples += array.shape[0]
+        except Exception as e:
+            print(f"    Warning: Failed to process {dataset.id}: {e}")
+            continue
+
+    # Map to actual band names based on CropHarvest structure
+    # First 9 are S2 bands (without some bands like coastal aerosol, cirrus)
+    # Then 2 S1 bands
```
gabrieltseng, line 461: this is mixed up. the first 2 bands are S1, then followed S2 (this is why the values in the config are wonky for S1 and S2)

helios/evals/datasets/config/minmax_stats.json
```diff
@@ -0,0 +1,430 @@
...
+      },
+      "vv": {
+        "max": 31.7578125,
+        "min": -32.25
+      }
+    },
+    "sentinel2_l2a": {
+      "01 - Coastal aerosol": {
+        "max": 17369.0,
+        "min": -1960.0
```
gabrieltseng, line 192: noting I find it weird that these values are negative. S2 is supposed to be an unsigned int so I am not sure how pastis ended up with negative values

gabrieltseng: Maybe this shouldn't block this merge (I couldn't spot any issues in the min max calculations) but maybe this indicates we should take another look at pastis. 

It might be the atmospheric correction algorithm they use?

gabrieltseng: Okay there was some processing - from the [pastis](https://github.com/VSainteuf/pastis-benchmark) readme:

> The satellite imagery used in PASTIS was retrieved from [THEIA](https://github.com/VSainteuf/pastis-benchmark/blob/main/www.theia.land.fr): "Value-added data processed by the CNES for the Theia [www.theia.land.fr](http://www.theia.land.fr/) data cluster using Copernicus data. The treatments use algorithms developed by Theia’s Scientific Expertise Centres. "

scripts/compute_eval_dataset_minmax.py
Hgherzog: Thanks for catching this!

Hgherzog: Should be fixed now.

PR #381, opened 2025-10-10
Henryh/remove straggler

Hgherzog:
- Remove leftover helios code
- Add missing imports to all 
- change paths back to helios for the launch scripts. We will need to remove this dependency before OS https://github.com/allenai/helios/issues/382 

Here is confirmation we can still load other models without this code: https://wandb.ai/eai-ai2/helios_in_loop_evals/runs/or7uzmij 

I confirmed here that checkpoints load and helios eval process https://wandb.ai/eai-ai2/helios_in_loop_evals/runs/8x7hlart/logs

### DIFF
--- a/olmoearth_pretrain/evals/models/anysat/anysat_launch.py
+++ b/olmoearth_pretrain/evals/models/anysat/anysat_launch.py
@@ -92,7 +92,7 @@ def build_dataset_config(common: CommonComponents) -> OlmoEarthDatasetConfig:
     dataset_configs = [
         # ENsure this is any existing dataset
         OlmoEarthDatasetConfig(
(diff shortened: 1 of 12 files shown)


olmoearth_pretrain/train/callbacks/__init__.py
```diff
@@ -1,11 +1,13 @@
 """Callbacks for the trainer specific to OlmoEarth Pretrain."""
 
 from .evaluator_callback import DownstreamEvaluatorCallbackConfig
-from .speed_monitor import OlmoEarthSpeedMonitorCallback
-from .wandb import OlmoEarthWandBCallback
+from .speed_monitor import HeliosSpeedMonitorCallback, OlmoEarthSpeedMonitorCallback
```
favyen2, line 4: can these be renamed?

Hgherzog: No because they are in the checkpoint config.json, so we have aliases for now with a deprecation warning

PR #387, opened 2025-10-14
revert paths to weka files

Hgherzog:
Files on weka are still under the helios/ directory. I don't see any reason to move stuff internally on weka right now  so I am renaming here

### DIFF
--- a/olmoearth_pretrain/evals/models/galileo/single_file_galileo.py
+++ b/olmoearth_pretrain/evals/models/galileo/single_file_galileo.py
@@ -1731,7 +1731,7 @@ class GalileoWrapper(nn.Module):
         ]
         self.add_layernorm_on_exit = add_layernorm_on_exit
         if use_pretrained_normalizer:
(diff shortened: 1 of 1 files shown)


olmoearth_pretrain/evals/models/galileo/single_file_galileo.py
```diff
@@ -1731,7 +1731,7 @@ def __init__(
         ]
         self.add_layernorm_on_exit = add_layernorm_on_exit
         if use_pretrained_normalizer:
-            self.normalizer = Normalizer(std=True, normalizing_dicts=load_normalization_values(Path("/weka/dfive-default/henryh/helios/helios/olmoearth_pretrain/evals/models/galileo/normalization_config.json")))
+            self.normalizer = Normalizer(std=True, normalizing_dicts=load_normalization_values(Path("/weka/dfive-default/henryh/helios/helios/helios/evals/models/galileo/normalization_config.json")))
```
yawenzzzz: for the normalization file, you can use `Path(__file__).resolve().parent / "normalization_config.json")`, and `from pathlib import Path`, it works for me.

Hgherzog: Good catch that is much better.

PR #398, opened 2025-10-21
More changes for finetune

yawenzzzz:
Changes required to run finetune for different models and tasks.

### DIFF
--- a/olmoearth_pretrain/evals/finetune.py
+++ b/olmoearth_pretrain/evals/finetune.py
@@ -43,6 +43,7 @@ class BackboneWithHead(nn.Module):
             pooling_type=pooling_type,
             concat_features=False,
             use_pooled_tokens=use_pooled_tokens,
+            # Set this to False to avoid downsampling embeddings and labels for AnySat
             is_train=False,
         )
         self.task_type = task_type
@@ -68,8 +69,6 @@ class BackboneWithHead(nn.Module):
     ) -> torch.Tensor:
         """Forward pass through the model and head."""
         dev = next(self.wrapper.parameters()).device
-        # The wrapper requires both batch and labels as input, mainly for model
-        # like AnySat that modify the shape of labels during training
         emb, labels = self.wrapper(batch, labels)
         emb = cast(torch.Tensor, emb)
         emb_dim = emb.shape[-1]
@@ -276,7 +275,6 @@ def run_finetune_eval(
                 max_lr=lr,
                 min_lr=1.0e-6,
             )
-            # torch.nn.utils.clip_grad_norm_(ft.parameters(), 1.0)
             opt.step()
             opt.zero_grad()
 


scripts/get_max_eval_metrics_from_wandb.py
```diff
@@ -336,7 +336,7 @@ def save_metrics_to_csv(metrics_dict: dict[str, dict[str, float]], filename: str
                 print(f"\n{partition}:")
                 for metric in METRICS:
                     # Try original name
-                    key = f"eval/{metric}"
+                    key = f"eval/test/{metric}"
```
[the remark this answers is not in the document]
yawenzzzz: Fixed.

olmoearth_pretrain/internal/all_evals.py
```diff
@@ -418,120 +419,82 @@ def load_user_module(path: str) -> Any:
         norm_stats_from_pretrained=True,
         epochs=50,
     ),
+    # default: 32, galileo & anysat: 16, change patch size
+    # 256 * 256, limit to 32 * 32
+    # 24,576, 10 classes, segmentation
+    # anysat: ps = 8, bs = 8
+    # galileo: ps = 8, bs = 1
```
yawenzzzz: Move these final configs into the finetune script to override model and task settings.

olmoearth_pretrain/internal/full_eval_sweep_finetune.py
```diff
@@ -22,7 +22,7 @@
 logger = getLogger(__name__)
 
 # Fine-tune learning rates to sweep over.
-FT_LRS = [1e-5, 5e-5, 1e-4, 5e-4, 1e-3, 5e-3, 1e-2]
```
yawenzzzz, line 26: Tried smaller lr but not working well, for most models, the best finetune run is with lr within 1e-5 to 1e-4.

olmoearth_pretrain/internal/all_evals.py
yawenzzzz: Fixed.

olmoearth_pretrain/internal/full_eval_sweep_finetune.py
yawenzzzz: Fixed.

olmoearth_pretrain/evals/models/panopticon/panopticon.py
```diff
@@ -84,6 +84,7 @@ def _process_modality_data(self, data: torch.Tensor) -> list[torch.Tensor]:
         for i in range(t_dim):
             data_i = rearrange(data[:, :, :, i, :], "b h w c -> b c h w")
 
+            # TODO: 224 should not be hardcoded here
```
favyen2: It should be quick to add another one like `patch_size: int = 14` to the field so `input_size: int = 224`.

olmoearth_pretrain/evals/finetune.py
```diff
@@ -104,7 +105,7 @@ def _eval_cls(
         label = label.to(device=device)
         masked = _to_device(masked, device)
         with torch.amp.autocast(device_type=device.type, dtype=torch.bfloat16):
-            logits = module(masked, label)  # (B, C)
+            logits, label = module(masked, label)  # (B, C)
```
favyen2: It may be more clear to do `logits, _ = module(masked, label)` to make it clear that the model should not be changing the label during testing, only during training. And/or add a comment here since it is not clear what is going on without going to the get_eval_wrapper code.

olmoearth_pretrain/evals/models/panopticon/panopticon.py
yawenzzzz: Fixed.

olmoearth_pretrain/internal/all_evals.py
```diff
@@ -425,114 +426,59 @@ def load_user_module(path: str) -> Any:
         num_workers=2,
         pooling_type=PoolingType.MEAN,
         norm_stats_from_pretrained=False,
+        norm_method=NormMethod.NORM_NO_CLIP_2_STD,
```
Hgherzog, line 433: Can you change this for KNN and linear probe as well

olmoearth_pretrain/internal/full_eval_sweep_finetune.py
```diff
@@ -54,13 +68,14 @@ class ModelPreset:
...
     global_args: tuple[str, ...] = ()
-    # By default, we include the dataset stats.
     include_dataset_stats: bool = True
     launch_script_key: str | None = None
+    # Whether this model supports --model.use_pretrained_normalizer
+    supports_pretrained_normalizer: bool = False
 
 
-# TODO: We need to add the final norm settings for each model.
 MODEL_PRESETS: dict[str, ModelPreset] = {
```
Hgherzog, line 79: Ideally we should share these constants between the knn and LP and the full eval sweeps. I suggest making an issue and we do that in a seperate PR

olmoearth_pretrain/evals/finetune.py
yawenzzzz: make sense, fixed.

olmoearth_pretrain/internal/all_evals.py
yawenzzzz: Sure!

olmoearth_pretrain/internal/full_eval_sweep_finetune.py
yawenzzzz: agree, will add an issue on this.

PR #400, opened 2025-10-21
Renames in dataset creation code.

favyen2:
This PR just renames things, will add tutorial and move Beaker stuff around in a separate PR.


olmoearth_pretrain/dataset_creation/README.md
```diff
@@ -42,7 +42,7 @@ parallelize the materialize commands.
 For Sentinel-1:
 
     # Set DATASET_PATH to the location of your rslearn dataset from the previous step.
-    export DATASET_PATH=/weka/dfive-default/helios/dataset_creation/X/
+    export DATASET_PATH=/weka/dfive-default/olmoearth_pretrain/dataset_creation/X/
```
yawenzzzz, line 45: Do we want to update the weka or gcs path as well?

favyen2: I made a symlink so the WEKA path should be okay. The GCS path doesn't exist but I will change it in the next PR to be a placeholder rather than a specific GCS bucket.

PR #409, opened 2025-10-23
Henryh/de spookify masking

Hgherzog:
Hopefully cleaning up the confusing masking code so that it is more readable.

### DIFF
--- a/olmoearth_pretrain/train/masking.py
+++ b/olmoearth_pretrain/train/masking.py
@@ -1043,7 +1043,6 @@ class ModalityCrossMaskingStrategy(MaskingStrategy):
         num_encodable = len(encodable)
         upper_limit = num_encodable
         if not self.allow_encoding_decoding_same_bandset:
-
             upper_limit -= 1
 
         if self.max_encoded_bandsets is None:
@@ -1088,7 +1087,9 @@ class ModalityCrossMaskingStrategy(MaskingStrategy):
 
         num_decoded_bandsets = np.random.randint(min_decoded, max_decoded + 1)
         decoded_idxs = np.random.choice(
-            len(available), size=min(num_decoded_bandsets, len(available)), replace=False
+            len(available),
+            size=min(num_decoded_bandsets, len(available)),
+            replace=False,
         )
         return set([available[i] for i in decoded_idxs])
 


olmoearth_pretrain/train/masking.py
```diff
@@ -906,6 +924,10 @@ def __init__(
         self.max_decoded_bandsets = max_decoded_bandsets
         self.only_decode_modalities = only_decode_modalities
 
+    # =================================================================
+    # PUBLIC API
```
pjreddie: ?

olmoearth_pretrain/train/masking.py
```diff
@@ -959,119 +981,147 @@ def get_sample_present_modalities_bandsets(
...
-    def overide_strategy_mask(self, modality_spec: ModalitySpec) -> bool:
-        """Overide the mask for a modality depending on the strategy being modality cross masked.
+    def _select_bandsets_for_sample(
+        self, present: list[tuple[str, int]]
+    ) -> tuple[set[tuple[str, int]], set[tuple[str, int]]]:
+        """Select encoded and decoded bandsets for a single sample based on modality count."""
+        if len(present) == 1:
+            return self._select_bandsets_single_modality(present)
+        elif len(present) == 2:
+            return self._select_bandsets_two_modalities(present)
```
pjreddie: if `allow_encoding_decoding_same_bandset` is true (which it usually is) we don't have to do this. And if it's false will this ever work with 1 bandset? encoding the bandset and not decoding anything may give us errors in a weird edge case where this was true for a bunch of samples and we generated no loss but maybe that's unlikely.

also the 2 modality case should probably randomly select which bandset to encode vs decode instead of having static assignment unless they get shuffled earlier which they might?

olmoearth_pretrain/train/masking.py
Hgherzog: I asked cursor to comment organizing methods I can delete this

Hgherzog: happy to delete if unhelpful

olmoearth_pretrain/train/masking.py
Hgherzog: This makes sense. I will try to either remove or simplify this. If `allow_encoding_decoding_same_bandset=False`  and there is only 1 present modality bandset  that should just encode and not decode because it is safer.

olmoearth_pretrain/train/masking.py
```diff
@@ -959,119 +981,131 @@ def get_sample_present_modalities_bandsets(
...
+
+    def _randomly_select_decoded_bandsets(
+        self,
+        present: list[tuple[str, int]],
+        encoded: set[tuple[str, int]],
+        min_max_decode: tuple[int, int],
+    ) -> set[tuple[str, int]]:
+        """Randomly select which bandsets to decode."""
+        min_decoded, max_decoded = min_max_decode
+        if self.allow_encoding_decoding_same_bandset:
```
Hgherzog, line 1083: I think it is clearer to randomly select in both cases then to always take all of the other as we were doing before. If we want an always decode all option we should have that seperately or you just set min and max decoded bandsets to be very high.

olmoearth_pretrain/train/masking.py
Hgherzog: The 2 modality case I think fits into the multiple modality case so we can just delete it.

PR #413, opened 2025-10-25
Switch to FreezeUnfreeze for finetune

yawenzzzz:
1. Remove early stopping and previously lr scheduler, switch to FreezeUnfreeze and ReduceLROnPlateau. This setup is more similar to the finetuning recipe for our partner tasks, though training for less epochs, 50 instead of 100. 
2. Fix issue with AnySat, make sure only downsampling during train, not val/test.
3. Add patch size experiments for m_sa_crop_type and m_cashew_plant with OlmoEarth models.

> [!NOTE]
> Implements freeze–unfreeze finetuning with LR scheduling, threads is_train into embeddings/wrappers (fixing AnySat train-time subsampling), and trims/retunes finetune tasks and sweep configs.
> 
> - **Finetune**:
>   - Add freeze–unfreeze: freeze backbone for first 20% epochs, then unfreeze and drop LR 10×.
>   - Introduce `ReduceLROnPlateau`; remove per-step LR warmup/adjust and early stopping.
>   - Add `_set_backbone_trainable`; ensure eval uses `is_train=False`.
> - **Embeddings & Wrappers**:
>   - Thread `is_train` through `get_embeddings` and `EvalWrapper.__call__` (all subclasses).
>   - Replace `self.is_train` with `is_train` in `AnySatEvalWrapper` to control train-time pixel subsampling.
>   - Evaluator now passes `is_train` to `get_embeddings` and no longer sets it in wrapper ctor.
> - **Eval configs / sweep**:
>   - Streamline `FT_EVAL_TASKS` (comment out many; keep `m_sa_crop_type`, `m_cashew_plant` with batch/worker tweaks).
>   - Narrow LR sweep to `[1e-4, 5e-4, 1e-3]`; update run name suffixes to include `freezeunfreeze`; refresh example run strings.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3a18ece5de3fc28d1d252dcb7807c64fe8ac5647. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/olmoearth_pretrain/evals/models/anysat/anysat.py
+++ b/olmoearth_pretrain/evals/models/anysat/anysat.py
@@ -62,7 +62,7 @@ class AnySat(nn.Module):
     supports_multiple_modalities_at_once = True
     resolution: int = 10
 
-    def __init__(self, patch_size: int = 8) -> None:
+    def __init__(self, patch_size: int = 4) -> None:
         """AnySat wrapper."""
         super().__init__()
         self.model = torch.hub.load(
@@ -264,7 +264,7 @@ class AnySat(nn.Module):
 class AnySatConfig(Config):
     """olmo_core style config for AnySat."""
 
-    patch_size: int = 8
+    patch_size: int = 4
 
     def build(self) -> AnySat:
         """Build the Croma model."""
--- a/olmoearth_pretrain/evals/models/galileo/single_file_galileo.py
+++ b/olmoearth_pretrain/evals/models/galileo/single_file_galileo.py
@@ -1699,7 +1699,7 @@ class GalileoWrapper(nn.Module):
     def __init__(
         self,
         pretrained_path: UPath,
-        patch_size: int = 8,
+        patch_size: int = 4,
         month: int = 6,
         add_layernorm_on_exit: bool = True,
         use_pretrained_normalizer: bool = True,
@@ -1937,7 +1937,7 @@ class GalileoConfig(Config):
     """olmo_core style config for GalileoWrapper."""
 
     size: str = "base"
-    patch_size: int = 16
+    patch_size: int = 4
     month: int = 6
     add_layernorm_on_exit: bool = True
     use_pretrained_normalizer: bool = True


olmoearth_pretrain/train/callbacks/evaluator_callback.py
```diff
@@ -324,22 +358,51 @@ def _val_finetune(self) -> tuple[float, float]:
...
+            logger.info(
+                f"Best checkpoint already exists for {self.evaluation_name}, skipping finetuning"
+            )
+            return 0.0, 0.0
+        else:
+            logger.info(
+                f"Best checkpoint does not exist for {self.evaluation_name}, running finetuning"
+            )
+
+        try:
```
favyen2: I don't think you need the try/except here. Anyway it will raise exception and exit if run_finetune_eval encounters error. Can just do like this:
```python
val_result, test_result = run_finetune_eval(...)
logger.info(...)
model.load_state_dict(original_state)
if original_training_mode:
    model.train()
else:
    model.eval()
```

olmoearth_pretrain/train/callbacks/evaluator_callback.py
```diff
@@ -324,22 +358,51 @@ def _val_finetune(self) -> tuple[float, float]:
...
-            patch_size=self.patch_size,
-            pooling_type=self.pooling_type,
-            use_pooled_tokens=self.use_pooled_tokens,
-            train_loader=train_loader,
-            val_loader=val_loader,
-            test_loader=test_loader,
-        )
-        logger.info(
-            f"Downstream evaluator {self.evaluation_name} val score: {val_result}, test score: {test_result}"
+        # Check if best checkpoint already exists
```
Hgherzog: Suggestion: Move the checkpointing into a helper function

olmoearth_pretrain/train/callbacks/evaluator_callback.py
```diff
@@ -312,38 +341,64 @@ def _val_finetune(self) -> tuple[float, float]:
...
+                model.eval()
 
-        torch.cuda.empty_cache()
-        gc.collect()
-        return val_result, test_result
+            torch.cuda.empty_cache()
+            gc.collect()
+            return val_result, test_result
 
     def val(self) -> tuple[float, float]:
```
Hgherzog, line 405: Why do we need to update the train vs eval state again after validation?

olmoearth_pretrain/evals/eval_wrapper.py
```diff
@@ -50,7 +50,6 @@ def __init__(
         pooling_type: PoolingType,
         concat_features: bool = False,
         use_pooled_tokens: bool = False,
-        is_train: bool = True,
```
Hgherzog, line 53: Is this only used by anysat? Also, what was reason to move this out of the wrapper and into the method signature ?

olmoearth_pretrain/train/callbacks/evaluator_callback.py
yawenzzzz: Fixed!

olmoearth_pretrain/train/callbacks/evaluator_callback.py
```diff
@@ -312,38 +341,64 @@ def _val_finetune(self) -> tuple[float, float]:
...
+            self.evaluation_name,
+            f"lr{self.ft_lr}",
+            "best.ckpt",
         )
+        if os.path.exists(best_checkpoint_path):
+            logger.info("Best checkpoint already exists, skipping finetuning")
+            return 0.0, 0.0
+        else:
+            logger.info("Best checkpoint does not exist, running finetuning")
+            val_result, test_result = run_finetune_eval(
```
Hgherzog, line 379: When do we set model.eval() for running evaluations? I couldn't find it.

Hgherzog: Is it implicitly set in the saved checkpoint? I think it is better to explicitly set it or not before running evals.

olmoearth_pretrain/evals/eval_wrapper.py
yawenzzzz: yes, this is only used by Anysat, the reason of moving it into the method signature is to avoid reinitialize the same wrapper twice, AnySat needs to do label sampling only in train, but using original label in val/test. with this, we can just pass the `is_train` to get different behaviors of label sampling or not.

olmoearth_pretrain/train/callbacks/evaluator_callback.py
yawenzzzz: this is redundant, we already have `model.train()` and `module.eval()` in the `finetune.py`

yawenzzzz: i removed it.

olmoearth_pretrain/train/callbacks/evaluator_callback.py
yawenzzzz: this is redundant, we already have `model.train()` and `module.eval()` in the `finetune.py`

olmoearth_pretrain/train/callbacks/evaluator_callback.py
yawenzzzz: this is not checkpointing, but instead checking if best checkpoint exists or not, and the path will be passed to the `run_finetune_eval` function.  this is mainly used for resume, if best checkpoint exists, the task will be skipped.

Hgherzog: same nit applies, up to you

yawenzzzz: fixed.

PR #420, opened 2025-10-30
local vs gcs paths in upath produce totally different types

cmwilhelm:
> [!NOTE]
> Switch to UPath for resolving and opening model artifacts, and add tests validating GCS path handling.
> 
> - **Model loading**:
>   - `_resolve_artifact_path` now returns `UPath`; construct paths with `UPath` for both IDs and local/remote paths.
>   - `_load_model_from_config` and `_load_state_dict` accept `UPath` and use `path.open(...)`; `torch.load` reads via file handle.
> - **Tests**:
>   - Add GCS path test expecting `NotFound`/`FileNotFoundError`/`OSError` to verify remote path handling.
>   - Import `NotFound` from `google.api_core.exceptions`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b0e575cbf0b785b9ecbe9cfe2c1b51b43e1cf5b5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/olmoearth_pretrain/model_loader.py
+++ b/olmoearth_pretrain/model_loader.py
@@ -110,7 +110,8 @@ def _load_model_from_config(path: UPath) -> torch.nn.Module:
     return model_config.build()
 
 
-def _load_state_dict(path: PathLike) -> dict[str, torch.Tensor]:
+def _load_state_dict(path: UPath) -> dict[str, torch.Tensor]:
     """Load the model state dict from the specified path."""
-    state_dict = torch.load(path, map_location="cpu")
+    with path.open("rb") as f:
+        state_dict = torch.load(f, map_location="cpu")
     return state_dict


olmoearth_pretrain/model_loader.py
```diff
@@ -92,7 +92,7 @@ def load_model_from_path(
 
 def _resolve_artifact_path(
```
favyen2, line 93: Sounds like this function has caused a few bugs, the external API is separate for model ID vs checkpoint path, so I would just put this code into load_model_from_id and load_model_from_path, the latter would just be running `UPath(model_path) / filename` which is simple enough that a separate function adds unnecessary complexity.

PR #429, opened 2025-11-03
Cleanup eval datasets and models

gabrieltseng:
- [x] Remove CropHarvest
- [x] Remove sickle 
- [x] Remove CopernicusFM
- [x] Make Breizhcrops and optional dependency
    - Example run where breizhcrops is not installed (and skipped): https://beaker.allen.ai/orgs/ai2/workspaces/earth-systems/work/01K94QVCDETSXB0KEJMMD0W11W/logs?jobId=01K94QVCNEN0934QRZQ8MFZ4Z7
    - Example run where breizhcrops is installed and runs: https://beaker.allen.ai/orgs/ai2/workspaces/earth-systems/work/01K94RJ1K4HSYZ0TN3BF6Z5N7C?taskId=01K94RJ1KR2VV2KYWS4WP4C6DT&jobId=01K94RJ1Q8JC1AXT16NAQ591T9

---

> [!NOTE]
> Remove CropHarvest/SICKLE datasets and CopernicusFM model, make BreizhCrops optional, and purge/update configs, scripts, tests, docs, and deps accordingly.
> 
> - **Evals/Datasets**:
>   - Remove `cropharvest` and `sickle` datasets, configs, paths, stats, processors, and tests; strip all references across scripts and tasks.
>   - Make `breizhcrops` optional: add runtime import guard, internal mapping refactor; update docs and pipelines accordingly.
> - **Models**:
>   - Remove CopernicusFM model, wrapper, launch script, and model enum/dispatch logic.
> - **Config/Validation**:
>   - Tighten `input_modalities` support (drop `cropharvest`/`sickle`), remove env vars/paths, and delete related entries in `minmax_stats.json` and dataset configs.
> - **Scripts/CI/Tests**:
>   - Purge eval sweep/finetune scripts and unit tests referencing removed datasets/models; update expected tasks in tests.
> - **Docs/Dependencies**:
>   - Update Pretraining docs to drop removed dataset env vars; remove `breizhcrops` from eval extras; adjust lockfile and torch markers; add explicit optional install where needed.
> - **Misc**:
>   - Minor utility additions (visualize config in official scripts), and small evaluator message/logic cleanups.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1fcd73fb681f6ad2069ca16d5d468050c786ca32. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


olmoearth_pretrain/internal/all_evals.py
```diff
@@ -547,7 +464,7 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
     build_train_module_config = user_mod.build_train_module_config
     build_dataset_config = user_mod.build_dataset_config
     build_dataloader_config = user_mod.build_dataloader_config
-    build_visualize_config = user_mod.build_visualize_config
+    build_visualize_config = getattr(user_mod, "build_visualize_config", None)
```
gabrieltseng: Without this change, I was getting the following error:
```
Traceback (most recent call last):
  File "/Users/gabrieltseng/Documents/code/helios/olmoearth_pretrain/internal/all_evals.py", line 467, in <module>
    build_visualize_config = user_mod.build_visualize_config
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'user_module' has no attribute 'build_visualize_config'. Did you mean: 'build_trainer_config'?
```

olmoearth_pretrain/train/callbacks/evaluator_callback.py
```diff
@@ -118,6 +118,10 @@ def __init__(
         self.device = device
         # Add all task attributes to self
         self.dataset = task.dataset
+        # datasets can have an "is_active" attribute. if its false, we
+        # should skip it. We assume its true by default. Currently (20251103)
+        # this only applies to Breizhcrops
+        self.is_dataset_active = getattr(self.dataset, "is_active", True)
```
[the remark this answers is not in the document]
gabrieltseng: fixed

olmoearth_pretrain/evals/datasets/breizhcrops.py
```diff
@@ -83,6 +84,29 @@ def __init__(
...
+                logger.info(
+                    "breizhcrops package not available. The Breizhcrops eval will be skipped."
+                )
+                return None
+
+        self.bc_selected_bands = SELECTED_BANDS
+
+        self.input_to_output_band_mapping = [
+            SELECTED_BANDS[LEVEL].index(_helios2bc_name(b)) for b in EVAL_S2_BAND_NAMES
+        ]
```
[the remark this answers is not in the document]
gabrieltseng: fixed

olmoearth_pretrain/internal/all_evals.py
favyen2: I think `scripts/official/script.py` is missing `build_visualize_config`, I get the same error when running `python scripts/official/large.py` for example. I don't think it should be changed here.

olmoearth_pretrain/evals/datasets/breizhcrops.py
```diff
@@ -83,6 +80,19 @@ def __init__(
             norm_method: Normalization method to use, only when norm_stats_from_pretrained is False
             monthly_average: Whether to compute a monthly average of the timesteps
         """
+        if not self.is_active():
```
favyen2: Is there a way to document this in the evaluation documentation, and have the default be to exclude breizhcrops, instead of having this automatic disabling of the eval?

gabrieltseng: I can definitely add to the documentation.

Why would you prefer to exclude Breizhcrops ? (although this is sort of what happens currently by default)?

Currently, a user only has to pip install breizhcrops for it to be included. If we remove breizhcrops from `ALL_EVAL_TASKS`, a user has to install breizhcrops *and* update the launch scripts.

olmoearth_pretrain/internal/all_evals.py
gabrieltseng: added

olmoearth_pretrain/evals/datasets/breizhcrops.py
favyen2: Is there no way to pass a subset of tasks for evaluation? For training, as far as I can tell breizhcrops is already disabled, and it takes a list of tasks.

favyen2: It just feels like it would be a lot less messy if the user could just pass a list of tasks to enable it when they want to evaluate on breizhcrops. Since otherwise breizhcrops is a special case.

gabrieltseng: > Is there no way to pass a subset of tasks for evaluation? 

I don't think there is a way to easily do this (cc @Hgherzog )? but I think its okay if we just remove Breizhcrops from `ALL_EVAL_TASKS` for now.

So I will:
1. Remove Breizhcrops from `ALL_EVAL_TASKS` so it doesn't get run
2. Remove the `is_active` check
3. raise a `RunTimeError` if Breizhcrops gets run without the import

gabrieltseng: Done in https://github.com/allenai/olmoearth_pretrain/pull/429/commits/c2430a53cbf7e6688b97b87db914bf48006a5247

Hgherzog: You can always use the `--trainer.callbacks.downstream_evaluator.tasks_to_run=/[<task1>,<task2>` override. We could also easily add a tasks to skip override as well.

olmoearth_pretrain/evals/datasets/breizhcrops.py
```diff
@@ -83,6 +80,25 @@ def __init__(
...
```
favyen2: I was imagining you wouldn't need this is_active code if breizhcrops wasn't enabled by default.  (Main reason I suggested having doc show to disable it by default was to avoid the extra code here.) But maybe that was not correct. Anyway I think it is fine, was just hoping to reduce the clutter a bit .

gabrieltseng: i still wanted to keep it since the breizhcrops is guaranteed to be installed (throws a more informative error message + specific version to install), but I can remove it (or rename it to something better like `is_breizhcrops_package_installed`

gabrieltseng: oh actually you're right I can just remove it

PR #454, opened 2025-12-24
Henryh/olmo core free inference

Hgherzog:
### Summary

This PR decouples model inference from the `olmo-core` training library, allowing users to load and run OlmoEarth models from Hugging Face without installing `ai2-olmo-core`. This significantly reduces the dependency footprint for inference-only use cases.

### Motivation

Previously, loading a model required `olmo-core` because the `Config` base class and config deserialization logic came from that library. This was unnecessary for inference—users only need to deserialize a `config.json` and build the model weights. By making `olmo-core` optional, we:

- Reduce install size and dependency conflicts for downstream users
- Enable easier integration into production inference pipelines
- Maintain full training functionality when `olmo-core` is installed

#### Core Infrastructure

**`olmoearth_pretrain/config.py`** (NEW)
- Centralized conditional import logic for `Config`, `StrEnum`, and `DType`
- Implements `_StandaloneConfig` that mimics `olmo-core`'s `Config` for inference-only mode
- Exports `OLMO_CORE_AVAILABLE` flag and `require_olmo_core()` guard function
- Issues a single warning on first import if `olmo-core` is unavailable

**`olmoearth_pretrain/datatypes.py`** (NEW)
- Moved `MaskValue` and `MaskedOlmoEarthSample` from `train/masking.py`
- These are used by model forward passes and shouldn't live in training code

**`olmoearth_pretrain/types.py`**
- Now only contains `ArrayTensor` type alias

#### Import Updates

Updated all files to import from the new centralized locations:
- `from olmo_core.config import Config` → `from olmoearth_pretrain.config import Config`
- `from olmoearth_pretrain.train.masking import MaskedOlmoEarthSample` → `from olmoearth_pretrain.datatypes import ...`

Files updated:
- `nn/flexi_vit.py`, `nn/latent_mim.py`, `nn/mae.py`, `nn/galileo.py`, `nn/st_model.py`, `nn/pooled_modality_predictor.py`
- `model_loader.py`
- `data/dataloader.py`, `data/dataset.py`, `data/transform.py`, `data/concat.py`
- `train/loss.py`, `train/masking.py`, `train/train_module/train_module.py`
- `internal/experiment.py`
- `evals/datasets/__init__.py`
- Various eval model configs
- `dataset/convert_to_h5py.py`
- `inference_benchmarking/*.py`

#### Training Guards

**`olmoearth_pretrain/train/__init__.py`** and **`olmoearth_pretrain/internal/__init__.py`**
- Added explicit `require_olmo_core()` guards that raise `ImportError` with clear instructions if training code is accessed without `olmo-core`

#### Dependency Management

**`pyproject.toml`**
- Moved `ai2-olmo-core` from base dependencies to `[dependency-groups] training`
- Added `inference_deps_only` pytest marker
- Added `tests_inference/` to test paths

#### Testing

**`tests_inference/test_model_loading.py`** (NEW)
- Integration tests that verify model loading works without `olmo-core`
- Tests `EncoderConfig` deserialization and building
- Runs in a separate test root to ensure clean dependency isolation

#### CI

**`.github/workflows/tests.yaml`**
- Added `test-inference-only` job that runs with only base + dev dependencies (no `olmo-core`)
- Main `test` job now uses `uv sync --locked --all-groups`

#### Documentation

**`README.md`**
- Added "Inference-Only Installation" section

### Backward Compatibility

- `train/masking.py` re-exports `MaskValue` and `MaskedOlmoEarthSample` for backward compatibility
- When `olmo-core` is installed, behavior is identical to before
- All existing tests pass

### Testing
Split tests into main tests and the inference_only tests
`uv run --only-group dev pytest -v tests_inference/`

### DIFF
--- a/tests_minimal_deps/test_model_loader.py
+++ b/tests_minimal_deps/test_model_loader.py
@@ -135,17 +135,6 @@ class TestConfigExport:
         """Test that the exported Config has from_dict for model loading."""
         assert hasattr(Config, "from_dict")
 
-    def test_olmo_core_flag_reflects_installation(self) -> None:
-        """Test that OLMO_CORE_AVAILABLE correctly reflects olmo-core installation."""
-        try:
-            import olmo_core  # noqa: F401
-
-            expected = True
-        except ImportError:
-            expected = False
-
-        assert OLMO_CORE_AVAILABLE == expected
-
 
 # =============================================================================
 # EncoderConfig Loading Tests


pyproject.toml
Hgherzog, line 1: We could also have an inference minimal dependencies be a separate group that someone could install with `--only-group` flag using uv and keep the minimal group as something where you can run tests and other code. (though we were already using additional groups in the test.)

yawenzzzz (approved):
LGTM!

PR #464, opened 2026-01-16
Feature/configurable tokenization

Hgherzog:


olmoearth_pretrain/data/constants.py
```diff
@@ -195,6 +195,142 @@ def is_static_in_space_and_time(self) -> bool:
...
+
+@dataclass
+class ModalityTokenization:
+    """Custom tokenization configuration for a single modality.
+
+    Specifies how bands should be grouped into tokens. Each band_group
+    becomes a separate token.
+    """
+
+    band_groups: list[TokenizationBandSet]
```
gabrieltseng, line 226: Do we need a TokenizationBandSet class here? Since it will always be called within a ModalityTokenization can the band_groups just be a list of tuples / list of lists?

tests/unit/test_tokenization_config.py
```diff
@@ -0,0 +1,195 @@
...
+                        # SWIR group
+                        TokenizationBandSet(bands=["B11", "B12"]),
+                    ]
+                )
+            }
+        )
+
+        indices = config.get_bandset_indices("sentinel2_l2a")
+        # B02=0, B03=1, B04=2, B08=3, B8A=7, B11=8, B12=9
+        assert indices == [[0, 1, 2], [3, 7], [8, 9]]
```
gabrieltseng, line 65: Can this test call `Modality.SENTINEL2_L2A.band_order` instead of testing against hard coded indices?

gabrieltseng: (same for all tests below)

PR #465, opened 2026-01-16
Add (cursor's) script to copy wandb runs

gabrieltseng:


scripts/copy_wandb_run.py
Hgherzog, line 1: I would remove the date from this name

scripts/copy_wandb_run.py
```diff
@@ -0,0 +1,157 @@
+#!/usr/bin/env python3
+"""Copy specific metrics from one wandb run to another project, preserving step values."""
```
Hgherzog, line 2: I would add an example command of how to copy the base official run

Hgherzog: `python scripts/tools/20260116_copy_wandb_run.py \
  --src-entity eai-ai2 \
  --src-project 2025_10_02_phase2 \
  --run-id v9a79d89 \
  --dest-entity eai-ai2 \
  --dest-project <your project> \
  --run-name "phase2.0_base_lr0.0001_wd0.02"`

scripts/copy_wandb_run.py
gabrieltseng: updated

scripts/copy_wandb_run.py
gabrieltseng: updated

PR #467, opened 2026-01-22
New masking, and move the masking into the dataloader

gabrieltseng:
### TL;DR

This PR (optionally) moves the masking functionality into the dataloader, and out of the `train_batch` functions. This is nice because the masking code can be done in the dataloader workers, allowing the GPU to just be used for the model forward / backward passes.

This results in a 1.5 - 2x speedup during pre-training. It also means runs are much less sensitive to the speed of the masking code, so we have a bit more liberty to experiment there. For example, I had one masking experiment which nearly doubled training time; it now is the same as the original masking.

There are also a few other changes (I've noted them in the github comments below) which result in meaningful speed ups:

1. Vectorize the missing values computation in the dataloader, and avoid creating an `OlmoEarthSample` just to know the expected mask shape.
2. `non_blocking=True` when moving the tensors from CPU to GPU (as best as I can tell, a free speedup since we pin memory)
3. Small improvements to speed `unmask` up
4. Avoid CPU syncs in the model

(2) and (4) do a lot of the speedups, so if we want 70% of the gain with 1% of the code change that's the way to go. But in my opinion this is not that complicated and allows us to better take advantage of our CPUs.

### LR (long read)

I have mostly been using the `throughput/device/model duration (s)` metric. This measures the total time required for a `train_batch` function. This provides some signal, but is not ideal; different GPUs can be slower or faster (I assume because of other things happening on the node but I am not sure). Below, you can see different restart events significantly affect the model duration.

Screenshot 2026-01-27 at 9 50 26 AM

I have been running things on single GPUs (easier to get scheduled) so all the numbers are x8 compared to what it should actually look like. 

**Overall**: the changes in this PR (blue) are faster than main (yellow). 

Screenshot 2026-01-27 at 9 53 11 AM

Just removing the CPU sync in the model (the boolean evaluation in the linear projection) helps a lot. I am open to only having this change (which is in the much smaller branch `gabi/just-syncs`)

Screenshot 2026-01-27 at 9 57 13 AM

However, I do want to emphasize that moving the dataloading off the GPU really can speed things up, especially if (i) masking is a bit slower, and (ii) bs64 fits ("old school" == existing implementation, with all the small speed ups like non_blocking and avoiding cpu syncs):

Screenshot 2026-01-27 at 5 09 12 PM

### DIFF
--- a/scripts/official/ablations/base_mae.py
+++ b/scripts/official/ablations/base_mae.py
@@ -58,6 +58,26 @@ MAX_PATCH_SIZE = 8
 MIN_PATCH_SIZE = 1
 
 
+def get_masking_config() -> MaskingConfig:
+    """Get the masking configuration for the experiment."""
+    return MaskingConfig(
+        strategy_config={
+            "type": "modality_cross_random",
+            "encode_ratio": 0.5,
+            "decode_ratio": 0.5,
+            "allow_encoding_decoding_same_bandset": True,
+            "only_decode_modalities": [
+                "worldcover",
+                "srtm",
+                "openstreetmap_raster",
+                "wri_canopy_height_map",
+                "cdl",
+                "worldcereal",
+            ],
+        }
+    )
+
+
 def my_build_common_components(
     script: str,
     cmd: SubCmd,
@@ -123,22 +143,7 @@ def build_train_module_config(
     return MAETrainModuleConfig(
         optim_config=AdamWConfig(lr=0.0001, weight_decay=0.02, fused=False),
         rank_microbatch_size=32,
-        masking_config=MaskingConfig(
-            strategy_config={
-                "type": "modality_cross_random",
-                "encode_ratio": 0.5,
-                "decode_ratio": 0.5,
-                "allow_encoding_decoding_same_bandset": True,
-                "only_decode_modalities": [
-                    "worldcover",
-                    "srtm",
-                    "openstreetmap_raster",
-                    "wri_canopy_height_map",
-                    "cdl",
-                    "worldcereal",
-                ],
-            }
-        ),
+        masking_config=get_masking_config(),
         mae_loss_config=LossConfig(
             loss_config={"type": "mae", "loss_function": "SmoothL1Loss", "beta": 0.1}
         ),
@@ -157,8 +162,6 @@ def build_train_module_config(
 
 def build_dataloader_config(common: CommonComponents) -> OlmoEarthDataLoaderConfig:
     """Build the dataloader config for an experiment."""
-    # things should be set during building
-
     return OlmoEarthDataLoaderConfig(
         num_workers=16,
         global_batch_size=512,
@@ -169,6 +172,8 @@ def build_dataloader_config(common: CommonComponents) -> OlmoEarthDataLoaderConf
         max_patch_size=MAX_PATCH_SIZE,
         work_dir=common.save_folder,
         seed=3622,
+        num_masked_views=1,  # MAE only needs 1 view (no contrastive)
+        masking_config=get_masking_config(),
     )
 
 


olmoearth_pretrain/train/masking.py
```diff
@@ -1599,6 +1599,159 @@ def __init__(
         )
 
 
+@MASKING_STRATEGY_REGISTRY.register("random_with_decode")
+class RandomWithDecodeMaskingStrategy(MaskingStrategy):
```
Hgherzog, line 1615: I think this already exists in the `class RandomFixedModalityMaskingStrategy(FixedModalityMaskingStrategy):`  modality class

olmoearth_pretrain/datatypes.py
```diff
@@ -2,15 +2,570 @@
...
+    @property
+    def modalities(self) -> list[str]:
+        """Get the modalities present in the sample.
+
+        Includes timestamps and latlon
+        """
+        return [modality for modality in self.as_dict(ignore_nones=True).keys()]
+
+    def to_device(
+        self, device: torch.device, non_blocking: bool = True
```
gabrieltseng, line 109: We have pin_memory set to True. This article suggests a good next step is to use non_blocking when moving the tensors from CPU to GPU. 

https://docs.pytorch.org/tutorials/intermediate/pinmem_nonblock.html

Our GPU time spent accessing memory seems high, so maybe this will help? 

Image

olmoearth_pretrain/datatypes.py
```diff
@@ -2,15 +2,570 @@
...
+from torch.distributed import DeviceMesh
+from torch.distributed.tensor import distribute_tensor
 
+from olmoearth_pretrain.data.constants import MISSING_VALUE, TIMESTAMPS, Modality
 from olmoearth_pretrain.types import ArrayTensor
 
-if TYPE_CHECKING:
-    from olmoearth_pretrain.data.dataset import OlmoEarthSample
+
+class OlmoEarthSample(NamedTuple):
```
gabrieltseng, line 19: Moved to avoid circular imports

olmoearth_pretrain/datatypes.py
gabrieltseng: Seem pytorch devs expect it to have no negative side effects: https://discuss.pytorch.org/t/should-we-set-non-blocking-to-true/38234/8

olmoearth_pretrain/nn/flexi_vit.py
```diff
@@ -482,23 +482,17 @@ def apply_embedding_to_modality(
...
-                patchified_data = embedding_module(
-                    patchified_data, **modality_specific_kwargs
-                )
-            else:
-                mask_shape = token_mask.shape + (self.embedding_size,)
-                patchified_data = torch.zeros(
-                    mask_shape, dtype=modality_data.dtype, device=token_mask.device
-                )
+
+            buffer_name = self._get_buffer_name(modality, idx)
```
gabrieltseng, line 489: The CPU sync caused by the boolean evaluation is slower than just running the convolutional layers.

Image

olmoearth_pretrain/train/masking.py
gabrieltseng: They are not exactly the same - RandomWithDecodeMaskingStrategy applies the random mask per instance & modality, and splits band sets.

scripts/official/script.py
gabrieltseng: It's possible these parameters (num_workers, prefetch_factor) are sub optimal when we mask on the dataloader, since in this situation the CPUs are doing far more work. 

Run to see if this is the case:
https://beaker.org/orgs/ai2/workspaces/earth-systems/work/01KG2RFXJQ7G2R7WQGTEBP6MQW?taskId=01KG2RFXTH291TNCSN57YKN66N&jobId=01KG2RFXY6TQNN967RDA5CEF8S

olmoearth_pretrain/data/dataset.py
```diff
@@ -32,529 +29,90 @@
...
 
 logger = logging.getLogger(__name__)
 
 
-class OlmoEarthSample(NamedTuple):
-    """A sample of the data from the OlmoEarth Pretrain dataset.
-
-    This is a namedtuple that contains the data of a single sample or a batch of samples from the OlmoEarth Pretrain dataset.
-    For each modality, we have an ArrayTensor named by the modality, along with the latlon and timestamps.
+def collate_single_masked_batched(
```
Hgherzog: Suggestion: I think having a separate collate.py module would be sensible at this point rather than having one in datatypes and 2 here

olmoearth_pretrain/data/dataset.py
```diff
@@ -835,19 +393,54 @@ def _fill_missing_timesteps(
             dtype=self.dtype,
         )
 
-        # Copy the existing data to the appropriate timestep positions
+        # Copy the existing data to the appropriate timestep positions using vectorized indexing
```
Hgherzog:
```suggestion
        # Copy the existing data to the appropriate timestep positions
```

olmoearth_pretrain/data/dataset.py
```diff
@@ -835,19 +393,54 @@ def _fill_missing_timesteps(
...
-                full_timesteps_data[..., idx, :] = modality_data[..., i, :]
+        num_to_copy = min(len(present_indices), t)
+        if num_to_copy > 0:
+            full_timesteps_data[:, :, present_indices[:num_to_copy], :] = modality_data[
+                :, :, :num_to_copy, :
+            ]
 
         return full_timesteps_data
 
+    @staticmethod
```
Hgherzog: In the `OlmoEarthSample` class we have a very similar method. Can we condense to share that code?

olmoearth_pretrain/data/dataset.py
```diff
@@ -863,12 +456,27 @@ def fill_sample_with_missing_values(
             f"Timestamps shape {sample_dict['timestamps'].shape[0]} does not match max_sequence_length {self.max_sequence_length}"
         )
         missing_modalities = []
-        sample = OlmoEarthSample(**sample_dict)
+
```
Hgherzog: Why are we avoiding creating the OlmoEarth sample early?

olmoearth_pretrain/data/dataset.py
Hgherzog: It may also be reasonable to have a function that takes in the olmoearth sample or a similar data container that does this so it can be used across the codebase.

olmoearth_pretrain/nn/flexi_vit.py
```diff
@@ -482,23 +482,17 @@ def apply_embedding_to_modality(
...
-                ]
-                patchified_data = embedding_module(
-                    patchified_data, **modality_specific_kwargs
-                )
-            else:
-                mask_shape = token_mask.shape + (self.embedding_size,)
-                patchified_data = torch.zeros(
-                    mask_shape, dtype=modality_data.dtype, device=token_mask.device
-                )
+
```
Hgherzog, line 488: We should be able to remove fast pass from the signature in this case

tests/unit/data/test_dataloader_masking.py
```diff
@@ -0,0 +1,396 @@
...
+    OlmoEarthSample,
+    collate_double_masked_batched,
+    collate_olmoearth_pretrain,
+    collate_single_masked_batched,
+)
+from olmoearth_pretrain.datatypes import MaskedOlmoEarthSample
+from olmoearth_pretrain.train.masking import MaskingConfig
+
+
+class TestMaskedOlmoEarthSampleToDevice:
```
Hgherzog: This test is unneeded as the data will start on CPU

olmoearth_pretrain/train/masking.py
Hgherzog: Can you add to the docstring on it to make the distinction clear

Hgherzog (commented):
I am excited about the direction of the changes and I think we should commit to them more fully. 
- I think we should at least default to using the new set-up 
- I am honestly pro just ensuring that all the official scripts still work with the new dataloading setup and saying goodbye to the old objectively worse imo set-up. That way we don't have all the complicated if then handling of it and everyone always uses faster training. If there are more serious cons of not switching over, it would be great if you could highlight them for me.

olmoearth_pretrain/data/dataset.py
gabrieltseng: updated

olmoearth_pretrain/data/dataset.py
gabrieltseng: this means we don't have to create the OlmoEarthSample twice (we recreate it in L400).

olmoearth_pretrain/train/masking.py
```diff
@@ -1556,7 +1552,11 @@ def apply_mask(
             )
             if mask is None:
                 continue
+            instance = getattr(masked_sample, modality)
             mask[:] = MaskValue.DECODER.value
+            mask = self.fill_mask_with_missing_values(
+                instance, mask, Modality.get(modality)
+            )
```
gabrieltseng, line 1551: @Hgherzog When I was comparing RandomWithDecodeMaskingStrategy to FixedModalityMaskingStrategy I noticed this small fix.

olmoearth_pretrain/train/masking.py
gabrieltseng: Added

scripts/estimate_token_ratios.py
```diff
@@ -0,0 +1,648 @@
+"""Estimate the percentage of tokens encoded and decoded for a given masking strategy.
+
+This script samples from all possible input sizes (patch_size, sampled_hw_p) and
+passes the data through a masking strategy to compute what ratio of tokens are
+encoded vs decoded.
+
+Example usage:
+    python scripts/estimate_token_ratios.py --num_samples 1000
+"""
```
gabrieltseng, line 9: Here are some results from this script. Existing masking strategy:

```bash
$ python scripts/estimate_token_ratios.py --masking_type modality_cross_random
Running with masking config: {'type': 'modality_cross_random', 'encode_ratio': 0.5, 'decode_ratio': 0.5, 'allow_encoding_decoding_same_bandset': True, 'only_decode_modalities': ['worldcover', 'srtm', 'openstreetmap_raster', 'wri_canopy_height_map', 'cdl', 'worldcereal']}
Training modalities: ['sentinel2_l2a', 'sentinel1', 'landsat', 'worldcover', 'srtm', 'openstreetmap_raster', 'wri_canopy_height_map', 'cdl', 'worldcereal']
Sampling 1000 configurations...
Sampling: 100%|██████████████████████████████████████████| 1000/1000 [00:02<00:00, 461.52it/s]

============================================================
TOKEN RATIO STATISTICS
============================================================

Number of samples: 1000

--- Total Token Counts ---
  Mean total tokens:       1517.7
  Std total tokens:        702.7
  Min total tokens:        78
  Max total tokens:        2178

--- Encoded Ratio (including missing as denominator) ---
  Mean:   0.2487 (24.87%)
  Std:    0.0963
  Min:    0.0417 (4.17%)
  Max:    0.4615 (46.15%)
  Median: 0.2436 (24.36%)

--- Decoded Ratio (including missing as denominator) ---
  Mean:   0.3316 (33.16%)
  Std:    0.1002
  Min:    0.0285 (2.85%)
  Max:    0.5000 (50.00%)
  Median: 0.3386 (33.86%)

--- Encoded Ratio (excluding missing tokens) ---
  Mean:   0.2762 (27.62%)
  Std:    0.1047
  Min:    0.0645 (6.45%)
  Max:    0.4930 (49.30%)

--- Decoded Ratio (excluding missing tokens) ---
  Mean:   0.3686 (36.86%)
  Std:    0.1067
  Min:    0.0313 (3.13%)
  Max:    0.5202 (52.02%)

--- Percentiles (Encoded, excluding missing) ---
  5th percentile: 0.1238 (12.38%)
  25th percentile: 0.1875 (18.75%)
  50th percentile: 0.2718 (27.18%)
  75th percentile: 0.3621 (36.21%)
  95th percentile: 0.4603 (46.03%)

--- Percentiles (Decoded, excluding missing) ---
  5th percentile: 0.1754 (17.54%)
  25th percentile: 0.3043 (30.43%)
  50th percentile: 0.3799 (37.99%)
  75th percentile: 0.4431 (44.31%)
  95th percentile: 0.5005 (50.05%)
============================================================
```
and with the `random_with_decode` masking strategy:
```bash
$ python scripts/estimate_token_ratios.py --masking_type random_with_decode
Running with masking config: {'type': 'random_with_decode', 'encode_ratio': 0.5, 'decode_ratio': 0.5, 'only_decode_modalities': ['worldcover', 'srtm', 'openstreetmap_raster', 'wri_canopy_height_map', 'cdl', 'worldcereal']}
Training modalities: ['sentinel2_l2a', 'sentinel1', 'landsat', 'worldcover', 'srtm', 'openstreetmap_raster', 'wri_canopy_height_map', 'cdl', 'worldcereal']
Sampling 1000 configurations...
Sampling: 100%|██████████████████████████████████████████| 1000/1000 [00:02<00:00, 347.52it/s]

============================================================
TOKEN RATIO STATISTICS
============================================================

Number of samples: 1000

--- Total Token Counts ---
  Mean total tokens:       1517.7
  Std total tokens:        702.7
  Min total tokens:        78
  Max total tokens:        2178

--- Encoded Ratio (including missing as denominator) ---
  Mean:   0.1821 (18.21%)
  Std:    0.0347
  Min:    0.0417 (4.17%)
  Max:    0.2308 (23.08%)
  Median: 0.1917 (19.17%)

--- Decoded Ratio (including missing as denominator) ---
  Mean:   0.3501 (35.01%)
  Std:    0.0957
  Min:    0.2179 (21.79%)
  Max:    0.6250 (62.50%)
  Median: 0.3083 (30.83%)

--- Encoded Ratio (excluding missing tokens) ---
  Mean:   0.2020 (20.20%)
  Std:    0.0347
  Min:    0.0714 (7.14%)
  Max:    0.2687 (26.87%)

--- Decoded Ratio (excluding missing tokens) ---
  Mean:   0.3900 (39.00%)
  Std:    0.1058
  Min:    0.2609 (26.09%)
  Max:    0.9286 (92.86%)

--- Percentiles (Encoded, excluding missing) ---
  5th percentile: 0.1250 (12.50%)
  25th percentile: 0.1765 (17.65%)
  50th percentile: 0.2143 (21.43%)
  75th percentile: 0.2289 (22.89%)
  95th percentile: 0.2388 (23.88%)

--- Percentiles (Decoded, excluding missing) ---
  5th percentile: 0.2911 (29.11%)
  25th percentile: 0.3088 (30.88%)
  50th percentile: 0.3365 (33.65%)
  75th percentile: 0.4649 (46.49%)
  95th percentile: 0.6148 (61.48%)
============================================================
```

gabrieltseng:
> If we go forward with switching over. It would be great to repro first 20k steps or so of some of the official models as additional confirmation

The ablations running in https://github.com/allenai/olmoearth_pretrain/pull/474 seem to suggest things are behaving fine.

olmoearth_pretrain/internal/experiment.py
```diff
@@ -80,6 +80,8 @@ class CommonComponents(Config):
     training_modalities: list[str]
     launch: OlmoEarthBeakerLaunchConfig | None = None
     nccl_debug: bool = False
+    # Enable dataloader-side masking to offload masking from GPU to dataloader workers
+    dataloader_side_masking: bool = False
```
Hgherzog:
```suggestion
    dataloader_side_masking: bool = True
```

gabrieltseng: removed legacy support entirely as discussed on slack, so this argument is has been removed

olmoearth_pretrain/train/masking.py
```diff
@@ -1556,7 +1552,11 @@ def apply_mask(
             )
             if mask is None:
                 continue
+            instance = getattr(masked_sample, modality)
             mask[:] = MaskValue.DECODER.value
+            mask = self.fill_mask_with_missing_values(
```
Hgherzog: I don't think this mask is actually used later in the function, and the fill_mask_with_missing_values returns a cloned tensor rather than manipulating the underlying data.

olmoearth_pretrain/train/masking.py
```diff
@@ -1619,6 +1619,173 @@ def __init__(
...
+        self,
+        encode_ratio: float = 0.5,
+        decode_ratio: float = 0.5,
+        only_decode_modalities: list[str] = [],
+    ):
+        """Random masking strategy except for decode modalities, which only get decoded."""
+        self._encode_ratio = encode_ratio
+        self._decode_ratio = decode_ratio
+        self.only_decode_modalities = only_decode_modalities
+        self.generator = np.random.default_rng(0)
```
Hgherzog: I think we want to just use the global state for this as this will restart to the same order when we checkpoint.

olmoearth_pretrain/train/masking.py
gabrieltseng: I think the issue is that for the `decoded_modalities` this is the mask which is used. This means the masked sample will have an incorrect mask for these decoded modalities, since it won't respect the missing values.

olmoearth_pretrain/train/masking.py
gabrieltseng: updated this for all masking classes

scripts/official/script.py
```diff
@@ -118,20 +126,28 @@ def build_train_module_config(
...
+    (GPU). This improves throughput by offloading CPU-bound masking operations to
+    dataloader workers.
+    """
     return OlmoEarthDataLoaderConfig(
-        num_workers=16,
+        num_workers=12,
         global_batch_size=512,
         token_budget=2250,
-        prefetch_factor=4,
+        prefetch_factor=2,
```
gabrieltseng, line 153: This change seems to be yielding more memory errors:

```
[2;36m[2026-02-01 18:24:55][0m [1;31mERROR   [0m [2m[olmo_core.train.trainer:702, rank=2][0m Training failed due to:
RuntimeError: DataLoader worker (pid 261746) is killed by signal: Bus error. It is possible that dataloader's workers are out of shared memory. Please try to raise your shared memory limit.
[2;36m[2026-02-01 18:24:55][0m [1;31mERROR   [0m [2m[olmo_core.train.trainer:702, rank=0][0m Training failed due to:
```

I think this can be handled by tweaking these values (16 workers seems high?) but i want to highlight this in any case

olmoearth_pretrain/data/dataset.py
```diff
@@ -29,534 +26,26 @@
     deprecated_function_alias as _deprecated_function_alias,
 )
 from olmoearth_pretrain.config import Config
+from olmoearth_pretrain.data.collate import (  # noqa: F401
```
Hgherzog: The 2 new collate functions definitely don't need to be imported and I think we should be good to drop the collate alias for helios for the old ones as we don't have that path saved in a config.

olmoearth_pretrain/data/dataset.py
```diff
@@ -863,12 +356,27 @@ def fill_sample_with_missing_values(
             f"Timestamps shape {sample_dict['timestamps'].shape[0]} does not match max_sequence_length {self.max_sequence_length}"
         )
         missing_modalities = []
-        sample = OlmoEarthSample(**sample_dict)
+
+        # Extract h, w, t from sample_dict to avoid creating OlmoEarthSample early
+        time = sample_dict["timestamps"].shape[0]
```
Hgherzog, line 348: Nit: Move this extraction into a method

scripts/official/script.py
Hgherzog: That makes sense as now tensors for masks and inputs must be stored in shared memory so  reducing makes sense. Do we still saturate dataloaders with this new setup?

scripts/official/script.py
```diff
@@ -71,29 +71,37 @@ def build_common_components(
     return config
 
 
+def get_masking_config() -> MaskingConfig:
+    """Get the masking configuration for the experiment."""
+    return MaskingConfig(
+        strategy_config={
+            # "type": "modality_cross_random",
```
Hgherzog: The official script should keep the same masking strategy we released with.

olmoearth_pretrain/train/masking.py
Hgherzog: I agree that your change is needed otherwise the missing values wont be respected. However, the mask that you output is not actually used anywhere else because the function returns cloned data. So, the result is the missing mask is still never applied. In practice, the decoded modalities never have missing values during training because they are single timestamp so they are either None or there. 

Here is a quick script that shows the error:
```"""Demo script showing the mask mutation bug in FixedModalityMaskingStrategy."""

from collections import namedtuple

import torch

from olmoearth_pretrain.data.constants import MISSING_VALUE, Modality
from olmoearth_pretrain.datatypes import MaskValue
from olmoearth_pretrain.train.masking import MaskingStrategy

MaskedSample = namedtuple("MaskedSample", ["sentinel2_l2a", "sentinel2_l2a_mask"])


def buggy_approach():
    """This is what the current code does - the fix is discarded."""
    # Create sample with some missing data (first spatial position is missing)
    b, h, w, t, c = 1, 4, 4, 2, 12  # S2 has 12 bands
    data = torch.ones(b, h, w, t, c)
    data[:, 0, 0, :, :] = MISSING_VALUE  # First spatial position is missing

    mask = torch.zeros(b, h, w, t, 3)  # 3 bandsets for S2
    sample = MaskedSample(sentinel2_l2a=data, sentinel2_l2a_mask=mask)

    # Get mask from the namedtuple
    mask = sample.sentinel2_l2a_mask
    instance = sample.sentinel2_l2a
    modality = Modality.get("sentinel2_l2a")

    # In-place modification - this WORKS
    mask[:] = MaskValue.DECODER.value
    print(f"After mask[:] = DECODER:  sample mask unique values = {sample.sentinel2_l2a_mask.unique()}")

    # Use the ACTUAL fill_mask_with_missing_values function
    strategy = MaskingStrategy()
    mask = strategy.fill_mask_with_missing_values(instance, mask, modality)

    print(f"After fill_mask_with_missing_values:")
    print(f"  Local mask unique values:  {mask.unique()}")
    print(f"  sample.mask unique values: {sample.sentinel2_l2a_mask.unique()}")
    print()
    print(f"  Local mask has MISSING ({MaskValue.MISSING.value}): {(mask == MaskValue.MISSING.value).any()}")
    print(f"  sample.mask has MISSING ({MaskValue.MISSING.value}): {(sample.sentinel2_l2a_mask == MaskValue.MISSING.value).any()}")
    print()
    print("^ BUG: MISSING value was NOT applied to sample.sentinel2_l2a_mask!")

if __name__ == "__main__":
    print("=== BUGGY APPROACH (current code) ===")
    buggy_approach()
    print()
```

olmoearth_pretrain/train/utils.py
```diff
@@ -5,51 +5,49 @@
...
-
-        # Create a new OlmoEarthSample from the sliced fields
-        microbatches.append(OlmoEarthSample(**microbatch_dict))
-
-    return microbatches
+    # Compute split sizes (last chunk may be smaller)
+    split_sizes = [microbatch_size] * (num_microbatches - 1)
+    split_sizes.append(batch_size - microbatch_size * (num_microbatches - 1))
+
+    # Split all non-None fields at once using torch.split
```
Hgherzog: Nit: This isn't really splitting all non-None fields at once so remove comment

scripts/vnext/tokenization_ablations/base_all_bands_single_token.py
```diff
@@ -105,12 +107,18 @@ def build_model_config(common: CommonComponents) -> LatentMIMConfig:
 
 
 if __name__ == "__main__":
+    # Use functools.partial to pass tokenization_config to builders that need it.
+    # This ensures the masking strategy uses the same band groupings as the model.
     main(
         common_components_builder=build_common_components,
         model_config_builder=build_model_config,
-        train_module_config_builder=build_train_module_config,
+        train_module_config_builder=partial(
```
Hgherzog: Suggestion: add the tokenization config to the common components.

tests/unit/data/test_dataloader_masking.py
Hgherzog, line 1: Move this into `tests/unit/data/test_dataloader.py` to stick consistently with our 1 module to 1 test file paradigm

olmoearth_pretrain/train/train_module/contrastive_latentmim.py
```diff
@@ -29,32 +31,8 @@
...
-
-        strategy.tokenization_config = tokenization_config
-
-        for child in vars(strategy).values():
-            if isinstance(child, MaskingStrategy):
-                _set_config(child)
-
-    _set_config(masking_strategy)
+# Backward compatibility alias
+_propagate_tokenization_config = propagate_tokenization_config
```
Hgherzog: We do not need this alias if this is in another function

scripts/vnext/tokenization_ablations/base_all_bands_single_token.py
gabrieltseng: moved

olmoearth_pretrain/data/dataset.py
gabrieltseng: removed all 3 imports

scripts/official/script.py
gabrieltseng: Here is a run with fewer data loaders: https://wandb.ai/eai-ai2/2026_01_12_v1.5_release/runs/km3py3wp

model duration is still 99.8 % of the training time.

olmoearth_pretrain/train/train_module/contrastive_latentmim.py
gabrieltseng: removed

olmoearth_pretrain/train/masking.py
gabrieltseng: oh i see. thank you for the clear explanation! updated

olmoearth_pretrain/data/dataset.py
gabrieltseng: moved

olmoearth_pretrain/train/utils.py
gabrieltseng: removed. The intent of the comment was "at once vs. in a loop" but i see how its confusing.

scripts/official/script.py
gabrieltseng: updated

tests/unit/data/test_dataloader_masking.py
gabrieltseng: moved

PR #471, opened 2026-01-26
ERA5 exp

yawenzzzz:


robmarkcole:
@yawenzzzz they use an era5 head in thor https://arxiv.org/pdf/2601.16011

PR #477, opened 2026-02-02
Quantization using distribution of embeddings on pre-training dataset

favyen2:
Try quantizing to 8/4/2/1-bit using distribution of embeddings on pre-training dataset.


olmoearth_pretrain/evals/embedding_transforms.py
```diff
@@ -46,6 +52,112 @@ def dequantize_embeddings(quantized: torch.Tensor) -> torch.Tensor:
...
+                config[key] = {
+                    "quantiles": torch.from_numpy(f[key]["quantiles"][:]),
+                    "midpoints": torch.from_numpy(f[key]["midpoints"][:]),
+                }
+        if "dim" in f:
+            config["dim"] = int(f["dim"][()])
+    return config
+
+
+def quantize_embeddings_percentile(
```
cmwilhelm, line 88: Did this percentile-based bucketing approach behave substantially different than the statistics-naive quantization scheme in https://github.com/allenai/olmoearth_run/blob/006496243c8f00ada3b74a77874e87a93bfa661e/src/olmoearth_run/runner/tools/postprocessors/combine_geotiff.py#L48? 

I magine it's probably pretty important with the very low-bit quantizations, but do you have a sense at int8?

favyen2: AFAIK neither approach showed any drop in performance at int8, here is Mike's results:
https://github.com/allenai/olmoearth_pretrain/blob/main/scripts/archived/2026-01-024_embedding_analysis/quant_comparison_rounded.csv

In the platform I think you should just go with the simple fixed quantization.

favyen2: also I computed the per-band distribution here, but I found that all of the bands follow almost the same distribution

cmwilhelm: Normal distribution centered over 0.0? That would be convenient for int1 😅

favyen2: Something like this:

quantiles_pdf

Definitely centered at 0 so I do think for 1-bit you can just do `< 0` vs `>= 0`. I didn't think it was normal but I guess it does look roughly normal.

cmwilhelm: Guess it doesn't need to be a normal distribution when there are only two quantiles.

olmoearth_pretrain/evals/embedding_transforms.py
```diff
@@ -46,6 +52,112 @@ def dequantize_embeddings(quantized: torch.Tensor) -> torch.Tensor:
...
+
+    The HDF5 file is generated by scripts/tools/compute_embedding_quantiles.py.
+
+    Args:
+        path: Path to quantiles.h5 file
+
+    Returns:
+        Dictionary with keys like "8bit", "4bit", "2bit", "1bit", each containing:
+            - "quantiles": torch.Tensor of shape (dim, num_buckets+1)
+            - "midpoints": torch.Tensor of shape (dim, num_buckets)
```
yawenzzzz, line 72: Suggestion: add a note say that quantiles are for quantization and midpoints are for dequantization

olmoearth_pretrain/evals/embedding_transforms.py
```diff
@@ -46,6 +52,112 @@ def dequantize_embeddings(quantized: torch.Tensor) -> torch.Tensor:
...
+
+    Returns:
+        Float32 tensor of same shape
+    """
+    original_shape = quantized.shape
+    dim = original_shape[-1]
+
+    # Flatten to (N_total, dim)
+    # Convert to uint8 first to handle int8 wrap-around (128-255 stored as -128 to -1)
+    flat = quantized.reshape(-1, dim).to(torch.uint8).long()
```
yawenzzzz, line 148: I'm wondering why don't we just quantize to unit8, with this, there's no need to convert to uint8 in the dequantization

yawenzzzz (approved):
LGTM! just a small question about the int8 vs. unit8

olmoearth_pretrain/evals/embedding_transforms.py
favyen2: added!

olmoearth_pretrain/evals/embedding_transforms.py
favyen2: Currently I think it won't work with uint8 because of the way `midpoints[d_indices, flat]` is being indexed later on. But the cast could be pushed until later if it is using too much memory.

PR #480, opened 2026-02-04
Fix finetune checkpointing

yawenzzzz:
Save both last and best checkpoint, and if a job gets preempted, it can load from last checkpoint and start from last epoch. Test run here: https://beaker.org/orgs/ai2/workspaces/earth-systems/work/01KGJYNQF4P6S9NV396HE5EW2A?taskId=01KGJYNQPWZV36NJ6TM0BPFJRH&jobId=01KGJZ16GTRWP1PXZKZHCGNPT1 (I manually preempted the job and it works well)

### DIFF
--- a/olmoearth_pretrain/evals/finetune/train.py
+++ b/olmoearth_pretrain/evals/finetune/train.py
@@ -66,7 +66,9 @@ def compute_eval_metrics(
     if task_config.task_type == TaskType.CLASSIFICATION:
         val_result = eval_cls(ft, val_loader, device, task_config.is_multilabel)
     else:
-        val_result = eval_seg(ft, val_loader, device, task_config.num_classes, patch_size)
+        val_result = eval_seg(
+            ft, val_loader, device, task_config.num_classes, patch_size
+        )
 
     test_result: EvalResult | None = None
     if test_loader is not None:


olmoearth_pretrain/evals/finetune.py
```diff
@@ -376,6 +461,20 @@ def run_finetune_eval(
             logger.info(
                 f"New best validation metric {best_val_metric:.4f} at epoch {epoch + 1}"
             )
```
Hgherzog: Can we add an integration test for this functionality? (like just checking that checkpoints are saved and discovered by the same run.

Hgherzog (approved):
LGTM! Couple small comments.

olmoearth_pretrain/evals/finetune.py
```diff
@@ -401,6 +500,11 @@ def run_finetune_eval(
     else:
         logger.info("No best checkpoint path provided, skipping saving best checkpoint")
 
+    # Clean up resume checkpoint after successful completion
+    if resume_checkpoint_path and os.path.exists(resume_checkpoint_path):
```
Hgherzog: The best checkpoint is not cleaned up though?

olmoearth_pretrain/train/callbacks/evaluator_callback.py
```diff
@@ -382,6 +382,16 @@ def _get_best_checkpoint_path(self) -> str:
         )
         return best_checkpoint_path
 
+    def _get_resume_checkpoint_path(self) -> str:
+        """Get the resume checkpoint path for resumable training."""
+        resume_checkpoint_path = os.path.join(
+            self.trainer.save_folder,
```
Hgherzog, line 388: We should also include the step from the pretrained checkpoint we are training off of as we may want to finetune for the same run at different points in training

olmoearth_pretrain/evals/finetune.py
yawenzzzz: OK, this make sense, I added the integration test for the saving and loading checkpoint. Also, seems there's a recent test called `test_finetune.py` which is basically testing the newly added segmentation metrics, I renamed it to `test_segmentation_metrics.py` since it's not a specific integration test for finetune.

yawenzzzz: Another thing is that cursor helped with a quick refactor as the finetuning code looks chunky, which i think makes different components more clear.

Hgherzog: I can TAL at this again. I think we should just remove those additional tests in test_finetune.py as we already have unit tests elsewhere for this.

olmoearth_pretrain/evals/finetune.py
yawenzzzz: yes, and this is intentionally, we may need the best checkpoint later to run evaluation on another test set.

olmoearth_pretrain/evals/finetune.py
```diff
@@ -208,6 +210,69 @@ def _set_backbone_trainable(backbone: nn.Module, requires_grad: bool) -> None:
...
+        shutil.move(tmp_path, path)  # Atomic on POSIX
+        logger.info(f"Saved training checkpoint to {path} at epoch {epoch}")
+    except Exception:
+        os.close(tmp_fd)
+        if os.path.exists(tmp_path):
+            os.remove(tmp_path)
+        raise
+
+
+def _load_training_checkpoint(path: str, device: torch.device) -> dict | None:
```
Hgherzog: Suggestion: I would prefer we raise an error here if the path doesn't exist or the checkpoint doesn't load and then catch that outside this function instead of returning None which is pretty opaque.

olmoearth_pretrain/train/callbacks/evaluator_callback.py
yawenzzzz: Oh the step is indeed included, when you pass the checkpoint path to the `full_eval_sweep_finetune.py`, the finetune job name will include the step:
```
--checkpoint_path /weka/.../phase2.0_base_lr0.0001_wd0.02/step667200
run name: phase2.0_base_lr0.0001_wd0.02_step667200_FT_lr0.0001
```
And the finetune checkpoint will be saved per task, sth like:
```
/weka/dfive-default/olmoearth_pretrain/checkpoints/yawenzzzz/phase2.0_base_lr0.0001_wd0.02_step667200_FT_defaults/m_eurosat/lr0.0001/last.ckpt
```

olmoearth_pretrain/evals/finetune.py
yawenzzzz: Fixed: i added a path existence check. If the path doesn’t exist, training starts from fresh, if it does, the checkpoint will be loaded and any load failures or missing keys are reported.

PR #481, opened 2026-02-04
Add py.typed file

Hgherzog:


favyen2 (approved):

favyen2:
Actually do we need to update pyproject.toml similar to https://github.com/allenai/rslearn/blob/master/pyproject.toml

favyen2 (approved):

cmwilhelm (commented):
Does this need to come with a (micro?) version bump for it to be picked up as a new version?

PR #497, opened 2026-02-19
Check eval performance when extracting center token for classification tasks

favyen2:
This adds an option to pool over the center tokens only (across timesteps/modalities) instead of all tokens for classification tasks. The motivation is previous observations that this helps for some tasks like Nandi.

I also added an option to use bilinear interpolation followed by linear layer for segmentation tasks.

It doesn't seem to make a big difference for most tasks: https://wandb.ai/eai-ai2/2026_02_14_eval_changes


yawenzzzz (approved):
LGTM!

PR #507, opened 2026-03-04
Fix WorldCereal data by using SINGLE_COMPOSITE and setting correct NODATA

favyen2:
WorldCereal consists of overlapping GeoTIFFs, but only one of the GeoTIFFs will be valid (not NODATA) at any given location. Previously, this means that matching may pick the wrong GeoTIFF since it is just arbitrarily selecting one that intersects the window. SINGLE_COMPOSITE will put all intersecting GeoTIFFs into the item group, and FIRST_VALID compositing method (along with setting correct nodata value for the materialization) ensures we handle it correctly.

This seems to increase the number of tiles with non-zero WorldCereal data by about 20%.

The change is a simple config change, but the rslearn_to_olmoearth scripts are also updated to latest version of rslearn (where RasterFormat returns RasterArray instead of numpy array).

### DIFF
--- a/data/rslearn_dataset_configs/config_worldcereal.json
+++ b/data/rslearn_dataset_configs/config_worldcereal.json
@@ -6,13 +6,19 @@
           "bands": [
             "tc-annual-temporarycrops-classification"
           ],
-          "dtype": "float32"
+          "dtype": "float32",
+          "nodata_vals": [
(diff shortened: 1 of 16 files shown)


Farbum (approved):
LGTM, thanks!

PR #526, opened 2026-04-06
Remove constraint on max sequence length

yawenzzzz:
When rslearn aligns timestamps across multiple modalities, the resulting temporal sequence can be > 12 timesteps. Previously, `CompositeEncodings` pre-allocated a fixed-size position encoding table based on `max_sequence_length` (default 12), which will raise an error when the input had more timesteps than the table size. The fix here computes the temporal position encoding on-the-fly from the actual number of input timesteps, removing the fixed limit. It's also backward compatible since we directly drop the `pos_embed` parameter from old checkpoints.

### DIFF
--- a/olmoearth_pretrain/nn/flexi_vit.py
+++ b/olmoearth_pretrain/nn/flexi_vit.py
@@ -2,6 +2,7 @@
 
 import logging
 import math
+import warnings
 from dataclasses import dataclass
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


favyen2 (approved):
thanks for updating this!

robmarkcole:
I'm quite keen to test this, @yawenzzzz do you have any results to share?

robmarkcole:
@favyen2 hit the bug this addresses, good to merge? Also needs to be applied in https://github.com/allenai/olmoearth_pretrain_minimal

robmarkcole:
@favyen2 I just hit this with 1_1 via pretrain minimal, be good to patch soon!

yawenzzzz:
> I'm quite keen to test this, @yawenzzzz do you have any results to share?

I don't have major results for this - this is a bug i encountered when running with more than 12 timesteps. I will merge it now, also with our new change in position encoding (to remove tiling artifacts) - using RoPE instead of the sincos, there is no max length limit with it too.

PR #531, opened 2026-04-09
minor readability update:  flexi vit

Farbum:


olmoearth_pretrain/nn/flexi_vit.py
```diff
@@ -649,12 +649,12 @@ def __init__(
         # 0.25 of the dimension
         self.embedding_dim_per_embedding_type = int(embedding_size * 0.25)
         # Position encodings for time dimension initialized to 1D sinusoidal encodings
-        self.pos_embed = nn.Parameter(
+        self.register_buffer(
+            "pos_embed",
```
gabrieltseng, line 653: I think this won't affect old checkpoints but did you run a check to confirm (i.e. we can still load existing checkpoints with this change)?

gabrieltseng (approved):

PR #532, opened 2026-04-09
Fix one more edge case for finetuning with preemption

yawenzzzz:
One more edge case if the job gets preempted: 

When a job is preempted after all finetune epochs complete but before the best checkpoint is saved, the resumed run skips the training loop (since all epochs are done), leaving val_result as None, which blocks both val and test metrics from being logged to wandb.

Also, removed some redundant code.

### DIFF
--- a/olmoearth_pretrain/internal/all_evals.py
+++ b/olmoearth_pretrain/internal/all_evals.py
@@ -476,6 +476,7 @@ FT_EVAL_TASKS = {
         epochs=50,
         primary_metric=EvalMetric.MIOU,
     ),
+    # Cashew plant requires a larger patch size; 16 performed best.
     "m_cashew_plant": DownstreamTaskConfig(
         dataset="m-cashew-plant",
         ft_batch_size=4,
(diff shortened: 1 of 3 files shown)


yawenzzzz:
@gabrieltseng i ran into this issue when testing the single bandset finetuning, and now the finetuning runs should be more robust to preemption. 

btw, the is the command I ran finetuning:
```
python olmoearth_pretrain/internal/full_eval_sweep_finetune.py \
  --checkpoint_path /weka/dfive-default/olmoearth_pretrain/checkpoints/yawenzzzz/single_bandset_no_s1_drop_random_time_dropout_0.2/step600000 \
  --module_path scripts/vnext/single_bandset_band_dropout/base_band_dropout_no_s1_drop_random_time.py \
  --cluster ai2/jupiter \
  --project_name 2026_04_08_v1.1_finetune \
 ```

**Review by gabrieltseng — APPROVED** [review]

File: olmoearth_pretrain/internal/all_evals.py [hunk 1]
```diff
@@ -504,6 +485,7 @@ def load_user_module(path: str) -> Any:
         norm_stats_from_pretrained=False,
         norm_method=NormMethod.NORM_NO_CLIP_2_STD,
         epochs=50,
+        patch_size=16,
```
gabrieltseng, line 488: should we sweep this like we did in v1? or do you think its fine to just run ps16

yawenzzzz: i don't think we need to sweep it given the patch analysis we have, right now, with the single bandset checkpoints, i got 80.5 on test for cashewplant which is better than v1. Also, the numbers we reported in v1 are also using patch size 16.

PR #535, opened 2026-04-15
Fix copy_wandb_run script

yawenzzzz:
Our previous script is not working, Hadrien tried to copy this run `base_token_masked_all` but failed. I asked cursor to update the script, and it's working now.


Farbum (approved):
LGTM, tested it and it ran smoothly.

favyen2 (approved):
I ran into this too and gabi pointed me here. the changes look good to me, thanks for the fix, and the improvements (tqdm while iterating over the metrics, and having the ability to pick a few metrics so we can skip the GPU metrics) are great! I fixed lint error and will merge once the checks are passing.

PR #538, opened 2026-04-16
rslearn task ingestion: verbose and tag filtering

Farbum:
Process was silent -> added a bunch of logs / tqdm progress bars to track progress
Provided tags were filtered after copying windows which is not ideal for huge dataset. The windows are now filtered before copying starts.


Hgherzog:
@Farbum  Can you link an example run of this eval with the base model?

Farbum:
> @Farbum Can you link an example run of this eval with the base model?

Yes I will do that soon but this PR should be independent from any task, it is just refining the conversion logic

Hgherzog (approved):
LGTM, probs best to not push the new dataset registry info until the dataset is ready

PR #541, opened 2026-04-21
new modality - era5 land daily 

Farbum:
Adds era5 land daily data as new modality

### DIFF
--- a/olmoearth_pretrain/data/constants.py
+++ b/olmoearth_pretrain/data/constants.py
@@ -440,6 +440,36 @@ class Modality:
         image_tile_size_factor=-256,
     )
 
+    ERA5L_DAY_10 = ModalitySpec(
+        name="era5l_day_10",
+        # ERA5-Land daily 9 km/pixel bands that we store at 2.56 km/pixel, T=720 daily timesteps, non-spatial.
+        tile_resolution_factor=16,
+        band_sets=[
+            BandSet(
+                [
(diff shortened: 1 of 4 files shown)


robmarkcole:
Re tasks which could benefit from pretraining, consider soil moisture and crop yield - both strongly driven by weather of course

Farbum:
> Re tasks which could benefit from pretraining, consider soil moisture and crop yield - both strongly driven by weather of course

Working on this :)

olmoearth_pretrain/data/constants.py
```diff
@@ -440,6 +440,36 @@ class Modality:
...
+                    "sp",
+                    "ssr",
+                    "ssrd",
+                    "str",
+                    "swvl1",
+                    "swvl2",
+                    "t2m",
+                    "tp",
+                    "u10",
+                    "v10",
```
yawenzzzz: Could you add a short comment for these shortnames? people may be curious about what each variable mean in ERA5 daily data.

yawenzzzz (approved):
LGTM! Just one minor comment regarding the variables.

yawenzzzz:
@cursoragent Could you do a final review of this PR?

olmoearth_pretrain/data/constants.py
Farbum: Claude did :)

PR #519, opened 2026-03-30
Single bandset ablations

gabrieltseng:
Current best recipe is time masking of 0.5

PR #547, opened 2026-04-28
Update `st_model.py`

favyen2:
`st_model.py` was no longer working due to other changes. This updates it to sync with recent changes.

I also added some scripts for windowed attention experiments using STEncoder.

Depends on #519.


scripts/archived/2026_04_14_misc_single_bandset/base_no_discrim_mse.py
Hgherzog, line 1: Rather than update archived scripts, I would rather just launch a couple new st model runs to verify this still works and there is something not in archvied to build off of.

Hgherzog:
Can you share an example run with this to show it works?

olmoearth_pretrain/train/loss.py
```diff
@@ -991,6 +991,32 @@ def compute(
         return F.l1_loss(pred, target)
 
 
+@LOSS_REGISTRY.register("cosine_similarity")
```
Hgherzog, line 994: Is this something we are still actively looking into and investigating? If not, it would be better to not add it back IMO

favyen2:
> Can you share an example run with this to show it works?

https://wandb.ai/eai-ai2/2026_04_24_st_model/runs/hsqc93mb

By not working I meant it runs into errors because the code was not compatible with other changes in the codebase. So the run just shows that it is training without errors. The errors happen on the dry run though, so the run isn't really showing anything useful. But anyway that is one of the runs I started.

olmoearth_pretrain/train/loss.py
favyen2: Yeah the cosine similarity seems to perform better than L2 loss. I am looking into the influence on embeddings to not have patch discrimination effect of making pushing embeddings apart in areas that look quite uniform. I can move it to a different PR if desired though.

scripts/archived/2026_04_14_misc_single_bandset/base_no_discrim_mse.py
favyen2: Yes these are new runs, I am not updating old archive scripts.

Hgherzog (approved):
LGTM, only comment is it would be good to have the current working script in vnext instead of archived.

favyen2:
> LGTM, only comment is it would be good to have the current working script in vnext instead of archived.

OK I will copy one of the scripts to vnext directory and make sure it is consistent with the hidden1 script.

favyen2:
Or I guess it will be compatible with the older single bandset band dropout model, I think the hidden model is not in vnext yet. I think it makes sense to put stuff in archived or similar directory until we are confident we want to incorporate it into the release, so probably we will copy the hidden1 over to vnext when it is finalized, but anyway for the windowed attention model I will create a config for scripts/vnext/single_bandset_band_dropout/.

scripts/archived/2026_04_14_misc_single_bandset/base_no_discrim_mse.py
favyen2: added `scripts/vnext/single_bandset_band_dropout/st_encoder_windowed_attention.py`

Hgherzog:
@favyen2 The convention I was following is scripts/official has the stuff that is released (experiments abaltions etc) and vnext is just general stuff we are actively working on. archived is stuff we arent really working on actively right now.

PR #552, opened 2026-05-05
Use nodata_value instead of nodata_vals.

favyen2:
nodata_vals is deprecated but also throwing error because https://github.com/allenai/rslearn/pull/624 has still not been merged.


piperwolters (approved):
looks good to me

PR #555, opened 2026-05-07
add skip task names arg to finetune sweep script

piperwolters:
For convenience, copy code from the full_eval_sweep.py that allows you to skip FT eval tasks.


yawenzzzz (approved):
LGTM!

PR #556, opened 2026-05-09
GEO-Bench-2, UNet finetuning head, Finetuning dice loss

piperwolters:
- Add GEO-Bench-2 tasks that 1) don't already exist and 2) aren't object detection tasks
   - I was originally not including high-res source imagery tasks, but olmoearth does surprisingly well with not being trained with it, so going to keep those
- Add a regression TaskType to accommodate biomassters task ---> merged in Henry's regression implementation instead
- Add handling for multiple evals running locally on the same node 
- Updating dependencies -- not sure why we're pinning lightning currently?
- "UNet" (no skip connections) finetuning head
   - Tried FPN and a few variations,  but keeping just this one in this PR 
- Further dice loss functionality for finetuning


olmoearth_pretrain/data/normalize.py
```diff
@@ -111,6 +111,13 @@ def _normalize_computed(
             std_vals.append(std_val)
         min_vals = np.array(mean_vals) - self.std_multiplier * np.array(std_vals)
         max_vals = np.array(mean_vals) + self.std_multiplier * np.array(std_vals)
+        # Some datasets (e.g. biomassters S1) have more channels than the modality's
+        # band_order (e.g. 4 channels = ascending + descending VV/VH). Tile stats to match.
+        n_channels = data.shape[-1]
```
favyen2: How does this work later when applying the model? If we're only using the first two bands, then it could make sense to move this to the dataset, similar to for PASTIS where it repeats two of the bands that are missing from PASTIS.

olmoearth_pretrain/evals/datasets/geobench_v2_dataset.py
```diff
@@ -0,0 +1,335 @@
...
+from olmoearth_pretrain.evals.metrics import SEGMENTATION_IGNORE_LABEL
+from olmoearth_pretrain.evals.task_types import TaskType
+from olmoearth_pretrain.train.masking import MaskedOlmoEarthSample
+
+logger = logging.getLogger(__name__)
+
+_S2_ORDER = list(Modality.SENTINEL2_L2A.band_order)
+
+
+def _s2_names(band_order: Any) -> list[str]:
```
favyen2, line 60: Do you need to support both of these or could you have them all be the first method (map from modality -> band order)

olmoearth_pretrain/evals/datasets/geobench_v2_loaders.py
```diff
@@ -0,0 +1,457 @@
...
+        torch.zeros((0, 4), dtype=torch.float32),
+        torch.zeros((0,), dtype=torch.int64),
+        torch.zeros((0, img_h, img_w), dtype=torch.int64),
+    )
+
+
+class _BaseGeobenchDataset(Dataset):
+    """Loads a .tortilla index, filters by split, provides helpers."""
+
+    TORTILLA: str | list[str]
```
favyen2: Similar here, seems easier to always make it a list.

olmoearth_pretrain/evals/knn.py
```diff
@@ -352,7 +352,9 @@ def _run_knn_for_k(
     """
     train_embeddings = train_embeddings.to(device)
     test_embeddings = test_embeddings.to(device)
-    train_labels = train_labels.to(device)
+    # Flatten to (N,): (N, 1) labels would make train_labels[top_k_indices] shape
+    # (..., 1), causing one_hot to produce (..., 1, C) and break the weight broadcast.
+    train_labels = train_labels.to(device).reshape(-1)
```
favyen2: For this one, is it possible to move the reshaping to the dataset so we don't need to worry about different shapes here? Or is it something else (I didn't fully understand the need)?

piperwolters: Yes definitely. This was a hack for biomassters to work, meant to move this edge case into that dataset

olmoearth_pretrain/evals/datasets/geobench_v2_dataset.py
```diff
@@ -0,0 +1,336 @@
...
+    band_order: Any,
+    slug: str,
+) -> OlmoEarthSample:
+    device = next(iter(sample.values())).device
+
+    if slug == "caffe":
+        # Grayscale aerial [0–255]; broadcast to all 12 S2 channels and rescale
+        # to S2 DN range so the OlmoEarth normalizer (mean ~1188) sees valid signal.
+        filled = sample["image"].float()[:1].repeat(len(_S2_ORDER), 1, 1)
+        filled = filled * (10000.0 / 255.0)
```
Hgherzog: Nit: Use constant for S2 dynamic range max

olmoearth_pretrain/evals/datasets/geobench_v2_dataset.py
```diff
@@ -0,0 +1,336 @@
...
+
+
+def _sample_to_olmoearth(
+    sample: dict[str, torch.Tensor],
+    band_order: Any,
+    slug: str,
+) -> OlmoEarthSample:
+    device = next(iter(sample.values())).device
+
+    if slug == "caffe":
```
Hgherzog: Nit: Make an ENUM for all the task names we are doing control flow stuff on

olmoearth_pretrain/evals/datasets/geobench_v2_dataset.py
```diff
@@ -0,0 +1,336 @@
...
+        hwtc = _bchw_to_hwtc(filled)
+        return OlmoEarthSample(
+            sentinel2_l2a=hwtc, timestamps=_timestamps(hwtc.shape[2], device)
+        )
+
+    if slug in ("burn_scars", "cloudsen12", "spacenet7"):
+        image = sample["image"].float()
+        src = _s2_names(band_order)
+        if slug == "burn_scars":
+            # GeoBench2 stores HLS reflectance in [0, 1]; OlmoEarth pretraining
```
Hgherzog, line 183: Is this Harmonized Landsat Sentinel data or just Sentinel 2 data?

olmoearth_pretrain/evals/datasets/geobench_v2_dataset.py
```diff
@@ -0,0 +1,336 @@
...
+    image: torch.Tensor, src_names: list[str], device: torch.device
+) -> OlmoEarthSample:
+    """Build a single-modality S2 OlmoEarthSample from an arbitrary-band image."""
+    hwtc = _bchw_to_hwtc(_align_s2_to_sentinel2_l2a(image.float(), src_names))
+    return OlmoEarthSample(
+        sentinel2_l2a=hwtc, timestamps=_timestamps(hwtc.shape[2], device)
+    )
+
+
+def _sample_to_olmoearth(
```
Hgherzog, line 152: I think we should have another clear way to store what the source imagery of the dataset is if it is not directly one of the sources we are training against, rather than just making a comment in the code. Perhaps systemizing it in a config somewhere

olmoearth_pretrain/evals/datasets/geobench_v2_dataset.py
```diff
@@ -0,0 +1,336 @@
...
+        return len(self._inner)
+
+    def __getitem__(self, idx: int) -> tuple[MaskedOlmoEarthSample, torch.Tensor]:
+        """Load, normalize, and mask sample at idx, returning (masked_sample, label)."""
+        raw = self._inner[idx]
+        device = next(
+            (v.device for v in raw.values() if torch.is_tensor(v)), torch.device("cpu")
+        )
+        olmo = _sample_to_olmoearth(raw, self._band_order, self._slug)
+        olmo = _apply_olmoearth_normalization(olmo, self._normalizer)
```
Hgherzog: We should also have the ability to use dataset normalization stats for each of these so we can compare against other models

olmoearth_pretrain/evals/datasets/geobench_v2_dataset.py
```diff
@@ -0,0 +1,336 @@
...
+
+    if slug in ("burn_scars", "cloudsen12", "spacenet7"):
+        image = sample["image"].float()
+        src = _s2_names(band_order)
+        if slug == "burn_scars":
+            # GeoBench2 stores HLS reflectance in [0, 1]; OlmoEarth pretraining
+            # stats are in DN scale (~0–10000), so rescale to match.
+            image = image * 10000.0
+        if slug == "cloudsen12":
+            image = image[: len(src)]  # tortilla has 14 bands; truncate to 12
```
Hgherzog, line 187: What are the 14 Bands?

olmoearth_pretrain/evals/datasets/geobench_v2_dataset.py
```diff
@@ -0,0 +1,336 @@
...
+        dtype=x.dtype,
+        device=x.device,
+    )
+    for j, name in enumerate(target_names):
+        if name in idx:
+            out[:, j] = x[:, idx[name]]
+    return out
+
+
+def _align_s2_to_sentinel2_l2a(
```
Hgherzog, line 89: I am concerned that zeroing out the values before normalization is problematic as the hidden layer that combines the bands will see it as a large negative value rather than zero. If we zeroed after normalization, it would be more similar to the band dropout we see in training. Another option is to do some sort of band imputation but that is probably not ideal.

Hgherzog: @piperwolters this may be an explanation for the burn scars performance discrepancy

olmoearth_pretrain/evals/datasets/geobench_v2_dataset.py
piperwolters: 12 sen2 bands (everything but band10) + s1 vv and vh

olmoearth_pretrain/evals/finetune/unet_head.py
```diff
@@ -0,0 +1,77 @@
...
+        self.up_blocks = nn.ModuleList()
+        for i in range(n_stages):
+            self.up_blocks.append(
+                nn.Sequential(
+                    nn.Upsample(scale_factor=2, mode="bilinear", align_corners=True),
+                    _ConvBnRelu(dims[i], dims[i + 1]),
+                )
+            )
+
+        self.head = nn.Conv2d(dims[n_stages], num_classes, kernel_size=1)
```
robmarkcole: Could make this more general to support regression also

olmoearth_pretrain/evals/datasets/geobench_v2_dataset.py
piperwolters: Ok added a SourceImagery field, that isn't actually used for deciding normalization or anything, but does tell us what the imagery source is for each task.

olmoearth_pretrain/evals/datasets/geobench_v2_dataset.py
piperwolters: Harmonized Landsat Sentinel data

olmoearth_pretrain/evals/datasets/geobench_v2_dataset.py
piperwolters: Okay, I wired up these GeoBench2 datasets to use per-dataset normalization stats when available, falling back to the current OlmoEarth pretraining stats otherwise. I'm not sure if this is the best way to do this if we don't always have stats (maybe we should always have stats)?

robmarkcole:
@piperwolters could you comment on how the newly contribute unet differs from `rslearn.models.unet.UNetDecoder`? In my own experiment, the result with this new implementation was slightly worse

piperwolters:
> @piperwolters could you comment on how the newly contribute unet differs from `rslearn.models.unet.UNetDecoder`? In my own experiment, the result with this new implementation was slightly worse

Favyen had reported lower results with that UNet than the linear head, so I was just trying a few other things. Good to know that you're seeing worse performance with this one, I'll try the rslearn architecture and compare performance on these geobench2 tasks!

piperwolters:
> @piperwolters could you comment on how the newly contribute unet differs from `rslearn.models.unet.UNetDecoder`? In my own experiment, the result with this new implementation was slightly worse

Results comparing the architecture I had pushed before and a unet closer to the rslearn version:
geobench_unet_comparison

Based on this, I'll go with the rslearn version.

olmoearth_pretrain/evals/datasets/config/geobench2_norm_stats.json
```diff
@@ -0,0 +1,14 @@
+{
+  "caffe": {
+    "sentinel1": {
+      "vh": {
+        "mean": 0.0,
+        "std": 1.0
+      },
+      "vv": {
+        "mean": 70.898598,
+        "std": 84.454284
```
yawenzzzz: Why the vv and vh stats look so different here?

olmoearth_pretrain/evals/finetune/train.py
```diff
@@ -115,6 +146,13 @@ def compute_eval_metrics(
                 primary_metric=primary_metric,
                 primary_metric_class=primary_metric_class,
             )
+        elif task_config.task_type == TaskType.REGRESSION:
+            test_result = eval_reg(
+                ft,
+                test_loader,
+                device,
+                primary_metric=primary_metric,
+            )
```
yawenzzzz, line 155: Great to have the regression tasks!

olmoearth_pretrain/train/callbacks/evaluator_callback.py
```diff
@@ -85,7 +85,11 @@ class DownstreamTaskConfig:
     # FT
     ft_lr: float | None = None
     ft_batch_size: int = 32
+    ft_grad_accum_steps: int = 1
     finetune_seed: int = 42
+    ft_head_type: str = (
+        "linear"  # "linear" or "unet" (unet only valid for segmentation/regression)
+    )
```
yawenzzzz, line 92: Thanks for adding the UNet head for finetuning!

olmoearth_pretrain/internal/all_evals.py
```diff
@@ -386,6 +386,181 @@ def load_user_module(path: str) -> Any:
         epochs=50,
         eval_mode=EvalMode.LINEAR_PROBE,
     ),
+    # GeoBench v2 (`gb2-*` datasets; keys use underscores for Hydra)
```
yawenzzzz, line 389: Is there any tasks under GeoBench v2 that are really slow to run? e.g., the dataset is large or sth, we reduce the cadence of running PASTIS to every 50 epochs just because it's very slow to run

olmoearth_pretrain/internal/all_evals.py
```diff
@@ -386,6 +386,181 @@ def load_user_module(path: str) -> Any:
...
+    "gb2_benv2": DownstreamTaskConfig(
```
yawenzzzz, line 390: also, have you tried other normalization approaches? (you may have mentioned this during the meeting) previously, there're some tasks like MADOS and also forestnet, which really requires dataset normalization stats, without that, the performance will be much worse.

yawenzzzz (approved):
LGTM! Left a few comments, if you take a look that will be great. Thanks for adding GeoBench v2!

yawenzzzz:
@cursoragent Can you do a final pass on this PR? see if there's any major issues

olmoearth_pretrain/evals/datasets/config/geobench2_norm_stats.json
piperwolters: The Caffe task only provided 1 SAR amplitude band, so I have just put it in vv here, and pad vh with zeros so that I can pass it in as Sentinel1 imagery. Does that seem ok or should I make a "new" modality for this task?

olmoearth_pretrain/internal/all_evals.py
piperwolters: Ah good point. I will change the eval cadence to 50 for benv2, biomassters, cloudsen12, kuro siwo and spacenet2!

olmoearth_pretrain/evals/datasets/config/geobench2_norm_stats.json
yawenzzzz: ah i see, that makes sense! Yes, this seems OK but would be helpful if you add a note or sth to the Caffe task so that other ppl can be aware of this issue (could be in `all_evals.py`)

olmoearth_pretrain/internal/all_evals.py
yawenzzzz: Nice!

olmoearth_pretrain/internal/all_evals.py
piperwolters: I have tried the dataset normalization approach and didn't get better performance (torchgeo-bench uses normalization similar to our default), but I have now added the dataset stats for each geobench2 task so if `norm_stats_from_pretrained=False` then dataset stats will be used else the default is used (pretraining dataset stats).

yawenzzzz: Thanks a lot!

PR #559, opened 2026-05-12
UNet head for finetuning

piperwolters:


robmarkcole:
@piperwolters could this be implemented in https://github.com/allenai/olmoearth_pretrain_minimal? then we could reuse in torchgeo

piperwolters:
> @piperwolters could this be implemented in https://github.com/allenai/olmoearth_pretrain_minimal? then we could reuse in torchgeo

yes! I will do that once this is cleaned up and merged.

PR #560, opened 2026-05-14
Add dice loss option to finetune args

piperwolters:


olmoearth_pretrain/evals/linear_probe.py
```diff
@@ -417,18 +417,25 @@ def weighted_dice_loss(
...
     )
     weights = weights / (weights.sum() + 1e-8)
 
-    loss = 1.0 - (weights * dice_per_class).sum()
-    return loss
+    dice_loss = 1.0 - (weights * dice_per_class).sum()
+
+    # CE loss provides gradient even when the minority class is absent from
+    # the batch, preventing the model from collapsing to all-background.
+    ce_loss = F.cross_entropy(logits, targets_masked, ignore_index=ignore_index)
```
Farbum: Would the Cross entropy loss here ignore the wrong index, since we are passing "targets_masked" which has already reassigned the ignore_index to 0? Should we pass targets instead of targets_masked?

piperwolters: You are so right!

olmoearth_pretrain/evals/linear_probe.py
```diff
@@ -417,18 +417,25 @@ def weighted_dice_loss(
 
     dice_per_class = (2.0 * intersection + smooth) / (cardinality + smooth)
 
-    # Class weights: inverse frequency of valid pixels per class
+    # Class weights: inverse frequency of valid pixels per class.
```
Farbum, line 420: minor remark: would it make sense to lower the smoothing factor? for heavy imbalanced classes (and I suppose the dice loss is implemented for these conditions), if a handful of minority class pixels are present in the window, then the smoothing factor could alter the loss significantly? Rslearn uses 1e-7, though I don't know if we need something that low.

PR #561, opened 2026-05-14
Add pretrain auxiliary probes and unify low-label eval config

Hgherzog:
## What changed

- Add S2-only pretrain auxiliary probes (worldcover, osm, srtm, canopy, cdl, worldcereal) with random and geographic holdout variants.
- Missing-timestep-aware masking for pretrain subset eval samples so `MISSING_VALUE` tokens are masked instead of fed to the model.
- Unify low-label eval configuration behind a single `label_fraction` field on `DownstreamTaskConfig`. Removes the old user-facing `partition` config and `--label_percentages` sweep arg. `label_fraction` scales train labels only; valid/test stay full size.
- Stream eval metrics to W&B per evaluator (with \`checkpoint_step\`) instead of buffering all tasks for a checkpoint.

W&B example runs are incoming.


olmoearth_pretrain/evals/datasets/pretrain_subset.py
```diff
@@ -52,25 +101,336 @@ def __init__(
...
+        if self.target_modality == OSM_TARGET_MODALITY:
+            return self._osm_label(label)
+        if self.target_modality == SRTM_TARGET_MODALITY:
+            return self._srtm_label(label)
+        if self.target_modality == WRI_CANOPY_TARGET_MODALITY:
+            return self._canopy_label(label)
+        if self.target_modality == CDL_TARGET_MODALITY:
+            return self._cdl_label(label)
+        if self.target_modality == WORLDCEREAL_TARGET_MODALITY:
+            return self._worldcereal_label(label)
```
gabrieltseng, line 394: nit: IMO, these don't need to be their own functions. i think it would be easier to follow if the logic was just in these `if` functions

olmoearth_pretrain/evals/datasets/pretrain_subset.py
```diff
@@ -52,25 +101,336 @@ def __init__(
...
+                    latlons=self._dataset.latlon_distribution,
+                    candidate_positions=eligible_positions,
+                    split=split,
+                    seed=label_seed,
+                    train_samples=train_samples,
+                    valid_samples=valid_samples,
+                    test_samples=test_samples,
+                    bin_size_deg=geographic_bin_size_deg,
+                ).tolist()
+            else:
```
gabrieltseng, line 165: It seems like it might be expensive to recompute these splits every time we launch the evals. 

If we add this to our in-loop evals (which i'd like to) , then we'd be calling `read_csv` tens of times (3 x targets) per eval cycle (~hundreds of Mbs of reads, since its 32Mb for this particular dataset)

Since we use a fixed dataset (in all_evals.py) i wonder if its worth just storing some index npys somewhere. 

This also makes it a bit easier to investigate what this eval looks like after the fact.

gabrieltseng: or just wrap the read_csv in an lru cache as a quicker fix

olmoearth_pretrain/internal/all_evals.py
```diff
@@ -404,6 +403,163 @@ def load_user_module(path: str) -> Any:
...
+# from the osm_sampling pretraining corpus used in scripts/official/*. Using
+# osmbig keeps WorldCover/OSM/SRTM probes out-of-sample. The other map
+# modalities (CDL, WORLDCEREAL, WRI canopy) aren't present in osmbig, so their
+# probes fall back to PRETRAIN_SUBSET_H5PY_DIR (in-distribution).
+PRETRAIN_AUX_EVAL_H5PY_DIR = "/weka/dfive-default/helios/dataset/osmbig/h5py_data_w_missing_timesteps_zstd_3_128_x_4/landsat_openstreetmap_raster_sentinel1_sentinel2_l2a_srtm_worldcover/1297928"
+
+MAP_MODALITY_PROBE_INPUTS = [
+    Modality.SENTINEL2_L2A.name,
+]
+MAP_MODALITY_PROBE_INPUT_SUFFIX = "s2"
```
gabrieltseng: is there a reason to have a different suffix (and another constant) rather than just using `sentinel2_l2a` from Modality.SENTINEL2_L2A.name?

olmoearth_pretrain/internal/all_evals.py
```diff
@@ -404,6 +403,163 @@ def load_user_module(path: str) -> Any:
...
+        input_modalities=MAP_MODALITY_PROBE_INPUTS,
+        epochs=50,
+        eval_mode=EvalMode.LINEAR_PROBE,
+        probe_lr=0.01,
+        primary_metric=primary_metric,
+        h5py_dir=h5py_dir,
+        pretrain_target_modality=target_modality,
+        pretrain_train_samples=6144,
+        pretrain_valid_samples=3072,
+        pretrain_test_samples=3072,
```
gabrieltseng, line 447: is there a reason these specific numbers were picked ?

olmoearth_pretrain/evals/datasets/__init__.py
```diff
@@ -33,29 +33,88 @@ class EvalDatasetPartition(StrEnum):
...
 
+LABEL_FRACTION_TO_PARTITION: dict[float, str] = {
+    0.01: EvalDatasetPartition.TRAIN_001X,
+    0.02: EvalDatasetPartition.TRAIN_002X,
+    0.05: EvalDatasetPartition.TRAIN_005X,
+    0.10: EvalDatasetPartition.TRAIN_010X,
+    0.20: EvalDatasetPartition.TRAIN_020X,
+    0.50: EvalDatasetPartition.TRAIN_050X,
+    1.00: EvalDatasetPartition.TRAIN1X,
+}
```
gabrieltseng: This is confusing. If we are going to swap to using floats in `get_eval_dataset` (which i think is the right call) then we shouldn't need this dict at all. 

`EvalDatasetPartition` is really just a hangover from GeoBench - all other tasks could support arbitrary partitions. 

I think we should just have floats by default, and relegate the EvalDatasetPartition object to just be within the GeoBench dataset class. 

I guess the only risk is that some random fraction (0.06) gets called which isn't supported by all datasets because of GeoBench, but I don't think that justifies the concept leakage here.

olmoearth_pretrain/evals/datasets/pretrain_subset.py
gabrieltseng: Having them like this makes them easier to unit test, leaving.

olmoearth_pretrain/evals/datasets/pretrain_subset.py
gabrieltseng: added

olmoearth_pretrain/internal/all_evals.py
gabrieltseng: (no but its fine for now)

gabrieltseng (approved):

PR #563, opened 2026-05-18
Mitigate memory leak in pre-training runs

favyen2:
The dataloader workers have some memory leak, maybe a leak internal to the h5py library, maybe something else.

We already use non-persistent workers (we actually pass `persistent_workers=True`, but we reset the dataloader each epoch which is effectively the same as having `persistent_workers=False`), but `torch.multiprocessing.set_sharing_strategy("file_system")` causes accumulation of `torch_shm_manager` processes -- they are supposed to exit when the corresponding dataloader process exits, but they don't, maybe due to the memory leak.

So removing this fixes the problem. I considered moving it to run when the evaluation dataloader processes start, but I wasn't able to reproduce file descriptor limit issues, and also it might still cause accumulation of the `torch_shm_manager` processes anyway.

### DIFF
--- a/olmoearth_pretrain/evals/datasets/geobench_dataset.py
+++ b/olmoearth_pretrain/evals/datasets/geobench_dataset.py
@@ -7,7 +7,7 @@ from types import MethodType
 
 import matplotlib.pyplot as plt
 import numpy as np
-import torch.multiprocessing
+import torch
 from einops import repeat
 from geobench.dataset import Stats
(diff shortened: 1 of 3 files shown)


piperwolters (approved):
Thanks for putting together the summary and fixing it! This makes sense with what I was seeing re: dataloading being the issue, but I didn't try both eliminating torch_shm_manager and making it non persistent at the same time.

PR #569, opened 2026-05-29
Add lfmc and single probe for regression tasks

Farbum:
**Summary**
Add LFMC regression eval tasks and improve probe architecture for regression

**Adding LFMC evaluation tasks**
- Added three new LFMC (Live Fuel Moisture Content) downstream evaluation dataset: lfmc_woody_3k
- Configured as per-pixel regression tasks using Sentinel-2 L2A with patch_size=4, pretrain normalization, and spatial splits
- Registered in the eval dataset registry with dataset-specific norm stats and split metadata

**Refactoring the linear probe for regression tasks**
- Introduced BilinearConvProbe: a new probe type that bilinearly upsamples patch embeddings to full resolution and applies a shared 1×1 conv, replacing the old per-patch blockwise linear, matching rslearn for regression task
- The shared conv mapping is trained by every valid pixel across all windows, providing much better regularization for sparse point-label regression targets like LFMC
- All regression tasks now auto-upgrade from LINEAR to BILINEAR_CONV by default — no per-task config change needed
- Removed dead code paths (the old blockwise-rearrange logic for regression) that became unreachable after the auto-upgrade

image


piperwolters (approved):
Thanks for adding functionality to the rslearn dataset and removing the old code! The bilinear conv probe makes sense to me for sparse prediction tasks.  I ran it with the [dense] biomassters task and saw surprisingly similar performance to the other probe.

PR #570, opened 2026-06-02
Add 50 cities dataset

gabrieltseng:
Example run: 

https://beaker.org/orgs/ai2/workspaces/earth-systems/work/01KT4T44VD6KVK2T1K0J83F5W8?taskId=01KT4T45160CAMBHFYQ79CCQTR&jobId=01KT4T455696WZ6ETF067BRPQQ

Dataset:

https://openaccess.thecvf.com/content/CVPR2026W/EarthVision/papers/Weitzel_50_Cities_to_visit_before_you_classify_-_A_world-wide_CVPRW_2026_paper.pdf

License: https://creativecommons.org/licenses/by/4.0/

This is not apples-to-apples with the paper, but its comparable enough I think. One main difference is that in the continent splits, the paper trains on one continent and evals on another. 

We have implemented 3 modalities:
S1, S2, S1 + S2

And three splits:
Random, by city, and by continent.

Overall, the numbers look reasonable :

split | modality | rope | 
-- | -- | -- | 
random | S2 | 0.6366 | 
random | S1 | 0.4535 | 
random | S1+S2 | 0.6731 |
by_city | S2 | 0.3970 | 
by_city | S1 | 0.2678 | 
by_city | S1+S2 | 0.3960 | 
by_continent | S2 | 0.3555 | 
by_continent | S1 | 0.1816 | 
by_continent | S1+S2 | 0.3682 |

The val / test cities / continents are random but consistent across runs. Specifically, we assign:

Val continent=['Africa'] | cities=['Cairo', 'Cape Town', 'Johannesburg', 'Nairobi']
Test continent=['South America'] | cities=['Buenos Aires', 'Itaituba', 'Lima', 'Rio de Janeiro', 'Santiago']

Val cities=['Buenos Aires', 'Cairo', 'Oslo', 'Shanghai', 'Toronto']
Test cities=cities=['Barcelona', 'Istanbul', 'Itaituba', 'Reykjavik', 'Wuhan']

(So for the continent split we train on [Asia, Europe, North America, Oceania])

The table above reports val results.


Farbum:
Would it make sense to sweep models and figure out which flavor (split strategy) shows the strongest discrimination across model?
Have you ran a checkpoint sweep to see how the model improves with pre-training?

gabrieltseng:
> Would it make sense to sweep models and figure out which flavor (split strategy) shows the strongest discrimination across model?

yes this is a good idea! but since its expensive to do I'd prefer to do it post-review once we are happy with the overall implementation. The point is partly to discriminate but mostly to geographically diversify our evals.

gabrieltseng:
> Have you ran a checkpoint sweep to see how the model improves with pre-training?

Running now !

Farbum:
Looks good to me!
In the current state, I think the city split is probable the most informative for our eval diversity goals.
I think the ideal set-up here would be a leave-one-continent-out averaged across all 6 folds but this is a broader discussion about our in-loop eval pipeline.

Farbum (approved):

olmoearth_pretrain/evals/datasets/configs.py
```diff
@@ -236,6 +236,40 @@ def from_dict(cls, d: dict[str, Any]) -> "EvalDatasetConfig":
...
+    # S1+S2 runs are separate EVAL_TASKS in all_evals.py.
+    "fifty_cities": EvalDatasetConfig(
+        task_type=TaskType.SEGMENTATION,
+        imputes=[],
+        num_classes=13,
+        is_multilabel=False,
+        height_width=64,
+        supported_modalities=[Modality.SENTINEL2_L2A.name, Modality.SENTINEL1.name],
+        timeseries=False,
+    ),
```
yawenzzzz, line 255: either add a note to indicate the random split or use name `fifty_cities_by_random`

olmoearth_pretrain/evals/datasets/fifty_cities_dataset.py
```diff
@@ -0,0 +1,379 @@
...
+    "08 - NIR": {"mean": 2253.265, "std": 1380.702},
+    "08A - Vegetation Red Edge": {"mean": 2295.290, "std": 1352.824},
+    "09 - Water vapour": {"mean": 2345.512, "std": 1454.604},
+    "11 - SWIR": {"mean": 2099.266, "std": 1370.791},
+    "12 - SWIR": {"mean": 1754.644, "std": 1325.780},
+}
+S1_BAND_STATS = {
+    "vh": {"mean": -17.401, "std": 5.338},
+    "vv": {"mean": -10.663, "std": 5.930},
+}
```
yawenzzzz, line 68: If min/max, mean/std are computed using the same script, why we hardcoded mean/std here, and put min/max in the json file under config?

yawenzzzz (commented):
Two minor comments.

olmoearth_pretrain/evals/datasets/fifty_cities_dataset.py
gabrieltseng: this matches the existing pattern (min / man stats in the config, mean / std hardcoded). E.g. pastis, mados. I am not sure why we used this pattern to begin with

gabrieltseng: For now i'd prefer to leave this as is for consistency but i agree this is weird and should be addressed. Made an issue here: https://github.com/allenai/olmoearth_pretrain/issues/579

gabrieltseng:
@Farbum here are some linear probing results. S2 and S1+S2 plateau but S1 increases steadily throughout. Also we see consistent benefits from adding S1 which is nice!

fifty_cities_by_city_over_steps
fifty_cities_by_continent_over_steps
fifty_cities_random_over_steps

olmoearth_pretrain/evals/datasets/configs.py
gabrieltseng: added a comment. I would prefer not to change the name since i already have quite a few runs with this naming.

PR #577, opened 2026-06-15
era5 encoder

Farbum:
Era5 encoder pretraining strategy with 3 objectives:
- Objective A: supervised loss on multiple tasks
- Objective B: reconstruction loss on SWT
- Objective C: forecast of s2 in latent space


olmoearth_pretrain/train/train_module/train_module.py
```diff
@@ -500,9 +500,7 @@ def _get_state_dict(
         model_state_dict = dist_cp_sd.get_model_state_dict(
             self.model, options=sd_options
         )
-        optim_state_dict = dist_cp_sd.get_optimizer_state_dict(
-            self.model, self.optimizer, options=sd_options
-        )
```
gabrieltseng, line 505: what's the motivation for this change?

olmoearth_pretrain/nn/era5_encoder.py
```diff
@@ -0,0 +1,608 @@
...
+        self.embedding_size = d_model
+        self.patch_kernel_size = config.patch_kernel_size
+        self.patch_stride = config.patch_stride
+        self.pooling = Era5Pooling(config.pooling)
+        self.add_day_of_year_features = config.add_day_of_year_features
+        self.add_relative_position_features = config.add_relative_position_features
+
+        # Learned per-band mask embedding for reconstruction masking
+        if config.use_mask_embed:
+            self.mask_embed = nn.Parameter(torch.zeros(1, 1, in_channels))
```
gabrieltseng, line 237: As we discussed - I don't love using a mask token in the encoder. It introduces a distribution shift, since the encoder should never see that mask token during inference / test time. 

[BERT](https://arxiv.org/pdf/1810.048050) uses an 80/10/10 strategy (80% [mask], 10% unchanged, 10% random) strategy

Image

But maybe with the other objectives - which wouldn't pass corrupted data - this is handled and its ok overall

olmoearth_pretrain/nn/era5_encoder.py
```diff
@@ -0,0 +1,608 @@
...
+        if not self.training or self.drop_prob == 0.0:
+            return x
+        keep_prob = 1.0 - self.drop_prob
+        shape = (x.shape[0],) + (1,) * (x.ndim - 1)
+        random_tensor = keep_prob + torch.rand(shape, dtype=x.dtype, device=x.device)
+        random_tensor.floor_()
+        return x.div(keep_prob) * random_tensor
+
+
+class _TransformerBlock(nn.Module):
```
gabrieltseng, line 82: We already have a lot of this in `attention.py`. 

(e.g. `DropPath` for the `_StochasticDepth`, `Block` for the `_TransformerBlock` and `Block` supports cross attn so I don't think the `_CrossAttentionBlock` is necessary either)

olmoearth_pretrain/nn/era5_encoder.py
```diff
@@ -0,0 +1,608 @@
...
+            token_mask = torch.cat([cls_valid, token_mask], dim=1)
+
+        # Guard against fully-padded sequences producing NaN in attention
+        empty_rows = ~token_mask.any(dim=1)
+        if empty_rows.any():
+            token_mask = token_mask.clone()
+            token_mask[empty_rows, 0] = True
+
+        # --- Transformer blocks ---
+        tokens = self.token_dropout(tokens)
```
gabrieltseng: nit - `token_dropout` makes me think you are dropping tokens out. But this is just normal dropout being applied, so maybe just `dropout` is  a better name for this

olmoearth_pretrain/nn/era5_encoder.py
```diff
@@ -0,0 +1,608 @@
...
+            "num_cls_tokens": torch.tensor(num_cls, device=tokens.device),
+            "num_data_tokens": torch.tensor(num_data, device=tokens.device),
+        }
+
+    # ------------------------------------------------------------------
+    # Internals
+    # ------------------------------------------------------------------
+
+    def _patchify(self, seq: Tensor, valid_mask: Tensor) -> tuple[Tensor, Tensor]:
+        """Conv1D patch embedding + derive per-token validity mask.
```
gabrieltseng: The code does quite a bit of work to handle this validity mask throughout, but it seems like the sequence length is fixed and this is always set to all False?

If we intend to use it in the future it might be worth keeping but for the moment it makes it trickier to follow the code without doing anything (there's an error in `multi_task_era5_dataset.py` which states `"ERA5 is a dense reanalysis product — all samples must have uniform length."`)

scripts/copy_wandb_run.py
```diff
@@ -19,9 +19,10 @@
 from collections import defaultdict
 
 import requests
-import wandb
 from tqdm import tqdm
 
+import wandb
```
gabrieltseng, line 24: this is an isort error which stems from having a folder named `wandb` in the repository

olmoearth_pretrain/data/multi_task_era5_dataset.py
```diff
@@ -0,0 +1,950 @@
...
+from olmoearth_pretrain.evals.task_types import TaskType
+
+logger = logging.getLogger(__name__)
+
+
+_ERA5_MODALITY = Modality.ERA5L_DAY_10
+
+
+@dataclass
+class Era5Sample:
```
gabrieltseng, line 66: I think this is fine for now but in the future - our `OlmoEarthSample` supports era5 so it would be better to work with it than to redefine a new sample type.

olmoearth_pretrain/nn/transforms/era5_corruption.py
```diff
@@ -0,0 +1,574 @@
...
+#    "evaporative_flux": ["e", "pev"],
+#    "hydro_flux": ["tp", "ro"],
+# }
+
+# TODO: for ablations
+# instead of reconstructing raw signals then swt transform to band signals,
+# we can decode bands directly, then reconstruct raw/time signal from inverse swt
+# This forces the reconstruction to pass through the multiscale representation
+
+DEFAULT_VARIABLE_GROUPS: dict[str, list[int]] = {
```
gabrieltseng, line 85: This approach seems atypical compared to other models which mask weather data (cBottle, ClimaX, Prithvi-W either mask all the variables together, or treat them as an unstructured set). 

In general, other research seems preoccupied with not making the masking task too *easy*, so it would be good to ablate this since grouping in this way probably makes the task easier.

Farbum: Yes so far there are 3 masking strategies, a "naive" masking strategy which masks variables randomly, the temporal interpolation strategy which masks most of the variables all together, and a structured x-variable masking strategy that is based on heuristics and relies heavily on DEFAULT_VARIABLE_GROUPS. Ablation experiments are running right now.

olmoearth_pretrain/data/multi_task_era5_dataset.py
Farbum: I agree that we should later try to have OlmoEarthSample support it. For now, though it is very convenient to link a sample to its associated task for loss/task routing. Ignore_mask is also a nice to have for now as it determines which timesteps the model sees. Once we figure out what is the best architecture/masking strategy we can merge with OlmoEarthSample. For now I am trying to have the era5 encoder as decoupled as I can to Olmoearth_pretrain for faster iterations.

olmoearth_pretrain/nn/era5_encoder.py
Farbum: Agreed. I cannot yet think of a way to change that BERT style masking strategy as it allows us to nicely deal with small targeted masks. The masking strategy ablation results should tell us if these small corruptions are superior to heavier/longer masks that could leverage a different masking style (MAE etc.). As of now, tokenizing at the day level seem a bit odd since there is little correlation between variables, and there are only 14 of them. Moving to hourly era5 data would free us from these constraints as daily tokenization would make much more sense.
Yes to your last comment, objective C (reconstruction of residual S2 images in the latent space) should correct this distribution shift as no masking will be applied. Idem for objective A (full supervision) if we end up adding it to the mix.

olmoearth_pretrain/nn/era5_encoder.py
Farbum: makes sense, updated the naming

olmoearth_pretrain/nn/era5_encoder.py
Farbum: Good catch, it was originally added to deal with the extra buffer to the original 365 sequence (for swt lossless decomposition), but the buffer protection is already enforced by the target_start parameter in the corruption function. I don't see a reason to keep it as to your point it complicates the code for no obvious benefits. Objective C with shorter sequence should not need it either as the _patch_pad_amount helper should protect against mismatch between kernel/stride sizes and input sequence length. I am removing it entirely, great catch!

PR #580, opened 2026-06-24
Update readme with v1.1 model weights

gabrieltseng:


favyen2 (approved):
Looks good but I wonder if it would make sense to link to the report so it's easier for reader to see what's changed.  For v1.2 we can do the same and just note that the paper describes both v1.1 and v1.2 changes. Anyway it's up to you, whether to merge as is or to add a link like that.

favyen2:
Oh I see the changelog actually has a reasonable summary of the model changes, in that case it sounds good to me.

favyen2:
hm but if we aren't linking to the report anywhere else it would make sense to link here

favyen2 (approved):

PR #582, opened 2026-06-25
in loop evals in new job

gabrieltseng:


piperwolters:
Tested here: [https://beaker.org/orgs/ai2/workspaces/earth-systems/work/01KVZWG2SP91GBFE74NDRK7S[…]kId=01KVZWG302R9WAPZF7PMBTGTZD&jobId=01KVZWG35B9PFXK0BCH9Z1KE13](https://beaker.org/orgs/ai2/workspaces/earth-systems/work/01KVZWG2SP91GBFE74NDRK7S9D?taskId=01KVZWG302R9WAPZF7PMBTGTZD&jobId=01KVZWG35B9PFXK0BCH9Z1KE13)

Successfully launched the eval job here: [https://beaker.org/orgs/ai2/workspaces/earth-systems/work/01KVZXN4F7NGRQCGRFF6PN9E[…]kId=01KVZXN4NN0V31FFM2DNF2T79R&jobId=01KVZXN4S22JY6KXEHRAHT7F97](https://beaker.org/orgs/ai2/workspaces/earth-systems/work/01KVZXN4F7NGRQCGRFF6PN9E3D?taskId=01KVZXN4NN0V31FFM2DNF2T79R&jobId=01KVZXN4S22JY6KXEHRAHT7F97)

Just needed to push olmoearth_pretrain/internal/loop_eval_launch.py from one of your other branches, Gabi, and make sure beaker_eval_module_path is set.
Also, for this test run I set eval cadence to 10 steps and max steps 50 (very small), so as the eval jobs for steps 10 and 20 were wrapping up, their checkpoints got deleted and then the eval jobs "failed". This won't happen again while we have a normal eval step rate, but something to note.

olmoearth_pretrain/internal/checkpoint_sweep_evals.py
```diff
@@ -277,6 +289,34 @@ def parse_steps(steps_str: str | None) -> list[int] | None:
     return [int(s.strip()) for s in steps_str.split(",") if s.strip()]
 
 
+def get_train_run_eval_tasks(
```
gabrieltseng, line 284: without this function, the keys in the training script had to exactly match all_evals (and custom args would get lost).

This lead to errors like this:

```
[2;36m[2026-07-01 10:35:48][0m [34mINFO    [0m [2m[olmoearth_pretrain.train.callbacks.evaluator_callback:1133, rank=0][0m Skipping pastis_sentinel2 because it is not in the tasks_to_run list
```
because the script named it `pastis` not `pastis_sentinel2`

olmoearth_pretrain/internal/checkpoint_sweep_evals.py
```diff
@@ -277,6 +290,72 @@ def parse_steps(steps_str: str | None) -> list[int] | None:
...
+            "(%s); falling back to the shared eval catalog.",
+            e,
+        )
+        return None
+    evaluator_config = train_trainer_config.callbacks.get("downstream_evaluator")
+    tasks = getattr(evaluator_config, "tasks", None)
+    return tasks or None
+
+
+def add_experiment_to_loop_eval_group(
```
gabrieltseng, line 312: currently each eval loop kicks off a new beaker job which spams the beaker console. this keeps them in a single group

gabrieltseng: okay the group doesn't actually prevent it from spamming the beaker console but i'm leaving it since it does help with run organization

piperwolters (approved):
This now doesn't try to log for ephemeral checkpoints (which was causing issues earlier), is logging all evals for a run to 1 wandb process (cleaner than 1 process per eval per job), and added checkpoint step plot so we can double check which step evals are for. 

looks good to me!

PR #586, opened 2026-06-30
Attempt to add rank max learning rate logic into in-loop-evals PR

piperwolters:
Based on tests, this is seeming to run fine with `run_as_beaker_job` True or False. When True, each eval job is auto launched with rank_max_lr_num_gpus (default is 8). If False, the GPU amount just uses the training world size. 

Logic taken from: https://github.com/allenai/olmoearth_pretrain/pull/505

Testing in progress here:
- in same job: https://beaker.org/orgs/ai2/workspaces/earth-systems/work/01KWDEKQ2DMH7F7RXS7ME2T03K?taskId=01KWDEKQ8XS07P3JEG12JJ9EED&jobId=01KWDEKQCB293H1C1841GKDFJZ
- in separate job: 
   - train: https://beaker.org/orgs/ai2/workspaces/earth-systems/work/01KWDZCEAHQZZBPWW6HEG7B5ZD?taskId=01KWDZCEKH4XM4VZF3V0TAZRNQ&jobId=01KWDZCEVHM5TQGGT0DGQNE00V
   - eval: https://beaker.org/orgs/ai2/workspaces/earth-systems/work/01KWE49Q8BMXGD52RJ8N0WPJT2?taskId=01KWE49QEPN8WBNH6YBTDBRV82&jobId=01KWE49QJ27YFCP5PXXYY6B4RV


gabrieltseng:
comment from @pjreddie - since we run the evals in a separate job now, can we sweep learning rates but do it sequentially, so this only requires a single GPU ? 

This might make the evals very slow to run though

piperwolters:
> comment from @pjreddie - since we run the evals in a separate job now, can we sweep learning rates but do it sequentially, so this only requires a single GPU ?
> 
> This might make the evals very slow to run though

Hmm good point. Is there usually a "rush" to get results when sweeping learning rates, or is it more of an after-thought once we've found a good model?

PR #587, opened 2026-07-01
Add AUROC metric option

favyen2:


favyen2:
@claude can you review this PR?

olmoearth_pretrain/evals/metrics.py
```diff
@@ -216,6 +228,51 @@ def from_regression(
...
+    Returns:
+        Mean of per-class ROC AUC. Classes with only positive or only negative
+        ground-truth samples are skipped (AUC undefined). Returns NaN if every
+        class is degenerate.
+    """
+    aucs = []
+    for c in range(num_classes):
+        y_true_c = (labels_idx == c).astype(int)
+        # AUC is undefined when a class has no positives or no negatives.
+        if y_true_c.min() == y_true_c.max():
```
Farbum: Would that raise an error when all labels = ignore_label, as the labels become a size-0 array? Should we add a guard an return nan in that case?

Farbum (commented):
LGTM, just a minor comment.

What are your thoughts about adding PRAUC as well, since it is quite useful for imbalanced classes. We could also add it as a separate PR.

olmoearth_pretrain/evals/metrics.py
favyen2: I think that's true, if labels_idx is empty then it could get ValueError from the metric. Changed it to return NaN if there is no data.

favyen2:
Added PRAUC as well.

Farbum (approved):
LGTM, thanks for adding PRAUC

PR #589, opened 2026-07-01
Add AEF tasks

favyen2:
This adds a few of the AlphaEarth supplementary evaluation tasks that were in rslp.olmoearth_evals.

These were originally obtained from https://zenodo.org/records/16585402 and were converted to rslearn datasets in https://github.com/allenai/rslearn_projects/pull/174

In this PR they are now made available as in-loop evaluation tasks.

Here is W&B running the evals on v1.1 base checkpoint: https://wandb.ai/eai-ai2/20260701_olmoearth_pretrain_eval_aef


piperwolters (approved):
looks good to me

PR #592, opened 2026-07-06
Add vessel type/length evals along with kenya_intercropping and surface_fuels

favyen2:
Add evals:

* kenya_intercropping and surface_fuels are from partner tasks from the workshop that are derived from public datasets.
* In addition, this adds `{sentinel2,sentinel1,landsat}_vessel_{type,length}` (attribute prediction for vessel type and length)

The PR includes code changes to support "scalar regression" (regress one value for the window). Previously there was only per-pixel regression support.

W&B: https://wandb.ai/eai-ai2/20260703_olmoearth_pretrain_eval


olmoearth_pretrain/evals/datasets/configs.py
```diff
@@ -13,7 +13,11 @@ def get_eval_mode(task_type: TaskType) -> str:
     """Get the eval mode for a given task type."""
     if task_type == TaskType.CLASSIFICATION:
         return "knn"
-    if task_type in (TaskType.SEGMENTATION, TaskType.REGRESSION):
+    if task_type in (
+        TaskType.SEGMENTATION,
+        TaskType.REGRESSION,
```
piperwolters: should we call it dense_regression ?

olmoearth_pretrain/evals/datasets/rslearn_builder.py
```diff
@@ -260,9 +260,14 @@ def _classify_task(task: Any) -> dict[str, Any]:
...
-        if kind in name:
-            return kind
+    if "segmentation" in name:
+        return "segmentation"
+    if "classification" in name:
+        return "classification"
+    if "regression" in name:
+        # PerPixelRegressionTask is dense (per-pixel); plain RegressionTask reads a
+        # vector target (one value per window) -> scalar regression.
+        return "regression" if "perpixel" in name else "scalar_regression"
```
piperwolters, line 270: Oh or perpixel_regression if that's how you have it here

olmoearth_pretrain/internal/all_evals.py
```diff
@@ -639,6 +639,128 @@ def load_user_module(path: str) -> Any:
...
+        num_workers=8,
+        pooling_type=PoolingType.MEAN,
+        norm_stats_from_pretrained=True,
+        norm_method=NormMethod.NORM_NO_CLIP_2_STD,
+        probe_lr=0.01,
+        eval_interval=Duration.epochs(10),
+        input_modalities=[Modality.LANDSAT.name],
+        epochs=50,
+        eval_mode=EvalMode.LINEAR_PROBE,
+        use_center_token=True,
```
piperwolters, line 690: What does this arg do?

piperwolters (approved):
Aside from maybe renaming regression -> dense_regression or perpixel_regression in the TaskType, looks good to me!

olmoearth_pretrain/evals/linear_probe.py
```diff
@@ -323,6 +323,9 @@ def forward(self, x: torch.Tensor) -> dict:
                 i=self.num_output_pixels_per_side_of_patch,
                 j=self.num_output_pixels_per_side_of_patch,
             )
+        elif self.task_type == TaskType.SCALAR_REGRESSION:
+            # Single value per sample: (B, 1) -> (B,).
+            logits = logits.squeeze(-1)
```
gabrieltseng, line 328: Is this necessary? If `num_classes > 1` then we wouldn't want this, so I'd either have a warning (during init, that scalar regression doesn't support num_classes > 1) or not have the unsqueeze so that we can support multiple scalar regression classes

gabrieltseng (approved):

olmoearth_pretrain/evals/datasets/rslearn_builder.py
favyen2: OK renamed REGRESSION -> PER_PIXEL_REGRESSION and SCALAR_REGRESSION -> WINDOW_REGRESSION.

olmoearth_pretrain/internal/all_evals.py
favyen2: It should be documented in `olmoearth_pretrain/evals/eval_wrapper.py`, it will train a linear probe on top of the token from the center spatial patch (still pooling across modalities/timesteps if applicable) instead of on the token after pooling spatially across all patches. Since vessel is in the middle of the image it makes sense for that task. I'll add a comment here to mention that it's because the crops are centered at the vessel.

olmoearth_pretrain/evals/linear_probe.py
favyen2: OK I added a ValueError if num_classes != 1 for regression tasks. I think supporting regressing multiple values within one regression task would need some thought about how to expose the per-sub-regression-task losses vs the average loss over all the sub-tasks and whether to support weighting different sub-tasks differently or things like that; but I think it makes sense to explore that later when a need comes up.

PR #593, opened 2026-07-09
similar but different eval

piperwolters:
30k Sentinel2 patches across 10 worldcover classes, selected so that every patch has an "RGB doppelganger" from a different class. Seems like a promising way to evaluate if our models are actually utilizing multispectral information. I'm not adding this to in-loop evals, but it could be something we evaluate models on if we want to assess this specifically. 

Linear probe experiment with olmoearth v1.2 achieved 0.862, compared to the reported 0.867

### DIFF
--- a/olmoearth_pretrain/internal/all_evals.py
+++ b/olmoearth_pretrain/internal/all_evals.py
@@ -76,6 +76,25 @@ EVAL_TASKS = {
         eval_mode=EvalMode.KNN,
         primary_metric=EvalMetric.ACCURACY,
     ),
+    # Similar But Different: 10-class ESA WorldCover land-cover classification on
+    # single-timestamp Sentinel-2 L2A 32x32 patches, built so RGB is uninformative
+    # (a probe for spectral-band reliance). Linear probe on frozen embeddings -- the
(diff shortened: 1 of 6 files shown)


olmoearth_pretrain/internal/all_evals.py
```diff
@@ -76,6 +76,25 @@ def load_user_module(path: str) -> Any:
...
+        probe_batch_size=128,
+        num_workers=8,
+        pooling_type=PoolingType.MEAN,
+        norm_stats_from_pretrained=True,
+        norm_method=NormMethod.NORM_NO_CLIP_2_STD,
+        probe_lr=0.01,
+        eval_interval=Duration.epochs(5),
+        input_modalities=[Modality.SENTINEL2_L2A.name],
+        epochs=50,
+        eval_mode=EvalMode.LINEAR_PROBE,
```
gabrieltseng: From the announcement (https://geospatialml.com/posts/similar-but-different/) this is a classification task, so it should default to kNN. Is there a reason to default to linear probe?

gabrieltseng (approved):

PR #595, opened 2026-07-13
Training speedups

gabrieltseng:
Three changes which speed up v1.2 model training by ~43 hrs.

| config | to 500k steps | ratio | median step | full run |
| --- | --- | --- | --- | --- | 
| unfused FSDP (v1.2, original) | 108.9 hrs | 1.000 | 0.636 s | 147 h (projected) | 
| + fused AdamW | 92.3 h | 0.848 | 0.534 s | ~130 h (projected) |
| + DDP | 80.4 h | 0.739 | 0.479 s | 107.2 h (completed) | 
| + proj-only target | 78.4 h | 0.720 | 0.465 s | 104.4 h (completed) |

Confirmed in https://wandb.ai/eai-ai2/2026_07_08_fused_adamw_e2e that the losses and the downstream evals stay consistent.


scripts/official/v1_2/base.py
```diff
@@ -181,7 +181,9 @@ def build_train_module_config(
 ) -> ContrastiveLatentMIMTrainModuleConfig:
     """Build the train module config for an experiment."""
     return ContrastiveLatentMIMTrainModuleConfig(
-        optim_config=AdamWConfig(lr=0.0001, weight_decay=0.02, fused=False),
+        # fused=True: single-kernel AdamW update, ~20% faster steps at matched
+        # loss (validated vs run q3x9lvww in 2026_07_09_compile_fused_ab).
+        optim_config=AdamWConfig(lr=0.0001, weight_decay=0.02, fused=True),
```
favyen2, line 186: Might be nice to have the speedup config here so it's obvious when branching off of it. (Like the ddp and lightweight target encoder parts.) Or comment here saying that the other version works the same but is faster.

olmoearth_pretrain/train/train_module/train_module.py
```diff
@@ -443,6 +460,39 @@ def optim_step(self) -> None:
...
 
+    def _broadcast_params(self) -> None:
+        """Broadcast params and buffers from rank 0 so replicas start identical."""
+        if not (dist.is_available() and dist.is_initialized()):
+            return
+        group = self.dp_process_group
+        for t in itertools.chain(self.model.parameters(), self.model.buffers()):
+            dist.broadcast(t.detach(), src=0, group=group)
+
+    def _all_reduce_grads(self) -> None:
```
favyen2, line 471: So when DDP is enabled, it handles the gradient syncing manually? Could finding a workaround where DDP built-in sync functionality can still be used make it faster? My AI agent used a similar all reduce approach for gradients for NAIP discriminator but it was saying the performance impact would be negligible since it was a very small part of the model, so I was thinking maybe there's additional speedup if you use the DDP approach where maybe it collects the gradients more efficiently (in parallel with the backward pass I think).

favyen2 (approved):
Looks great, thanks for finding these optimizations!

olmoearth_pretrain/train/train_module/train_module.py
favyen2: e.g. maybe we can make it run the two masked views together in one forward pass where they're concatenated on the batch dimension; probably requires some work though to thread it through the latent mim nn.Module.

gabrieltseng: or get rid of the two forward passes

gabrieltseng: i agree this is worth investigating but maybe out of scope for this PR? (or -- we should get this merged in to benefit from the speedups while we keep investigating)

scripts/official/v1_2/base.py
gabrieltseng: Added to `base_faster.py`

PR #601, opened 2026-07-23
Various dataset improvements

favyen2:
- Support non-grid datasets, e.g. you could create 128x128 windows not aligned to any grid. The window names are based on an arbitrary example ID instead of based on the CRS/col/row.
- Drop the res_0.625 and res_160 groups so all windows are 10 m/pixel. This greatly simplifies the code and removes a lot of complexity that we weren't using.
- Share some arguments across all rslearn -> OlmoEarth format converters. Added `--group` argument too so you don't need to use the `res_10` name.
- Use PerLayerStorageFactory for all data sources to reduce the number of files. Also update Sentinel-1 and Sentinel-2 L2A to use the olmoearth_datasets service for faster prepare.
- Ingest OpenStreetMap from 2x2 tiles instead of the country-level PBFs so it is much faster and doesn't run out of memory.
- Miscellaneous modernization to not use deprecated APIs in rslearn and other places. Like `window.data.read_raster` instead of initializing the RasterFormat directly.


favyen2:
I'm going to split this into two PRs.

PR #603, opened 2026-07-24
Various dataset creation improvements

favyen2:
- Use PerLayerStorageFactory for all data sources to reduce the number of files. Also update Sentinel-1 and Sentinel-2 L2A to use the olmoearth_datasets service for faster prepare.
- Ingest OpenStreetMap from 2x2 tiles instead of the country-level PBFs so it is much faster and doesn't run out of memory.
- Drop the res_0.625 and res_160 groups so all windows are 10 m/pixel. This greatly simplifies the code and removes a lot of complexity that we weren't using. It removes about 700 lines of code (it says +7,035 but 6,300 of that is just the OpenStreetMap config, specifying all those 2x2 tiles).
- Miscellaneous modernization to not use deprecated APIs in rslearn and other places. Like window.data.read_raster instead of initializing the RasterFormat directly.


olmoearth_pretrain/dataset/convert_to_h5py.py
```diff
@@ -626,18 +618,6 @@ def _filter_samples(
...
-            # We require at least one spacetime varying modality to have 12 timesteps
-            # e.g., for Presto dataset, only 43 samples not meeting this requirement
-            longest_timestamps_array = self._find_longest_timestamps_array(
-                spacetime_varying_modalities
-            )
-            if len(longest_timestamps_array) < YEAR_NUM_TIMESTEPS:
-                logger.info(
-                    "Skipping sample because it does not have at least 12 timesteps"
-                )
-                continue
```
yawenzzzz, line 639: Shall we keep this? this ensures S2 or S1 or Landsat has at least 12 timesteps across the whole year, and later we will do temporal subsetting on this, not sure if removing this will cause any issue or not.

yawenzzzz (approved):
LGTM! only a minor comment regarding the longest timestep limit we have in h5 generation.

olmoearth_pretrain/dataset/convert_to_h5py.py
favyen2: yeah that makes sense, I think I removed it because in my open set dataset I have some windows that need to be shorter but for this PR it can keep the check

PR #604, opened 2026-07-29
Graceful shutdowns

gabrieltseng:
Previously, the sigterm from preemptions was getting fed through to the dataloader workers. These didn't shut down gracefully (crashed the training ranks midstep), occasionally requiring a GPU reboot.

This change means that the sigterm is ignored by the dataloader workers, allowing olmo-core's graceful shutdown to happen.


yawenzzzz (approved):
Thanks for the fix!

PR #608, opened 2026-09-15
Pretraining dataloader sampling controls

gabrieltseng:
Split out of gabi/perceiver (PR 1 of 8).


olmoearth_pretrain/data/dataloader.py
```diff
@@ -105,6 +113,30 @@ def __init__(
...
             token_budget: Optional token budget per instance.
+            patch_size_probs: Optional per-patch-size sampling probabilities aligned
+                with ``range(min_patch_size, max_patch_size + 1)``. If None, patch
+                sizes are sampled uniformly (the historical behaviour). Use this to
+                oversample small patch sizes for models deployed at that resolution.
+            time_priority_prob: Probability that a batch samples its number of
+                timesteps first (biased toward the full sequence) and then a spatial
+                grid that fits, rather than sampling the grid first. 0.0 reproduces
+                the historical space-first behaviour; >0 decorrelates grid size from
+                sequence length so that large-grid x full-year shapes occur.
```
favyen2: It's not clear here why "sampling the number of timesteps first and then a spatial grid that fits" would cause "large-grid x full-year shapes" to occur. Wouldn't those only occur if the token budget is increased?

olmoearth_pretrain/data/dataloader.py
```diff
@@ -105,6 +113,30 @@ def __init__(
...
+            temporal_bias: Skews the timestep draw toward the maximum of its feasible
+                window; timesteps are sampled with weight ``t ** temporal_bias``. 0.0
+                is uniform (the historical behaviour); larger values favour fuller
```
favyen2: I think it might be cleaner without the references to "the historical behavior".

olmoearth_pretrain/data/dataloader.py
```diff
@@ -105,6 +113,30 @@ def __init__(
...
+                sequences, restoring full-season exposure that uniform sampling
+                dilutes.
+            min_tokens_per_instance: Minimum token count a sampled shape must cost.
+                Shapes below the floor are excluded, so tiny grids are forced to pair
+                with long sequences (and vice versa) instead of collapsing to the
+                ``hw=1, t=1`` corner. 0 disables the floor.
+            max_timesteps: Maximum number of timesteps a sample can contribute.
+            tile_size: Spatial extent (in base-resolution pixels) of a training tile.
+                Used to bound the sampled grid so ``sampled_hw_p * patch_size`` fits.
+            exclude_only_decode_from_budget: If True, modalities the masking strategy
```
favyen2: This could be made cleaner by implementing this as a behavior change (always exclude decode-only from budget) instead of leaving `exclude_only_decode_from_budget` as an option.

olmoearth_pretrain/data/dataloader.py
```diff
@@ -139,6 +171,24 @@ def __init__(
         self.token_budget = token_budget
         self.patch_sizes = np.arange(min_patch_size, max_patch_size + 1)
         self.sampled_hw_p_list = sampled_hw_p_list
+        self.time_priority_prob = time_priority_prob
```
favyen2, line 182: The options are hard to understand from the individual argument descriptions in the docstring. It might be better to have a summary of the whole process at the top of the docstring. Like first we decide whether to sample timestep first or grid first, for timesteps first we use temporal bias then pick patch size based on patch_size_probs, for grid first we do this, it's all subject to max_timesteps / min_tokens_per_instance / token budget, etc.

olmoearth_pretrain/data/dataloader.py
gabrieltseng: removed

olmoearth_pretrain/data/dataloader.py
gabrieltseng: sorry for the unclear comment; removed

olmoearth_pretrain/data/dataloader.py
gabrieltseng: updated

olmoearth_pretrain/data/dataloader.py
gabrieltseng: I added the following description in the dataloader's docstring:

```
        The dataloader is responsible for subsetting our 128x128 24-timestep tiles.
        For a sampled patch size, we subset a spatiotemporal grid so that
        the number of tokens is <= token_budget and >= min_tokens_per_instance. Given
        this constraint, we use the following subsetting logic per microbatch:

        1. Define all possible grid sizes (hw, t) combinations that fit within the budget.
            Given the patch size we restrict hw so that hw * p is <= than the
            total 128x128 chip). We also restrict t to be <= max_timesteps
        2. with probability time_priority_prob, decide whether to sample timesteps or
            grid size.
            If sampling timesteps:
                i.  Sample some timestep t (from our possible timesteps, defined in step 1)
                ii. Sample some grid size hw which fits within this token budget and
                    yields at least min_tokens_per_instance
            If sampling grid size:
                i. Sample some grid size hw where there are timesteps that fit the min / max budgets
                ii. Sample a timestep t that respects the min / max budget
            In both the grid size and timestep sampling, we prefer more timesteps with a bias
            defined by temporal_bias.
        Decode only modalities are excluded from the token budget calculations.
```

Is this clearer?

olmoearth_pretrain/data/dataloader.py
```diff
@@ -570,31 +645,144 @@ def _get_batch_item_params_iterator(
...
-        """Get a generator that yields a tuple of (idx, patch_size, sampled_hw_p).
-
-        Changes patch_size and sampled_hw_p every rank_batch_size.
+    ) -> Iterator[tuple[int, int, int, int]]:
+        """Yield ``(idx, patch_size, sampled_hw_p, target_t)`` per instance.
+
+        The shape ``(patch_size, sampled_hw_p, target_t)`` is resampled every
+        ``rank_batch_size`` instances.
+
+        For a sampled patch_size, we subset a spatiotemporal grid (sampled_hw_p, target_t)
```
favyen2: Ah I see there was summary here. I think since the arguments are documented in the init docstring, it makes sense to put the summary there, and then it doesn't need to be duplicated here (would be difficult to keep the two copies of the text consistent), can just reference the docstring for details about the sampling process instead.

olmoearth_pretrain/data/dataloader.py
```diff
@@ -168,6 +213,34 @@ def __init__(
...
+        )
+
+        # Precompute per-instance band-set token rates so the shape sampler can
+        # invert the budget without a concrete sample in hand. Uses the full
+        # training modality set (a superset of any single sample), so the derived
+        # max_t is conservative and always fits the per-sample budget downstream.
+        training_modalities = getattr(self.dataset, "training_modalities", None)
+        if training_modalities is not None:
+            (
+                self._st_bandsets,
```
favyen2: I think verbosity would help here, `self._space_time_bandsets`, `self._space_only_bandsets`, `self._static_bandsets`, `self._time_only_bandsets`.

favyen2 (approved):

olmoearth_pretrain/data/dataloader.py
favyen2: It makes it much clearer at least for me!

PR #609, opened 2026-09-15
Perceiver-style register bottleneck, projection students and supervision head

gabrieltseng:
Split out of gabi/perceiver (PR 2 of 8).


olmoearth_pretrain/nn/attention.py
```diff
@@ -263,8 +276,16 @@ def sdpa(
...
-            if attn_mask is not None:
-                attn_mask = attn_mask[:, None, None].repeat((1, self.num_heads, n, 1))
+            if attn_mask is not None and attn_mask.dim() == 2:
+                # A 1D-per-sample key mask (B, Nk): add head/query dims and let SDPA
+                # broadcast over them. We deliberately do NOT materialize a dense
+                # (B, H, N, Nk) mask -- that ~N^2 tensor forces SDPA onto the math
+                # backend (full score materialization -> O(N^2) memory + slow). The
+                # broadcastable (B, 1, 1, Nk) mask lets the fused mem-efficient kernel
+                # handle padding instead.
+                attn_mask = attn_mask[:, None, None]
```
gabrieltseng, line 284: Reduces peak GPU memory allocation by ~30%, increases batches per second by ~9%

olmoearth_pretrain/evals/eval_wrapper.py
```diff
@@ -148,26 +212,39 @@ def __call__(
...
-                # Get spatial embeddings (B, H, W, D) then take center patch
-                batch_embeddings = pool_unmasked_tokens(
-                    batch_embeddings,
-                    self.pooling_type,
-                    spatial_pooling=True,
-                    concat_features=self.concat_features,
-                )
-                batch_embeddings = self._extract_center_token(batch_embeddings)
+            )
+            if (
```
favyen2, line 219: Seems like there should be error if `eval_on_projected_registers` is set but model doesn't have any registers / use_register_bottleneck. Currently it seems it would silently fallback to the pre-register encoder tokens.

olmoearth_pretrain/nn/flexi_vit.py
```diff
@@ -1477,6 +1761,15 @@ def __init__(
         rope_temporal_base: float | None = None,
         rope_temporal_coordinate_scale: float = 1.0,
         spatial_pos_encoding: str | None = None,
+        use_register_bottleneck: bool = False,
+        register_dim: int | None = None,
```
favyen2: Similar to `TokenizationConfig` can we move these arguments to a consolidated `RegisterBottleneckConfig`? e.g.:
```python
@dataclass
class RegisterBottleneckConfig(Config):
    register_dim: int
    latent_depth: int = 2
    num_heads: int | None = None          # None -> encoder num_heads
    per_depth_read_proj: bool = False
    attn_dim: int | None = None
    # detached student readout (could be its own small nested config, see below)
    projection_dims: list[int] | None = None
    projection_output_norm: bool = False
    back_projection_hidden: int | None = None
    def validate(self, encoder_num_heads: int, position_encoding: str) -> None:
        # the divisibility / head_dim % 4 / positive-dims checks, written ONCE
        ...
    def build(
        self,
        encoder_embedding_size: int,
        encoder_num_heads: int,
        mlp_ratio: float,
        position_encoding: str,
        rope_base: float,
        qk_norm: bool,
    ) -> "SpatialRegisterBottleneck":
        return SpatialRegisterBottleneck(
            encoder_embedding_size=encoder_embedding_size,
            register_dim=self.register_dim,
            num_heads=self.num_heads or encoder_num_heads,
            mlp_ratio=mlp_ratio,
            latent_transformer_depth=self.latent_depth,
            use_2d_rope=PositionEncoding.is_rope(position_encoding),
            rope_base=rope_base,
            qk_norm=qk_norm,
            per_depth_read_proj=self.per_depth_read_proj,
            attn_dim=self.attn_dim,
        )
```

olmoearth_pretrain/nn/flexi_vit.py
```diff
@@ -1603,12 +1940,96 @@ def __init__(
...
+                # teacher from student[..., :d]
+                if (
+                    register_back_projection_hidden is not None
+                    and register_back_projection_hidden <= 0
+                ):
+                    raise ValueError(
+                        "register_back_projection_hidden must be positive, got "
+                        f"{register_back_projection_hidden}"
+                    )
+                self.register_back_projections = nn.ModuleDict(
```
favyen2: It looks like `register_back_projections` is not used in the Encoder. It seems like it should be moved to the LatentMIM model instead, given that it is only called from the LatentMIM train module. Claude suggested that it could be like SupervisionHead and could compute the loss `head.loss(teacher, student)` within a `RegisterDistillationHead` class instantiated from `LatentMIM` and built from `RegisterDistillationHeadConfig` within `LatentMIMConfig`.

olmoearth_pretrain/nn/flexi_vit.py
```diff
@@ -1603,12 +1940,96 @@ def __init__(
...
+                attn_dim=register_attn_dim,
+            )
+            if self.register_projection_dims is not None:
+                if any(d <= 0 for d in self.register_projection_dims):
+                    raise ValueError(
+                        "register_projection_dims must be positive, got "
+                        f"{register_projection_dims}"
+                    )
+                student_dim = self.register_projection_dims[0]
+                self.register_projection = nn.Linear(resolved_register_dim, student_dim)
```
favyen2: Instead of having `self.register_projection` and `self.register_projection_norm` separately here, suggest to have a `RegisterStudent` containing them.

favyen2 (commented):
looked up to `flexi_vit.py` so far.

olmoearth_pretrain/train/train_module/train_module.py
```diff
@@ -418,9 +418,23 @@ def state_dict(self) -> dict[str, Any]:
     def state_dict_to_load(
         self, metadata: Metadata, optim: bool | None = None
     ) -> dict[str, Any]:
-        """Get the state dict to load."""
+        """Get the state dict to load.
```
favyen2, line 422: What is load_optim_state used for in the v1.3 experiments?

olmoearth_pretrain/model_loader.py
```diff
@@ -118,17 +150,377 @@ def _resolve_artifact_path(
...
+
+
+#: EncoderConfig fields removed from this version, keyed by name.
+#:
+#: Old checkpoints still carry these keys, and neither deserializer copes on its own:
+#: with olmo-core, ``Config.from_dict`` goes through omegaconf and RAISES on any unknown
+#: key, so even an inert leftover blocks the load; without olmo-core, the standalone
+#: path silently DROPS unknown keys, so a checkpoint that genuinely used one would load
+#: as though the feature had been off and quietly build a different model.
+#: :func:`patch_legacy_encoder_config` handles both: inert keys are stripped, active
```
favyen2: Can we make it so without olmo-core we keep patch_legacy_encoder_config to strip the keys that we don't need, but have it be strict so it raises if there are any extra keys? It doesn't seem like we should need to list all these potential extra keys explicitly here. Also most of the extra keys are anyway stuff isolated to these experiments that never made its way to main, so we don't need those.

olmoearth_pretrain/train/train_module/train_module.py
gabrieltseng: Ah these were leftover experiments when I tried continued training on the distillation head. I will revert these changes.

olmoearth_pretrain/evals/eval_wrapper.py
gabrieltseng: added a ValueError if we enter the `else` branch (and to the __init__)

olmoearth_pretrain/nn/flexi_vit.py
gabrieltseng: thanks - this is a good idea. added

gabrieltseng: For clarity, I renamed the RegisterBottleneck -> Perceiver since bottleneck could easily be confused with the distillation layer

olmoearth_pretrain/nn/flexi_vit.py
gabrieltseng: moved, added. thanks!

olmoearth_pretrain/nn/flexi_vit.py
gabrieltseng: I consolidated these two layers. I also generally renamed projection -> student

olmoearth_pretrain/model_loader.py
gabrieltseng: this is a good suggestion; thank you; implemented.

I've removed most keys from `REMOVED_ENCODER_FIELDS` but we still need a few extra keys explicitly listed to make sure they are harmless to remove.

olmoearth_pretrain/model_loader.py
```diff
@@ -118,17 +156,392 @@ def _resolve_artifact_path(
     return base / filename
 
 
+#: Removed config fields that the SHIPPED checkpoints (the v1.3 release and its two
```
favyen2, line 159: It's hard to understand what's going on here, maybe removing some of the constants and just putting it in the functions would help, I'm not sure, maybe you can go through it and see if you get some ideas to clean it up.

favyen2 (approved):
Looks good (although didn't really look carefully at everything since there are lots of changes) except if you can find a way to clean up `model_loader.py` that would be good.

gabrieltseng:
>  except if you can find a way to clean up model_loader.py that would be good.

I updated the checkpoint so that it is up to date with the code; this means we don't need to worry about the conversion anymore. I removed it from model_loader.py

olmoearth_pretrain/model_loader.py
gabrieltseng: simplified: https://github.com/allenai/olmoearth_pretrain/pull/609#issuecomment-5758165522

PR #610, opened 2026-09-15
Eval dataset plumbing, precomputed-embedding baselines and the AEF materializer

gabrieltseng:
Split out of gabi/perceiver (PR 3 of 8).


olmoearth_pretrain/evals/datasets/rslearn_builder.py
```diff
@@ -24,11 +24,50 @@
...
+        except OSError as e:
+            logger.warning(
+                "dataset index save failed (concurrent writer?), continuing "
+                "without cache: %s: %s",
+                type(e).__name__,
+                e,
+            )
+
+    save_windows._best_effort = True  # type: ignore[attr-defined]
+    DatasetIndex.save_windows = save_windows
```
favyen2, line 65: This seems like it could go in rslearn instead of overriding it here. Or change the evals to use the sqlite3 window metadata storage instead so no index is needed to have fast metadata access.

olmoearth_pretrain/evals/embedding_materializer/fetchers.py
```diff
@@ -0,0 +1,270 @@
...
+    Returns:
+        tuple of (start, end) datetimes spanning the year.
+    """
+    return (
+        datetime(year, 1, 1, tzinfo=UTC),
+        datetime(year, 12, 31, 23, 59, 59, tzinfo=UTC),
+    )
+
+
+class EmbeddingFetcher(ABC):
```
favyen2, line 62: Something doesn't make sense here -- there is a lot of code to wrap some rslearn data sources to acquire embeddings, but then usage is just to write the embeddings back to the rslearn dataset. It doesn't seem like any of this code should be needed if you're just using rslearn data sources to materialize embeddings into rslearn datasets.

olmoearth_pretrain/evals/studio_ingest/provenance.py
```diff
@@ -0,0 +1,186 @@
...
+
+# Env var override for the repo root, for run modes where walking up from
+# __file__ does not land in a checkout (e.g. a non-editable install).
+REPO_ROOT_ENV_VAR = "OLMOEARTH_REPO_ROOT"
+
+# Directory (repo-relative) where eval dataset configs are expected to live.
+RSLEARN_DATASET_CONFIGS_DIR = "data/rslearn_dataset_configs"
+
+
+def find_repo_root() -> Path | None:
```
favyen2, line 41: Instead of having this complexity here maybe we can make it similar to the normalization configs so the configs are installed together with the Python package and no locate code is needed.

olmoearth_pretrain/evals/studio_ingest/provenance.py
```diff
@@ -0,0 +1,186 @@
...
+        if (parent / "pyproject.toml").exists():
+            return parent
+    return None
+
+
+def sha256_of_file(path: str | Path | UPath) -> str:
+    """Compute the sha256 hex digest of a file (works on Weka/GCS via UPath)."""
+    upath = UPath(path)
+    digest = hashlib.sha256()
+    with upath.open("rb") as f:
```
favyen2, line 71: Since we pin >=3.11 I think you can just do `hashlib.file_digest(f, hashlib.sha256)`, probably don't need to define this `sha256_of_file` function at all.

scripts/tools/clear_stale_embedding_layers.py
```diff
@@ -0,0 +1,195 @@
+"""Delete embedding layers whose product year no longer matches their window.
```
favyen2, line 1: I don't think we need this script in the repo if it was a one-time update.

olmoearth_pretrain/evals/datasets/rslearn_dataset.py
```diff
@@ -124,9 +193,38 @@ def __init__(
             norm_method: Normalization method when not using pretrain stats.
             ds_norm_stats_json: Path to dataset norm stats JSON.
             ds_norm_stats: Dataset norm stats blob (e.g. from registry entry).
-            start_time: Start time for timestamp generation.
-            end_time: End time for timestamp generation.
+            start_time: Fallback start time for synthesized timestamps, used
```
favyen2, line 196: Are there some datasets where we don't have timestamps? It seems like to get images we should need timestamps. If there are no datasets without timestamps, then we could just raise an error if we unexpectedly encounter any windows without time range.

scripts/tools/wire_embedding_modalities.py
```diff
@@ -0,0 +1,612 @@
+"""Wire a materialized embedding product into eval datasets so evals can read it.
```
favyen2, line 1: This script seems more complicated than loading the embeddings separately from the model config or something like that.

olmoearth_pretrain/evals/datasets/rslearn_dataset.py
gabrieltseng: This just refactors existing functionality, but I agree its strange. I can try removing it but it might break old datasets?

olmoearth_pretrain/evals/embedding_materializer/fetchers.py
gabrieltseng: Agreed. Removed

